Issue #67, opened 2026-02-03
[FEATURE] Call Sign Size


phether:
## Is your feature request related to a problem?
You can make your own call much larger on the left side, just like in the original ham clock

whallmann:
It would be nice also to read the time from a little distance. 👍

Paul Hetherington ***@***.***> schrieb am Di., 3. Feb. 2026,
12:36:

accius:
@ThePangel resolved this. you can now go in the settings and change the callsign/time to a 2 from a 1

PR #86, opened 2026-02-03
Added the ability to resize the callsign. 

ThePangel:
This addresses issue #67 

I added the ability to resize the callsign name, from 0.1 to 2.0 of its base size (22px), In the main settings menu.

### DIFF
--- a/PLUGIN_DOCUMENTATION_SUMMARY.md
+++ b/PLUGIN_DOCUMENTATION_SUMMARY.md
@@ -1,0 +1,363 @@
+# 📚 Plugin Documentation Summary
+
+**Date:** 2026-02-03  
+**Status:** ✅ Complete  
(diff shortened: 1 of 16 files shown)


ThePangel:
Some localization is still needed, I just added the Spanish and English (those are the 2 languages I'm fluent in) entry of `"station.settings.callsignSize"`.

ThePangel:
https://github.com/user-attachments/assets/2b8972c2-e367-4bdd-ac08-50a48441bfd8

accius:
looks good, i pushed it to the main , would you mind also having it affect the times? so if you make the callsign bigger, it also adjusts the size of the times

ThePangel:
> looks good, i pushed it to the main , would you mind also having it affect the times? so if you make the callsign bigger, it also adjusts the size of the times

So bigger name bigger time? On it

PR #95, opened 2026-02-03
renamed callsignSize to headerSize, which now also affects the clocks

ThePangel:
This takes care the second comment in #67 

The `callsignSize` variable is now `headerSize`, since it now changes the size of all header elements (callsign, clocks, weather)

https://github.com/user-attachments/assets/1dc6e082-212c-48f5-99a2-7178ed89f851


accius:
thank you!

ThePangel:
no problem !

PR #103, opened 2026-02-04
An old school way to hide panels

brianbruff:
not sure i like this approach myself, but didn't see a lot of room to add panel collabsible icons..

disclaimer - changes done with opencode and glm-4.7 (not a patch on claude code, but got there in the end)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "ws": "^8.14.2"
   },
   "devDependencies": {
+    "@playwright/test": "^1.58.1",
     "@vitejs/plugin-react": "^4.2.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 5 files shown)


accius:
After digging in more I realized something I totally missed on my first look. The changes are against the monolithic file (public/index-monolithic.html), which is actually just preserved as a reference implementation. It's not meant to be modified directly. All active development happens in the modular codebase under src/ where the components, hooks, and all the newer features live. The monolithic basically just exists so contributors can reference the original structure and line numbers when porting things over.
The good news is the feature itself is solid and I definitely want to include it. The localStorage persistence and the map resize handling are both done right. It just needs to live in the right place.

brianbruff:
No problem, good to know,

Ill get a new pr and some screenshots  later today (ireland)

Tnx

PR #119, opened 2026-02-04
Dx weather

dmazan:
Break out the existing DE weather panel as a new component.  Reuse the new weather component to show the weather at the currently selected DX station location.  Add a feature toggle to make the inclusion of DX weather optional.  Also renamed the weather hook to remove "local" since it is also used for DX and updated the README.

It looks like there is some future thinking in how feature toggles are handled that wasn't entirely clear to me.  Yell if that part is wrong.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -358,10 +358,9 @@ const App = () => {
           dxSunTimes={dxSunTimes}
           handleDXChange={handleDXChange}
           localWeather={localWeather}
-          weatherExpanded={weatherExpanded}
-          setWeatherExpanded={setWeatherExpanded}
           tempUnit={tempUnit}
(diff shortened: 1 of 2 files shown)


accius:
this is currently showing conflicts and I do not have a minute to resolve them.  I will look at this when I get a second.  Thank you for your efforts and contributions, they are absolutely appreciated.

dmazan:
Yeah, no kidding.  The dockable theme from @brianbruff is a great addition but generated a lot of conflicts.  I've fixed them and created a new PR.  It also fixes, sort of, the expandable weather forecast but the dockable theme will eventually need some tweaks for panels that change size.  But let's let things stabilize a bit first.

PR #128, opened 2026-02-05
Dx weather

dmazan:
Replacement of the original PR for dx-weather that incorporates the dockable layout by @brianbruff.  Break out the existing DE weather panel as a new component. Reuse the new weather component to show the weather at the currently selected DX station location. Add a feature toggle to make the inclusion of DX weather optional. Also renamed the weather hook to remove "local" since it is also used for DX and updated the README.

Also fixes an issue with the DE weather in the dockable layout.  There remains an issue where panels that resize themselves are not accommodated by the dockable layout.  Something to be addressed in the future.

Should merge to main without conficts.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -358,10 +358,9 @@ const App = () => {
           dxSunTimes={dxSunTimes}
           handleDXChange={handleDXChange}
           localWeather={localWeather}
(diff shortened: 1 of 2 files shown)


accius:
this is perfect, merged to main. looks great!

Issue #19, opened 2026-02-02
[FEATURE] Improve DX Cluster Exclusion


rfreedman:
## Is your feature request related to a problem?
Existing DX Cluster "Exclude" feature excludes both DE and DX, should only exclude DX,
and only provides entry of callsigns (partial or full), which are used as wildcards to filter by callsign.

## Describe the Solution You'd Like
- This info provided by Greg, K3EA
- Keep the callsign filter as-is (but exclude DX only)
- Provide additional input for exclusion by IARU DXCC Entity (see https://www.arrl.org/files/file/DXCC/2022_DXCC_Current.pdf), by entry of the entity name (with autocomplete), or by callsign prefix, also to select an entity that is assigned that prefix. See, for example https://www.arrl.org/international-call-sign-series
 ### Examples:
  - enter "CA", and get a selection list of all entity names beginning with "CA", type an additional "N" to get all entity names beginning with "CAN", including Canada.
  - enter "VE" in the Entity Exclusion to exclude all Canada DX spots, regardless of actual callsign (VA, VO, VY, etc).
 

## Use Case
How would this feature benefit amateur radio operators? Be specific about the use case:
- [X] General HF operation
- [X] VHF/UHF operation
- [X] Satellite operation
- [X] Contesting
- [X] POTA/SOTA
- [X] DXing
- [ ] Emergency communications
- [ ] Other: ___________

## Would you be willing to help implement this?
- [X] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, but I'd use it!

## Similar Features in Other Software
See the original HamClock implementation and other Web-based DX Clusters

---
73!

PR #180, opened 2026-02-06
Issue #19 Improve DX Cluster Exclusion (complete)

rfreedman:
Revamp the Exlude filter tab, works and looks the same as the 'Zones' tab, but for DX (spots) instead of DE (spotters), and also has exclusions for both spot and spotter callsigns.

Not currently properly affecting the map - that will be taken care of in a later commit.

### DIFF
--- a/src/components/DXClusterPanel.jsx
+++ b/src/components/DXClusterPanel.jsx
@@ -20,15 +20,19 @@ export const DXClusterPanel = ({
 }) => {
   const getActiveFilterCount = () => {
     let count = 0;
+    if (filters?.continents?.length) count++;
     if (filters?.cqZones?.length) count++;
     if (filters?.ituZones?.length) count++;
-    if (filters?.continents?.length) count++;
     if (filters?.bands?.length) count++;
     if (filters?.modes?.length) count++;
     if (filters?.watchlist?.length) count++;
-    if (filters?.excludeList?.length) count++;
     if (filters?.callsign) count++;
     if (filters?.watchlistOnly) count++;
+    if (filters?.excludeContinents) count += filters.excludeContinents.length;
+    if (filters?.excludeCqZones) count += filters.excludeCqZones.length;
+    if (filters?.excludeItuZones) count += filters.excludeItuZones.length;
+    if (filters?.excludeCallList) count += filters.excludeCallList.length;
+
     return count;
   };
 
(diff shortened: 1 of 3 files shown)


rfreedman:
PR #180 has been merged and appears on openhamclock.com, but the new
'exclude callsigns' feature appears to not be working (I verified that it
does work on branch that was merged).

I haven't tested the other "exclude" options, but they should be verified
too on the main branchl

rfreedman:
I see that it's because your new useDXClusterData.js
<https://github.com/rfreedman/openhamclock/blob/main/src/hooks/useDXClusterData.js>
which
is taking the place of useDXCluster.js
<https://github.com/rfreedman/openhamclock/blob/main/src/hooks/useDXCluster.js>
does
not implement the filters that my commit implemented in useDXCluster.js
<https://github.com/rfreedman/openhamclock/blob/main/src/hooks/useDXCluster.js>
Currently still a problem in the public v3.14.14 and the staging v3.14.16

------------------------------
Rich Freedman
N2EHL

rfreedman:
BTW, I'm also curious about the usage of the apparent filtering functions
in utils/callsign.js, which are also not in sync with the latest filtering,
and probably should not duplicate the filtering that is now in
useDXClusterData.js (but as I mentioned, is not up to date with the latest
useDXCluster.js that was in my commit).

------------------------------
Rich Freedman
N2EHL

accius:
Rich, I created a unified function using your information, and made updates to everything uses that function to filter. Can you check staging for me, my computer is about to die.
________________________________

rfreedman:
ok, will do. thanks!

------------------------------
Rich Freedman
N2EHL

rfreedman:
I did a quick test - and most things appear to be working.
The one exception that I see is that I selected "all" in the "Exclude Spots
(DX) by ITU Zone" filter, and it did not seem to have any effect.

I haven't tested the the "include" filters on the Zones tab - I've got to
run out now to do some errands.
That will take quite a bit more time to test.

Given that trying to test the filtering in the actual running app is a
moving target (as the spots change), it might be time to introduce some
unit tests for the filtering functions.
If the filtering code is not easily tested because it's embedded in the
hook code, it might be easier to pull that code out to "util" function or
something, and call that from the hook,
so that the standalone functions can be tested by themselves.



------------------------------
Rich Freedman
N2EHL

PR #205, opened 2026-02-07
112 Add a Brisbane Timezone

alanhargreaves:
- Added an entry for Australia/Brisbane
- Added Canberra, Melbourne and Hobart for the Australia Eastern entry description


alanhargreaves:
What do I do with the bug, just put a comment in saying it's done and 
merged?

Regards,
Alan Hargreaves
VK2MG

PR #216, opened 2026-02-07
Add possibility to build app with electron

SebFox2011:
I fix the configuration and add Electron & Electron-builder to build and add ./dist-electron in gitignore.
Let you explain on Readme in the way you prefer and command to build a desktop app with Electron

To build app just run npm run electron-builder, app is packaged and unpackaged in dist-electron foolder.

Possible also if you have a compact view to create an Android build with Capacitor and IOS.


SebFox2011:
@accius you don't need this feature ?

Issue #212, opened 2026-02-07
[BUG] Bearings and distance information does not appear in the DX Location panel in the Dockable layout


dmazan:
## Describe the Bug
The newly added bearing and distance information in the DX Location panel does not appear in the Dockable layout.  It does appear in other layouts.

## To Reproduce
1. Go to Settings
2. Select Modern layout
3. DX Panel shows short path/long path bearings and distance
4. Go to Settings
5. Select Dockable layout
6. DX Panel no longer shows  short path/long path bearings and distance

## Expected Behavior
DX Panel shows short path/long path bearings and distance in Dockable layout

## Analysis
DockableApp.jsx duplicates the renderDXLocation code from App.jsx.  That code is updated in App.jsx but not in DockableApp.jsx.  That situation is likely to continue repeatedly as new features are added.

I haven't looked deeply enough to tell whether there might be render code changes needed to support the dockable layout.

## Proposed Solution
The cleanest solution would be to break out all of the renderXX methods to separate modules an import them into both App.jsx and DockableApp.jsx.  Maybe create a new directory /renders to contain all of the render code, each panel in its own module.

Since the DX Location panel is my favorite, I'm happy to take a shot at breaking that out.  @brianbruff your thoughts?  

---
73!

dmazan:
Maybe even better: also break out each of the layouts into a layouts/ directory.  App.jsx calls the appropriate layouts/ module, the layout module calls the appropriate renders/ modules as needed.  DockableApp.jsx might still be a stand-alone layout, calling the appropriate renders/ modules.  Or, it might just become a layouts/ module.

dmazan:
New pull request #217 submitted to apply this approach to harmonize the DE Location and DX Location panels between the Modern and Dockable themes.  (Those were the easiest as the panels only appear in those to themes.)

dmazan:
Updated PR #257 reapplies this fix following the layout refactor.

PR #217, opened 2026-02-07
Break out DE Location and DX Location rendering into separate render modules and reuse

dmazan:
Initial, partial effort at the issue described in bug #212.  DX Location render code was duplicated in App.jsx and DockableApp.jsx.  The recent enhancement to add short and long path bearings and distances to the DX Location panel only appeared in the App.jsx version (Modern theme) and did not make it into the Dockable theme handed by DockableApp.jsx.

DE Location and DX Location are fairly easy as those panels only appear in the Modern and Dockable themes.  Created new modules renders/renderDELocation and renders/DXLocation.  Replaced in-line code in App.jsx and DockableApp.jsx with calls to the renders/ modules.

And changes to either the DE or DX Location panels can now be made one time in the appropriate renders/ module and will appear in both the Modern and Dockable themes.

This approach should probably be repeated for other panels but may be more difficult as some appear in multiple themes.  It might be best to have App.jsx just call the selected theme and have a separate module for each theme to compose the appropriate renders/ modules.

### DIFF
--- a/src/layouts/ModernLayout.jsx
+++ b/src/layouts/ModernLayout.jsx
@@ -15,6 +15,9 @@ import {
   AnalogClockPanel
 } from '../components';
 
+import DELocationPanel from '../renders/renderDELocation.jsx';
+import DXLocationPanel from '../renders/renderDXLocation.jsx';
+
 export default function ModernLayout(props) {
   const {
     config,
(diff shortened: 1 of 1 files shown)


dmazan:
I'll redo this later today so it works with the refactored theme handling and can be merged without conflicts.

PR #218, opened 2026-02-07
Refacto App component because is too big

SebFox2011:
App.jsx became very to big to be maintain properly, i suggest cutting in app hooks logics and layout components.
After this it will be easilier to cut into smaller components.

Seb - 73's de F4JRN

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -5,30 +5,14 @@
 import { useState, useEffect, useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
-// Components
-import {
-  Header,
-  WorldMap,
-  DXClusterPanel,
-  POTAPanel,
-  ContestPanel,
-  SettingsPanel,
-  DXFilterManager,
-  PSKFilterManager,
-  SolarPanel,
-  PropagationPanel,
-  DXpeditionPanel,
-  PSKReporterPanel,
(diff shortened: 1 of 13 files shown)


dmazan:
This is excellent but still duplicates lots of code between the Modern and Dockable themes.  Any change to those panels will have to be done twice.  PR #217 breaks the DE Location and DX Location panels out so there is no duplication.  Could go either way: (a) pull #217 into this more complete refactor; or (b) merge this PR into a new release and then I'll redo my effort to break out the two panels.

SebFox2011:
@dmazan i don't know how to integrate your PR 217 in my refacto?
I let you decide what is the best choice and i could also adapt my job.

dmazan:
@SebFox2011 Easy enough.  Merge my branch (separate-render-modules).  the renders/ directory and DockableApp.jsx shouldn't generate any conflicts.  App.jsx will, reject them.  Copy the DE and DX Location calls from DockableApp.jsx to the new ModernLayout.jsx and you should be good.  If you run into trouble `git merge --abort` is your friend.

Or @accius can just merge your PR first and I'll do the merge and resubmit my PR.  No big deal either way.

Good job on the refactor.

SebFox2011:
@dmazan ok if ok for you i let @accius merge my PR and let you resubmit. Next time it will be for me :)

PR #225, opened 2026-02-07
Add Ambient Weather panel and hook

Delerius:
### Summary
Adds a dedicated **Ambient Weather** panel and integration hook.

### Changes
- New `AmbientPanel` with configurable display fields
- New `useAmbientWeather` hook for AmbientWeather.net API
- Ambient weather removed from DE Location panel
- Ambient added as its own panel (dockable + classic layout)
- Updated exports and layout defaults
- Updated `.env.example` with required `VITE_` keys

### Configuration
Add to `.env.local`:
- `VITE_AMBIENT_APPLICATION_KEY`
- `VITE_AMBIENT_API_KEY`

Optional:
- `VITE_AMBIENT_DEVICE_MAC`
- `VITE_AMBIENT_POLL_SECONDS` (min 15)

### Notes
Display preferences are stored in localStorage for ease of use and to avoid rebuilds.

### Screenshots

**Ambient Weather Panel**
Screenshot 2026-02-07 124742

Screenshot 2026-02-07 124801


accius (changes requested):
Please check all code (speciffically app.jsx) for conflicts, there has been some updates that created conflicts with this.  also does this replace completely the current weather? or is this an additional view that can be turned off and on?  I would prefer we don't jsut get rid of someones contributions based on what someone else prefers, I would rather give the user the choice to use either.  please check on these things and update.

Delerius:
Updated the branch with the latest upstream main and resolved conflicts (including App.jsx and DockableApp.jsx). Ambient Weather is now an additive as an optional dockable panel and does not replace/remove the existing Weather panel. Users can choose either/both. Please re-review.

accius:
thank you for your updates!

PR #236, opened 2026-02-08
Docs: add FLRIG rig control + API endpoints; align formatting (Prettier/.editorconfig)

echo-gravitas:
- Added Prettier config and aligned `.editorconfig` defaults to ensure consistent formatting across the team.  
- Documented formatting expectations in `README.md` (Contributing section).  
- Added FLRIG/rigctl rig control documentation, including configuration variables and `POST /api/rig/tune`.  
- Expanded the API endpoints list with the new rig control and other recently added endpoints.  
- Updated the Table of Contents to include the new rig control section.
- Use spot-provided timestamps for DX cluster data (DX Spider HHMMZ, HamQTH date+time, proxy timestamps) instead of fetch time, and pass them through `/api/dxcluster/paths`.
- Show relative time in the DX panel with a UTC time in parentheses (e.g., `5m ago (12:22z)`), updating every minute.
- Add i18n keys for relative time labels in English and German, and stub English defaults for all remaining language files.


ceotjoe:
I was also looking into this today.  :-)

If I understood that correctly you implemented flrig within the backend API. 

I'm wondering if that's a good idea. Currently it looks like that many people might use openhamclock simply from openhamclock.com or another shared source without local setup of the API backend. 

It should be possible to use the tuning of the rig only using UI not via the backend API. Somehow like Wavelog implemented it. 

I think it would be a good idea that the project decides about the implementation architecture of rig (possibly also rotor) control, before it's actually getting into main as part of the backend API. Means how local hardware get's controled via openhamclock.

The solution I had was some kind a helper (nodejs service) which acts as middleware between the UI and flrig. This works both ways. Means frequency get's updated to flrig and openhamclock UI is also receiving updates from the rig via flrig. So, that there can be a plate in the UI showing current set rig frequency for example. Status of PTT could also be taken and used for a possible "On Air" message within the UI.

I just wanted to bring my thoughts also in the discussion, don't take me wrong. :-)

rfreedman:
I have two comments:

1. I agree with @ceotjoe that it would be better to enable interfacing to rig and rotor control (flrig / rigctl) without running the back-end server locally. Perhaps a local proxy that could be run to interface to these and other "local" interfaces.

2. As far as the current commit - I'd like to see the top-level of this configuration changed, though until the scope of supporting local interfaces is settled, I don't have a suggestion for a new name. But I think that it's not a good idea to configure rigctl under a configuration item named "flrig", as these are two "competing" pieces of software. I think that the two configurations should be separated, and named appropriately.

Also, great work, @echo-gravitas! Thanks!

ceotjoe:
I let my coding assistant (currently testing Antigravity) develop a little PoC, which works already pretty well. If you like, you can try here: 

https://github.com/HAMDevs/openhamclock/tree/feat/rig-control-final

Basically it's currently a nodejs deamon (in future this could also be some binary for easier setup) as middleware between openhamclock UI and flrig. Daemon and readme is here: https://github.com/HAMDevs/openhamclock/tree/feat/rig-control-final/rig-control

Long term this could be in an own repository. I think same approach can also be used attaching other hardware.

There is also a plate named rig control (can be added in dockable mode) in openhamclock. which shows current frequency and mode of the radio. 

Clicking on DX-cluster, POTA and map updates frequency and mode (if it works :-) ). Hitting PTT on radio shows "On AIR" in UI.  

Openhamclock implementation should be in a way that it can be easily used by future plates (plugins) also. 

There is also some logic to get the mode out of some bandplan logic. But this is really a bit rough momentary.

Currently it's only a PoC so bare with me if something is not working properly.

rfreedman:
Looks good.me

BTW, I wrote something for myself a while back to, among other things, show
a gauge-style S meter based on rigctl
The repo is at https://github.com/rfreedman/python-rig-client
Feel free to incorporate that code if you're in the mood to add the S meter
:-)

One caveat - I've only ever used this with my IC-7300, so I'm not sure if
the signal strength value interpretation works correctly for other radios.

------------------------------
Rich Freedman
N2EHL

never-rain:
First of all, thank you all for the thoughtful and detailed feedback — I really appreciate the time and care everyone has put into reviewing and discussing this.

My current PR is mainly intended to document the existing rig control capabilities (FLRig / rigctl) and to establish a clear baseline (API visibility, configuration, formatting).

**I fully agree that the current approach is still backend-centric and therefore not ideal for all usage scenarios — especially for users who run OpenHamClock purely via openhamclock.com without a local backend.**

A possible way forward could be:
- Merge this PR as a first step, providing a documented and working baseline.
- Then work together on a more decoupled solution, e.g. a small local proxy/middleware that:
  - talks locally to FLRig or rotctl
  - and can be accessed by the OpenHamClock UI (including when used from openhamclock.com).

This would allow users without a local backend to still control their rigs locally, without forcing rig control logic into the core backend architecture.

If we agree that this PR should serve as a baseline, I’m happy to take care of resolving the current merge conflicts accordingly.

If, on the other hand, the consensus is that this direction is not desirable at all, I’m also fine with withdrawing the PR and revisiting the topic in a different form.

Thanks again for the great feedback!

ceotjoe:
@echo-gravitas I just updated the links to my PoC in my comment above. Probably you can have a look and tell me your thoughts. As I said it's just a PoC. It's already adapted to the massive architecture changes in front end. BTW sorry that I hijacked your PR, but I just was about to drop my own one, but this makes no sense without a basic architecture decision in that matter. So I searched for an existing one and this was yours. :-)

From my perspective we should at least discuss early about how such features can be integrated, that they can find their place in architecture. Otherwise it is always a complete rework to fit them again into huge UI changes.

never-rain:
@ceotjoe - thanks for pushing this discussion forward! I’ve now tried your implementation, and I agree that it’s very promising.

At the moment, it works well for me when the transceiver, FLRig, and OpenHamClock are all running on the same machine.

In my shack setup, however, things are slightly different:
- FLRig and OpenHamClock run on one machine, which is physically connected to the transceiver.
- The OpenHamClock UI is running on a different machine.

The solution I proposed in this PR works in that distributed setup as well, i.e. even when the UI is not running on the same host as OpenHamClock itself.

Conceptually, I would actually prefer your approach, if it can be extended to also support this scenario, where server/backend and UI are running on different machines.

So from my perspective, the key question going forward would be whether your solution can be adapted to cleanly support a separated server/UI setup. If that’s feasible, I think it could be a very strong direction.

never-rain:
@ceotjoe - One additional aspect I’d like to mention: the implementation I proposed also includes the option to send a tune command after changing frequency and operating mode.

This can be quite helpful for transceivers with an internal or external tuner, especially when operating multiband antennas. After a QRG or mode change, a tune cycle can be triggered automatically.

With the original HamClock setup, I had to manually initiate the tuning process on each band change. Having this as an optional feature could improve usability, while keeping it entirely configurable for those who don’t need it.

ceotjoe:
@echo-gravitas technically it should be no problem to have flrig, rig-control-daemon and UI on a different machines. All communicate via network connections. If this all happens in one local network it should work, from my perspective. But I did not try yet. Also daemon IP needs to be configurable within the UI and also the port it uses. As I said it's just a PoC until now. 

The daemon itself has also no dependency on openhamclock. Theoretically this can be a different repo. 

Problem with modes and possibly also the tuning stuff might be that it is different with each rig. Flrig expects correct modes being sent for the attached radio (it does no translation or so). Probably all rigs have different strings as mode names. I can only test on my Yaesu momentary.

Regarding tuning:

-  Is it good practice to tune on a frequency which is currently in use? I always tried to find a free one before. :-)

Frequency should work on most rigs I think, but modes and other stuff are a bit tricky probably.

@accius, @rfreedman  what do you think in general, looking into both solutions? Also shall the daemon being kept within the openhamclock repo or do you prefer having it extra?

ceotjoe:
@echo-gravitas I just pushed a new version where you can enable and disable rig control within UI settings and also enter a individual IP address and port. Of course this also needs to be changed in daemon config file accordingly. Probably you can test that in your setup. Would be great to receive feedback. :-)

never-rain:
Thanks for the update! I agree that your solution should work cleanly if the backend URL can be configured in the UI.
At the moment, the UI calls localhost:5555, which breaks as soon as the backend/server runs on a different machine.

Regarding setting the correct mode and tuning:
I tested this in my implementation with several Yaesu and ICOM transceivers via XML-RPC. FLRig was able to translate the commands correctly, depending on the selected rig configuration in FLRig.

I would see automatic tuning as an optional feature. At least ICOM, Yaesu, and Kenwood radios tune exactly on the configured frequency and typically use <10 W output power. Standard low-cut filters usually start around 50–300 Hz, so the tuning signal is not picked up by the receiver audio chain.

That said, when selecting CW segments on a band, automatic tuning should be disableable, as it could potentially interfere with other stations. For that reason, I’d clearly expose this as a configurable option in the settings, not as default behavior.

Thanks again for the update. I’ll try this during the day tomorrow and will get back to you with feedback.

ceotjoe:
> Thanks for the update! I agree that your solution should work cleanly if the backend URL can be configured in the UI. At the moment, the UI calls localhost:5555, which breaks as soon as the backend/server runs on a different machine.

That's what I wrote in my last comment. You can configure now the IP and port used within the UI settings (section Rig control) and also the daemon config file. I also added tuning option.

UI just sends that daemon should tune and in such case the daemon will send tune command to flrig after 3 seconds it updated the frequency. I also added a simulation mode which only simulates commands to flrig and logs that in console log of the daemon.

If you change settings in UI (IP. Port, tuning option) you need to reload the UI in the browser.

In case you want to try. Simply pull the repo I linked above again and you should have the current code, with all features I mentioned.

Looking forward to your feedback.

auroraonlineinc:
The original HamClock used flrig to interact with the UI as well when the radio transmitted the call sign would change to ON THE AIR.  Currently with WSJT-X when you transmit you see a small TX in red.
Being able to control your rig via OHC would be ideal.  Currently If I want to call someone In a specific location I need to copy it down and then enter it into WSJTX and hope that I get them.

auroraonlineinc:
> @echo-gravitas I just pushed a new version where you can enable and disable rig control within UI settings and also enter a individual IP address and port. Of course this also needs to be changed in daemon config file accordingly. Probably you can test that in your setup. Would be great to receive feedback. :-)

Where is this setting to enable/disable FLRIG ?

accius:
I would like to look at this further, but I am going to close this pull request for now.

ceotjoe:
@echo-gravitas , @auroraonlineinc lets continue discussion and stay connected here. I opened a discussion first before opening another PR. @accius seems busy with other topics currently.  

I also worked on a more lean implementation yesterday evening. Links to current branch are mentioned in the discussion article.

Goto this discussion: https://github.com/accius/openhamclock/discussions/306

PR #244, opened 2026-02-08
Reverse VOCAP reliability chart colors and frequencies to more closely match the original HameClock

dmazan:
Reverse colors (green=good, red=bad) and frequencies/bands (low freq at bottom) so that the VOACAP chart matches more closely the original HamClock.

### DIFF
--- a/src/components/PropagationPanel.jsx
+++ b/src/components/PropagationPanel.jsx
@@ -47,11 +47,11 @@ export const PropagationPanel = ({ propagation, loading, bandConditions }) => {
   
   // Heat map colors (VOACAP style - red=good, green=poor)
   const getHeatColor = (rel) => {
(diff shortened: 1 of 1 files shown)


accius:
i am working on a different approach and i am going to deny this merge for the moment. voacap standard is the opposite, so even tho hamclock did it, doesnt mean we have to.  I will be adding a switch so people can choose heatmap color or stoplight style color for voacap

dmazan:
Yes, that's definitely a better solution.  One additional thing to consider is whether to make it exactly like the original HamClock and also have the left-most edge start at the current hour and run for 24 hours, as opposed to the current where the left edge starts at 0000 and the current hour is highlighted. 

I don't have a strong feeling about this one, unless you decide to name the switch "VOACAP_LIKE_HAMCLOCK" or similar.

PR #257, opened 2026-02-08
Separate render modules

dmazan:
Initial, partial effort at the issue described in bug https://github.com/accius/openhamclock/issues/212. DX Location render code was duplicated in App.jsx and DockableApp.jsx. The recent enhancement to add short and long path bearings and distances to the DX Location panel only appeared in the App.jsx version (Modern theme) and did not make it into the Dockable theme handed by DockableApp.jsx.

DE Location and DX Location are fairly easy as those panels only appear in the Modern and Dockable themes. Created new modules renders/renderDELocation and renders/DXLocation. Replaced in-line code in App.jsx and DockableApp.jsx with calls to the renders/ modules.

And changes to either the DE or DX Location panels can now be made one time in the appropriate renders/ module and will appear in both the Modern and Dockable themes.

This approach should probably be repeated for other panels but may be more difficult as some appear in multiple themes. It might be best to have App.jsx just call the selected theme and have a separate module for each theme to compose the appropriate renders/ modules.

** Reworked to avoid conflicts with layout refactior **

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -22,6 +22,9 @@ import {
   AnalogClockPanel
 } from './components';
 
+import DELocationPanel from './renders/renderDELocation.jsx';
+import DXLocationPanel from './renders/renderDXLocation.jsx';
+
 import { loadLayout, saveLayout, DEFAULT_LAYOUT } from './store/layoutStore.js';
 import { DockableLayoutProvider } from './contexts';
 import './styles/flexlayout-openhamclock.css';
@@ -122,7 +125,7 @@ export const DockableApp = ({
   });
 
   useEffect(() => {
-    try { localStorage.setItem('openhamclock_panelZoom', JSON.stringify(panelZoom)); } catch {}
+    try { localStorage.setItem('openhamclock_panelZoom', JSON.stringify(panelZoom)); } catch { }
   }, [panelZoom]);
 
   const ZOOM_STEPS = [0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.5, 1.75, 2.0];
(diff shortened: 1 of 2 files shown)


dmazan:
Redone PR to include recent changes.  Main point is that it eliminates a buch of duplicate code, the main impact of which is that a number of recent feature additions only appear in the Modern layout and are not available in the Dockable layout.  Both themes now use the same rendering code to there is only one place too add features in the future.

accius:
Hey @dmazan — appreciate you tackling this. The underlying problem is 100% real and it's already caused bugs (the short/long path bearings not showing up in Dockable is exactly the kind of thing that happens when render logic is duplicated across themes).

The direction here is right — shared render modules that both themes import is where we should be heading. That said, I have some concerns about the PR as-is:

- **50 commits / +4,358 lines is a big surface area to review.** At our scale (2K+ concurrent users on the hosted site), I need to be really careful about large diffs introducing regressions. Would you be open to squashing this down and limiting the scope to just the DE/DX Location panels you described?

- **Incremental > monolithic.** I'd love to merge the DE + DX Location extraction as a clean, focused PR, then follow up with additional panels one at a time. Easier to review, easier to revert if something breaks, and other contributors can help with the remaining panels once the pattern is established.

- **If there are any formatting/whitespace changes mixed in**, could you strip those out? We just added Prettier with pre-commit hooks so formatting will be handled automatically going forward — keeps the diff focused on the actual logic changes.

If you can tighten this up to just the two panels with a squashed commit history, I'd be happy to get it merged. The architecture you're proposing in the longer term (themes composing shared render modules) is solid — just want to get there incrementally.

Thanks for putting the work in on this 🤙
73 de K0CJH

dmazan:
This PR was originally six commits to four files.  It kept getting left behind as later PR's were merged.  I've redone it twice to make it easy to merge.  All of those other commits are just merges or PR's that came after mine but were merged first.  There are no formatting deviations from me.  At this point, you are on your own.

accius:
cool

PR #338, opened 2026-02-11
Make SOTA Spots work

alanhargreaves:
Added functionality to maintain a daily updated cache of SOTA detail in server. js. If non-null, this cache adds summitDetails in to each spot for use in the client. This cache is consulted when the the /api/sota/spots is used and populates the summitDetail into each spot. This detail contains things like the lat/long.


alanhargreaves:
The client code already looks for summitDetails. The incorporation of this change gives the SOTA client the ability to place marks on the Worldmap.

alanhargreaves:
Once we've got this change merged that fixes the basic functionality, I'll add the extra button to bring it in to line with what has been done in POTA and WWFF with regard to displaying callsigns in a new merge.

PR #350, opened 2026-02-11
Modifications to WorldMap.jsx and Main.css

creinemann:
Modifications to WorldMap.jsx and Main.css for night time view and GIBS "city lights"
Corrected a syntax error


creinemann:
Do Not Pull, adding an option to make it selectable

Issue #352, opened 2026-02-11
[BUG]  DX lock buttons hidden or obscuring other buttons in Classic and Tablet layouts


denete:
## Describe the Bug
The DX lock button is hidden in classic layout and it obscures other buttons in tablet layout.

## To Reproduce
1. Go to classic layout
2. DX lock button is hidden on the world map (also the settings button is hidden)
3. Go to tablet layout
4. DX lock button is obscuring show / hide satellite tracks on the world map

## Expected Behavior
Buttons should be visible and able to be selected.

## Screenshots
Classic layout
<img width="331" height="184" alt="Image" src="https://github.com/user-attachments/assets/929f2cf5-b487-4a71-bf8b-c0b93bae9db6" />

Tablet layout
<img width="429" height="195" alt="Image" src="https://github.com/user-attachments/assets/23d04e22-5ab3-46ef-9171-2aa8b83a9e71" />

## Environment
- **OS**: Windows 11
- **Browser**:Brave 1.86.148
- **Node.js Version**: v25.6.1
- **OpenHamClock Version**: 15.2.11
- **Running As**: Web browser

## Additional Context
I have already written a fix and will submit a pull request.

---
73!

accius:
fixed in next update

PR #365, opened 2026-02-12
[BUG] DX lock buttons hidden or obscuring other buttons

denete:
Addresses https://github.com/accius/openhamclock/issues/352

DX Locked / Unlocked button is now visible and able to be used. Tested with all layouts and all themes.


denete:
@accius, hello. Could you provide some feedback on why this was closed and not merged? Not an issue, of course, but I would like to contribute to this project if possible. Thanks. - Dee

PR #357, opened 2026-02-11
[FEATURE] Rig Control Integration

ceotjoe:
# Feature: Rig Control Integration
## Overview
This PR introduces **Rig Control** capabilities to OpenHamClock, allowing the dashboard to directly interact with your amateur radio transceiver. Users can now visualize their rig's status (VFO frequency, mode, PTT) and control it directly from the web interface.
### 1. Real-Time Rig Status
*   **Frequency Display:** Shows current VFO frequency in real-time.
*   **Mode Indication:** Displays operating mode (USB, LSB, FM, etc.) and filter width.
*   **PTT Status:** Visual indicator when the radio is transmitting.
### 2. Interactive "Click-to-Tune"
*   **Map Integration:** Clicking any active spot on the World Map (DX Cluster, POTA, SOTA, or PSKReporter) automatically tunes the radio to that frequency and mode.
*   **Workflow Efficiency:** Instantly chase spots without manually dialing the VFO.
### 3. Integrated Controls
*   **PTT / Tune:** Trigger the radio's PTT or internal tuner directly from the Settings panel.
*   **Mode Switching:** Change operating modes via the UI.
## Architecture
The system uses a lightweight **Rig Control Daemon** ([rig-control/rig-daemon.js](cci:7://file:///Volumes/T7%20Shield/Development/openhamclock/rig-control/rig-daemon.js:0:0-0:0)) that runs alongside OpenHamClock.
*   **Bridges Protocols:** Connects to standard rig control software (**flrig** via XML-RPC or **rigctld** via TCP).
*   **Low Latency API:** Exposes a local HTTP API for the OpenHamClock frontend.
*   **Bandwidth Optimized:** Uses **Server-Sent Events (SSE)** to push updates to the browser only when the rig state changes, ensuring minimal network overhead (~0 KB/min when idle).
## Configuration
Rig control is disabled by default. It can be enabled in **Settings > Station Settings > Rig Control**:
*   **Host/Port:** Point to your local daemon (default `localhost:5555`).
*   **Tune Button:** Optional "Tune Enabled" checkbox to allow triggering the radio's ATU.
## Verification
*   Tested with **flrig** (IC-7300, FT-991A) and **rigctld** (simulation).
*   Verified correct frequency/mode parsing and PTT triggering.

Sorry for the huge PR, but integrating that feature needs some changes to existing code. On the other hand I ran prettier on the changed files, so there are also changes indicated because of that in areas I did not modify the logic itself. But they look prettier now. ;-)

[...]

PR #372, opened 2026-02-12
PstRotatorAz UDP integration, Rotator panel, and map bearing overlay

Delerius:
Adds full PstRotatorAz UDP rotator integration to OpenHamClock.

Backend:
- UDP listener for PstRotatorAz azimuth reporting
- /api/rotator/status endpoint
- Stale detection and safe fallback handling
- Environment-based configuration

Frontend:
- New Rotator panel with compass, azimuth input, and STOP control
- Shift-click on World Map to turn rotator toward clicked bearing
- Optional map bearing overlay (toggleable)
- Clean failure behavior if rotator is offline
- Configurable via .env and Vite proxy

Additional:
- POTA panel Call Sign ON/OFF toggle
- Minor cleanup and logging removal
- No breaking changes; feature is fully optional


ceotjoe:
Just a thought of mine. 

Could rotator integration and rig control (https://github.com/accius/openhamclock/pull/357) not use same integration strategy? Means using external service (like rig control daemon) to communicate with rotator. Which is connected via SSE to consuming UI component. This would also work if the end user is not running an own instance of the backend and only uses openhamclock.com. Basically this could easily be integrated into rig control daemon.

Delerius:
> Just a thought of mine.
> 
> Could rotator integration and rig control (#357) not use same integration strategy? Means using external service (like rig control daemon) to communicate with rotator. Which is connected via SSE to consuming UI component. This would also work if the end user is not running an own instance of the backend and only uses openhamclock.com. Basically this could easily be integrated into rig control daemon.

Thanks for the suggestion — that’s a good architectural direction.

Yes, rotator and rig control could absolutely share the same integration strategy via an external control service (e.g. a rig/rotator daemon) that handles hardware communication locally and exposes telemetry to the UI via SSE (or WebSocket) plus HTTP command endpoints.

For V1, the rotator integration was intentionally implemented using the existing backend for simplicity and to validate the UX (panel, map interaction, overlay, stale detection, etc.) without introducing additional moving parts. It keeps the feature fully optional and works for local installs out of the box.

That said, consolidating rig and rotator control into a single external service would offer several advantages:

Unified hardware abstraction layer

Real-time push updates via SSE instead of polling

Cleaner separation between UI and station control

Potential support for hosted openhamclock.com users via a local bridge service

The current implementation is already structured in a way that would allow swapping the transport layer later (e.g. replacing polling with SSE without major UI changes).

This is definitely something worth exploring for a future iteration (V2+), especially if rig control integration expands.

Appreciate the architectural input

auroraonlineinc:
Just curious, what rotator are you using?

Delerius:
> Just curious, what rotator are you using?

I installed a [ERC Version 4 USB ASSEMBLED] board (https://www.vibroplex.com/contents/en-us/p230.html) into my Yaesu G-450A controller. I then use PstRotatorAz software to control it. Doing the integration the way I did allows a very wide choice of rotators to be used with OHC.

auroraonlineinc:
Thanks, that seems to be the standard out there.

PR #402, opened 2026-02-13
Improve DX Cluster custom telnet flow, spot-time handling, and DX panel relative time

echo-gravitas:
## PR Summary

This PR improves DX Cluster behavior end-to-end, especially for `source=custom`, and adds relative spot time display in the DX panel.

### 1) Fixed custom endpoint hangs
- Fixed a Promise resolution bug in `tryDXSpiderNode()` that could leave `/api/dxcluster/paths?source=custom...` hanging indefinitely on connect/timeout/error paths.
- Ensured all code paths resolve cleanly.

**File:** `server.js`

---

### 2) Added `.env` support for custom telnet defaults
- `source=custom` now supports backend defaults from environment/config:
  - `DX_CLUSTER_HOST`
  - `DX_CLUSTER_PORT`
  - `DX_CLUSTER_CALLSIGN`
- Frontend query values still override defaults when provided.
- Callsign resolution now follows:
  1. request callsign  
  2. `DX_CLUSTER_CALLSIGN`  
  3. `CALLSIGN-56`  
  4. `GUEST`

**Files:** `server.js`, `.env.example`

---

### 3) Improved DX spot parsing (Telnet lines or HTML table format)
- Added parsing support for DXSpider `sh/dx` table output (not only `DX de ...` lines).
- This significantly increases parsed spots for cluster nodes that return table-style data.

**File:** `server.js`

---

### 4) Kept spots without coordinates in list data
- Spots are no longer dropped when station coordinates cannot be resolved.
- Map rendering still filters to valid coordinate pairs, but the DX list can now show non-mappable spots.

**File:** `server.js`

---

### 5) Reworked `source=custom` to persistent telnet session
- Replaced reconnect-per-poll behavior with persistent custom sessions:
  - per host/port/callsign session key
  - keepalive + reconnect handling
  - in-memory rolling spot buffer
- `sh/dx 25` is sent once per session lifecycle (initial snapshot), not on every frontend poll.
- `set/dx` is sent on connect to keep live stream updates.
- Disabled HamQTH fallback when `source=custom` is explicitly selected.

**File:** `server.js`

---

### 6) Added explicit command logging
- Added console log lines when `sh/dx 25` is sent, including host/port/callsign.
- Uses `logInfo`, so visible with `LOG_LEVEL=info` or `debug`.

**File:** `server.js`

---

### 7) Sorting now uses spot time (`HHMMz`) as primary source
- Added explicit conversion of spot-provided `HH:MMz` to UTC timestamp (with day rollover handling).
- Path entries now use this time as authoritative for ordering (fallback only if unavailable).

**File:** `server.js`

---

### 8) DX panel: relative time display
- Added relative time display in DX Cluster panel based on spot time.

**File:** `src/components/DXClusterPanel.jsx`

---

### 9) i18n updates for relative DX time
- Added `dxClusterPanel.relativeTime` key to all language files.
- German uses German phrasing; all other languages use English fallback text as requested.

**Files:**
- `src/lang/de.json`
- `src/lang/en.json`
- `src/lang/es.json`
- `src/lang/fr.json`
- `src/lang/it.json`
- `src/lang/ja.json`
- `src/lang/ko.json`
- `src/lang/nl.json`
- `src/lang/pt.json`
- `src/lang/sl.json`

---

### 10) Tooling and consistency updates
- Added `.editorconfig` to enforce consistent base formatting across editors/IDEs.
- Added `.prettierrc` to define a shared formatting standard for the codebase.
- Removed `package-lock.json` from `.gitignore` so lockfile changes are tracked in Git.

---

### 11) Dependency reproducibility policy
- Lockfile is now part of version control to ensure deterministic dependency resolution across all contributors/CI.
- Recommended install workflow is `npm ci` (instead of `npm install`) to guarantee everyone uses the exact locked dependency tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "i18next-browser-languagedetector": "^8.2.0",
     "mqtt": "^5.3.4",
     "node-fetch": "^2.7.0",
+    "papaparse": "^5.5.3",
     "react-i18next": "^16.5.4",
     "satellite.js": "^5.0.0",
     "ws": "^8.14.2"
--- a/server.js
+++ b/server.js
@@ -2659,20 +2659,16 @@ async function checkSummitCache() {
     logDebug('[SOTA] Refreshing sotaSummits');
     const response = await fetch('https://storage.sota.org.uk/summitslist.csv');
     const data = await response.text();
-    const rows = data.trim().split('\n');
-    rows.shift(); // discard the title line
-    const headers = rows
-      .shift()
-      .split(',')
-      .map((header) => header.trim());
+    const Papa = require('papaparse');
+    const csvresults = Papa.parse(data, {
+        skipFirstNLines: 1,
+        header: true
+      }
+    );
+
     let summit = {};
 
-    rows.forEach((row) => {
-      values = row.split(',').map((value) => value.trim());
-      const obj = {};
-      headers.forEach((header, index) => {
-        obj[header] = values[index].replace(/"/g, '');
-      });
+    csvresults.data.forEach( obj => {
       summit[obj['SummitCode']] = {
         latitude: obj['Latitude'],
         longitude: obj['Longitude'],
@@ -2681,6 +2677,7 @@ async function checkSummitCache() {
         points: obj['Points'],
       };
     });
+
     sotaSummits = {
       data: summit,
       timestamp: now,


accius:
@echo-gravitas Please take a look again at this, as there are quite a few conflicts.  Sorry you end up being the last PR sometimes.

never-rain:
@accius - Better late than never. 🤓

accius:
thanks. sorry again. I appreciate the understanding

PR #379, opened 2026-02-13
Revert "PstRotatorAz UDP integration, Rotator panel, and map bearing overlay"

accius:
Reverts accius/openhamclock#372

PR #405, opened 2026-02-13
Re-apply Rotator Panel v1 (post-revert) – clean merge with current upstream

Delerius:
This PR re-applies Rotator Panel v1 (original commit 058631c)
after it was removed via revert (#379).

Restores:
- RotatorPanel
- RotatorMapOverlay
- useRotator hook
- Backend UDP listener + /api/rotator endpoints
- Map bearing overlay (toggleable)

Preserves:
- Modular staging
- Updated mapLayers architecture
- SOTA integration
- ID Timer panel
- POTA label toggle behavior

Builds cleanly against current main.
No breaking changes.
Feature remains fully optional.


accius:
@Delerius, i am not sure how the rotator function is supposed to work.  I merged it to main, but i see no way of setting it up or using it (it does not seem to work properly.) can you give this a look?

PR #446, opened 2026-02-15
Fix Rotator usability + local-only safety (addresses setup confusion)

Delerius:
This PR addresses usability and stability concerns around the Rotator feature.

The original issue was that Rotator was merged but:
- There was no obvious configuration path
- Hosted deployments showed no clear indication that it is local-only
- Backend assumptions could cause confusion or errors

What this update does:

Rotator:
- Explicitly labeled as a Local-only integration (Settings → Integrations → Local-only)
- Hidden/disabled automatically on hosted deployments
- Added clear Quick Start instructions
- Requires explicit enable toggle
- No longer attempts backend interaction unless enabled
- No crashes if rotator backend is unavailable

N3FJP:
- Also clarified as Local-only
- Cleaned up UI by moving display controls into Integrations
- Added configurable display window (15 min → 24h)
- Added server-side retention control via N3FJP_QSO_RETENTION_MINUTES
- Prevented rendering errors

Hosted behavior:
- Hosted OHC shows no rotator controls
- No backend calls are attempted
- No runtime errors occur

Local behavior:
- Fully functional when running Node server locally
- Requires PstRotatorAz HTTP endpoint (configurable in .env)
- Setup documented inline

This should resolve the previous confusion about how Rotator is intended to work and ensure safe operation in both hosted and local environments.


alanhargreaves:
Is there a way to disable it when running locally? I'm working on some unrelated stuff and need to turn on debug logging. The rotator stuff is *really* chatty.

Strike that, looks like it was addressed in that batch of 70 of commits a few hours ago.

Delerius:
Note: PR #426 added a basic hosted safety gate for Rotator (local-only panel list + endpointUrl). This PR builds on that by adding explicit Local-only Integrations UX, documentation, and graceful behavior when backends are unavailable

PR #465, opened 2026-02-16
Add HFJ-350M antenna calculator panel

frankenstein91:
This PR introduces a dedicated calculator panel for the popular Comet HFJ-350M (Toy Box) portable antenna. It allows operators to quickly determine the correct antenna configuration without having to consult the manual during a field day.

Key Features:
   * Dual Input: Supports both frequency (e.g., 7.1) and band names (e.g., 40m).
   * Setup Guidance: Displays required coils, jumper terminals, and radial lengths based on the official manual.
   * Precision Calculation: Calculates the exact telescopic whip length (in mm) based on the frequency deviation from the standard values.
   * Visual Indicators: Features progress-bar style length displays for at-a-glance configuration.

Layout Integration:
   * Modern Layout: Permanently integrated into the left sidebar below the propagation panel.
   * Dockable Layout: Available as a new draggable panel ("HFJ-350M Calc" 📏) via the "+" menu.

### DIFF
--- a/src/components/HFJ350MPanel.jsx
+++ b/src/components/HFJ350MPanel.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo } from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 
 // Antenna data from the manual
@@ -180,9 +180,20 @@ export const HFJ350MPanel = () => {
     });
   };
 
+  // Load last input from localStorage on mount
+  useEffect(() => {
+    const savedInput = localStorage.getItem('hfj350m-last-input');
+    if (savedInput) {
+      setInput(savedInput);
+      calculate(savedInput);
+    }
+  }, []);
+
   const handleInputChange = (e) => {
-    setInput(e.target.value);
-    calculate(e.target.value);
+    const value = e.target.value;
+    setInput(value);
+    calculate(value);
+    localStorage.setItem('hfj350m-last-input', value);
   };
 
   const renderBar = (len, maxLen = 1266, color = "var(--accent-blue)") => {


accius:
While I love the idea, I think we are getting too specific with this panel.  I think this could be a tool by itself that you could advertise with Comet.  I am not going to commit this for now.  But I think you have something really cool, and again, would be more practical as it's own tool.

frankenstein91:
Hi @accius,

Thank you so much for the kind words and the positive feedback! I really appreciate you taking the time to review the PR. You are absolutely right that this could be a standalone tool—in fact, the code for this PR is a direct JavaScript port of a Python CLI tool I had already written for myself (https://github.com/frankenstein91/Comet-HFJ-350M-Toy-Box).

The main reason I decided to port it to OpenHamClock was actually a comment from a fellow ham radio operator. When I showed him my standalone tool, he said: "Oh, this is definitely a practical tool for fellow owners! ... provided they actually have a computer with them out in the field."

That really got me thinking. OpenHamClock is often running on dedicated portable displays right next to the transceiver during field days or POTA/SOTA activations. Integrating it here would mean operators have the tuning data right at a glance, without needing to pull out a laptop or phone with a separate app.

I completely understand your concern about this panel being too specific to just one antenna right now. But maybe we could look at it as a starting point? This doesn't have to be limited to the Comet HFJ-350M. The idea could easily evolve into a generic "Portable Antenna Calculator" panel where users can select from various popular manual tuning antennas. The HFJ-350M just happens to be the one I own and was able to write the logic for right now. Would you be open to the idea of a broader, multi-antenna tuning panel if we generalized it a bit more?

Sheilasib:
A calculator for a mobile antenna, really cool. Could you maybe add it as a plugin if it's not directly desired? Unfortunately, Python is not widely used on mobile devices. A website works on every device.

@frankenstein91 @accius

PR #495, opened 2026-02-17
fix: align clock updates to wall-clock second boundaries to prevent d…

ceotjoe:
## What does this PR do?

Replace setInterval(1000) with a self-correcting setTimeout loop that reschedules itself to fire at the next exact second boundary. This ensures the displayed time stays in sync with the NTP-synced system clock without accumulating drift across long sessions.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.  Check time displayed compared to a synchronized source (e.g. NTP) before and after the fix
2. 
3. 

## Checklist

### DIFF
--- a/src/hooks/app/useTimeState.js
+++ b/src/hooks/app/useTimeState.js
@@ -23,15 +23,27 @@ export default function useTimeState(configLocation, dxLocation, timezone) {
   const handleTimeFormatToggle = useCallback(() => setUse12Hour(prev => !prev), []);
 
   useEffect(() => {
-    const timer = setInterval(() => {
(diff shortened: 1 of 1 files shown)


ceotjoe:
Needs more testing as per feedback within issue. Is someone else able to check this? For me it's workin well within Safari on MacOs.

ceotjoe:
I checked again on my side I see no issues. It's exactly doing what it should. Former little lag is gone. I think we can go with it.

PR #513, opened 2026-02-18
WorldMap: Add DX Weather Hover Bubble and Interactive Path Highlighting

Delerius:
This PR enhances the DX Cluster experience on the World Map by adding weather overlays and interactive path highlighting while preserving existing map behavior.

✨ New Features

Live weather “hover bubble” for DX Cluster spots.

Detailed weather popup on click (temperature, wind, humidity, pressure, sun position).

DX path highlights when a spot is clicked.

DX path highlights when hovering callsigns in the DX Cluster panel.

Movable DX marker behavior remains unchanged.

POTA marker popups and hover behavior preserved.

🛠 Improvements

Refactored DX interaction handling for stability.

Implemented path styling updates without rebuilding map layers.

Normalized DX callsign matching for consistent highlight behavior.

Improved separation of hover vs click logic to prevent popup interference.

Result

DX spots now provide immediate weather insight on hover, detailed weather on click, and clear visual path highlighting — improving situational awareness and overall map usability.

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -99,6 +99,7 @@ export const DockableApp = ({
   togglePOTA,
   togglePOTALabels,
   toggleWWFF,
+  toggleWWFFLabels,
   toggleSOTA,
   toggleSatellites,
   togglePSKReporter,
@@ -395,7 +396,7 @@ export const DockableApp = ({
         dxLocation={dxLocation}
         onDXChange={handleDXChange}
         dxLocked={dxLocked}
-
+        onHoverSpot={setHoveredSpot}
         potaSpots={potaSpots.data}
         wwffSpots={wwffSpots.data}
         sotaSpots={sotaSpots.data}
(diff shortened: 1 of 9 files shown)


accius:
@Delerius DX Weather is a sore spot for me.  One of our contributors added DX Weather and I Have been chasing it ever since.  We are making thousands of calls per min. This causes us to be kicked from the weather servers.  I have had to purposely slow it down in order to stop us from overloading the weather server within a few minutes.  I Would like to revisit this in the future after you and I have had a minute to look at this. because although it works for you, it will catastrophically fail when you have 4000 users trying to click weather at a time.  Lets chat.  For now I am going to close this.

Delerius:
I understand your concerns, so I am going to gate DX weather behind a Local Install toggle (off by default) so hosted/public users never generate external calls. That avoids provider rate limits and keeps the feature available for personal installs. We can later revisit a proper server-side cache/proxy.

PR #550, opened 2026-02-20
Add WWBOTA spots

kwirk:
## What does this PR do?

Similar to POTA, SOTA and WWFF, this add World Wide Bunkers on the Air (WWBOTA).

This connects directly to WWBOTA API use SSE for live updates.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Check spots on appear in new WWBOTA dialog, and on the map.

## Checklist

Screenshot From 2026-02-20 18-58-07

Screenshot From 2026-02-20 19-34-09


kwirk:
Opened as draft to check if there is interest in including this, in which case I'll update README and documentation also.

alanhargreaves:
First up, thank you for using the changes I made with ActivatePanel. I was hoping that it would make it easier for other activator type stuff to happen.

I disagree with changing the "stale" to "error". We frequently see places like SOTA go for extended times with no actual updates.

You need to move the fetching of the spots into server.js like we do for the other spots. I would suggest creating an api for 'api/bota/spots'. On the server side it needs to do the least amount of work possible. Leave the filtering and working out of the json in the hooks code, which is what we use to link from the API to the panel code.

alanhargreaves:
The reason for having the api in the server.js is that we have a lot users running this code. If you do it in the client, you instantly add all users on openhamclock.com pulling from the spots server, rather than the handful of folks running servers doing it every couple of minutes. Just looking at the stats now, we have 1803 users online there and have had a peak of 2030. That's a huge hit to throw at any spots server.

kwirk:
Thanks @alanhargreaves for taking a look.

> I disagree with changing the "stale" to "error". We frequently see places like SOTA go for extended times with no actual updates.

My reasoning here was as this is permanent connection via SSE, rather than polling, it's good to see status that connection is in place. And also as WWBOTA is more popular in Europe and pretty new in North America, there may be times on inactivity outside Europe daylight hours, meaning warning of no update may not be indicative of an issue.

But I'm happy to change it to same as other xOTA services if you'd prefer to keep things consistent?

> You need to move the fetching of the spots into server.js like we do for the other spots. I would suggest creating an api for 'api/bota/spots'... Just looking at the stats now, we have 1803 users online there and have had a peak of 2030.

I think we'd be okay with the additional traffic (I'm admin for the WWBOTA API server): as the connections are SSE they are fairly lightweight compared to polling. I'm also conscious that by adding this feature we are asking you to take on the additional load on your server instead, so it may help you by us taking it directly.

Again, I'm happy to also make changes to `server.js` and polling similar to other xOTA services if this would be preferable?

accius:
@kwirk As admin of the WWBOTA, I am fine with it being put on your server for the load.  So if you are offereing, I will take it.  Bandwidth aint cheap.  also if the connection  is bad and  that is why you would want to show error instead of stale, i am fine with that, but it'd have to leave the functions alone in the rest of the tabs... (let them stay stale)

alanhargreaves:
I'm also more than willing to help get this done.

kwirk:
Thanks @accius and @alanhargreaves :grin: Ready for review now.

PR #570, opened 2026-02-22
Revert "DX Weather (local-only): hover overlay + popup injection"

accius:
Reverts accius/openhamclock#542


accius:
This has been reverted as it broke quite a bit of functionality.

PR #573, opened 2026-02-22
skip workflows on forks to reduce notification noise

frankenstein91:
## What does this PR do?

I have adjusted the GitHub workflows; they should no longer run in forks.


frankenstein91:
As requested in https://github.com/accius/openhamclock/discussions/552#discussioncomment-15889927

PR #574, opened 2026-02-22
"did i break it" - yes you did, so i brought it back. :-)

echo-gravitas:
fixes https://github.com/echo-gravitas/openhamclock/commit/4fab8f530c99d3b9f5ab22bebc72ee94d37b5e69

### DIFF
--- a/src/components/DXClusterPanel.jsx
+++ b/src/components/DXClusterPanel.jsx
@@ -2,10 +2,8 @@
  * DXClusterPanel Component
  * Displays DX cluster spots with filtering controls and ON/OFF toggle
  */
-import React from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


never-rain:
@ceotjoe  - let me check that quickly. I didn't noticed anything suspicious while testing.

ceotjoe:
@echo-gravitas I just looked into the diff.

never-rain:
@accius 
@ceotjoe 
All good. Click to tune works as expected.
`onSpotClick` is used from DXClusterPanel param.

never-rain:
@ceotjoe Since React 17, JSX no longer requires import React because the new JSX transform handles it automatically via react/jsx-runtime. But hooks like useState are real runtime functions, so they still need to be explicitly imported.

PR #576, opened 2026-02-22
Add N3FJP UDP Bridge (Windows) with installer and documentation

Delerius:
This PR adds an optional n3fjp-bridge/ folder providing a Windows-based UDP bridge for N3FJP Logger.

Overview

The bridge:

Connects to the N3FJP TCP API (default port 1100)

Forwards QSO updates via UDP (default port 12060)

Runs independently of OpenHamClock

Includes installer script and documentation

Does not modify or impact OHC runtime

Included

n3fjp-bridge/src/n3fjp_bridge.py

PowerShell installer

Optional hidden runner (VBS)

README (quick start)

UserGuide (detailed installation)

Example configuration files

Notes

Windows-only (N3FJP Logger dependency)

Fully optional utility

No changes to existing OHC functionality

Re-opened against Staging per updated branch workflow.

PR #567, opened 2026-02-22
Add N3FJP UDP Bridge (optional) with Windows installer + docs

Delerius:
Adds optional n3fjp-bridge/ folder at repo root

No runtime impact to OHC

This is an optional tool and does not change OHC runtime

Includes README + UserGuide + example configs + installer script

Bridge forwards N3FJP TCP API (1100) to UDP (12060)

PR #580, opened 2026-02-22
DX Weather (local-only): restore hover bubble + popup weather for POTA/WWFF/SOTA/WWBOTA

Delerius:
Restores hover bubble overlay + popup weather injection for POTA, WWFF, SOTA, and WWBOTA markers.

Uses consistent hover object shape (call/dxCall/lat/lon/source) and calls attachHoverHandlers() + attachPopupWeather() for each marker type.

Keeps DX highlight behavior intact; marker popups lock highlight when open.

Includes minor DX cluster filter adjustment (see dxClusterFilters.js).

Testing:

Verified POTA/WWFF/SOTA/WWBOTA markers show hover bubble and popup weather.

Verified DX Cluster dots retain hover bubble + popup weather and highlight line behavior.

### DIFF
--- a/scripts/generate-translations.js
+++ b/scripts/generate-translations.js
@@ -31,10 +31,22 @@ const universal = {
   'weather.unit.mi': 'mi',
   'weather.unit.mph': 'mph',
   // Wind directions are universal abbreviations
-  'weather.wind.N': 'N', 'weather.wind.NNE': 'NNE', 'weather.wind.NE': 'NE', 'weather.wind.ENE': 'ENE',
-  'weather.wind.E': 'E', 'weather.wind.ESE': 'ESE', 'weather.wind.SE': 'SE', 'weather.wind.SSE': 'SSE',
-  'weather.wind.S': 'S', 'weather.wind.SSW': 'SSW', 'weather.wind.SW': 'SW', 'weather.wind.WSW': 'WSW',
-  'weather.wind.W': 'W', 'weather.wind.WNW': 'WNW', 'weather.wind.NW': 'NW', 'weather.wind.NNW': 'NNW',
+  'weather.wind.N': 'N',
+  'weather.wind.NNE': 'NNE',
+  'weather.wind.NE': 'NE',
+  'weather.wind.ENE': 'ENE',
+  'weather.wind.E': 'E',
+  'weather.wind.ESE': 'ESE',
+  'weather.wind.SE': 'SE',
+  'weather.wind.SSE': 'SSE',
+  'weather.wind.S': 'S',
+  'weather.wind.SSW': 'SSW',
+  'weather.wind.SW': 'SW',
+  'weather.wind.WSW': 'WSW',
+  'weather.wind.W': 'W',
+  'weather.wind.WNW': 'WNW',
+  'weather.wind.NW': 'NW',
+  'weather.wind.NNW': 'NNW',
   // Plugin layer names that are proper nouns / brand names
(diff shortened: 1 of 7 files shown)


Delerius:
@accius
Just for clarity on the earlier revert, after looking through the history, it appears WWBOTA (#550) landed before DX Weather (#567), and since DX Weather touched shared icon handling logic (but not WWBOTA directly), the revert was likely due to base changes rather than DX Weather itself being invalid.

I’ve now rebased cleanly against the current Staging and scoped the icon logic per marker to avoid overlap going forward.

Delerius:
@accius
I need clarity on merge sequencing moving forward.

DX Weather (local-only) was reverted once due to integration conflicts, and I was asked to target staging. Since then, additional PRs touching WorldMap.jsx have been merged while DX Weather remains open.

When overlapping changes are merged without sequencing guidance, it creates a moving integration target and makes stabilization unnecessarily difficult for everyone.

I’m fully comfortable rebasing and resolving conflicts... that’s part of collaborative development... but I need to understand the intended merge order so I’m not repeatedly adjusting against shifting ground.

If there are architectural concerns about DX Weather’s direction, I’d appreciate direct feedback. Otherwise, I’d prefer clearer coordination when related files are involved.

I’m committed to contributing in alignment with the project... I just need predictable workflow to do that effectively.

Delerius:
@accius
This PR now shows a merge conflict in WorldMap.jsx after recent merges; I’m holding off on conflict resolution until we align on sequencing.

accius:
Hey, appreciate the work on the hover/popup weather feature — the actual map marker changes and the DX highlight refactor are solid.

I want to address the conflict issue directly because I think there's a straightforward fix.

The merge conflicts you're hitting aren't really a sequencing problem — they're a scoping problem. This PR has two commits: the feature work (WorldMap.jsx, dxClusterFilters.js), and then a blanket Prettier pass that touches APRSPanel, PSKFilterManager, SettingsPanel, DockableApp, generate-translations, useFloods, and useWildfires. None of those files have anything to do with hover overlays or popup weather. Every one of them becomes a conflict surface against active Staging work, and that's what keeps putting you in rebase loops.

Same thing happened with #576 — the n3fjp-bridge directory was clean and isolated, but the Prettier formatting on existing files created conflicts against the weather units work we'd just landed.

The fix going forward is simple: keep feature PRs scoped to the files your feature actually changes. If you want to submit a formatting-only PR separately, we can look at that on its own, but mixing formatting with feature work makes both harder to review and harder to merge.

If you strip the Prettier commit from this PR and rebase just the feature files against current Staging, I expect this will merge cleanly or close to it. The WorldMap.jsx conflict is legitimate shared territory and I'm happy to work through that with you — but the other six files don't need to be in the picture.

As far as merge order goes, I'm not going to freeze Staging around open PRs. That's not how this works. The best way to stay mergeable is to keep your diff small and focused. The tighter the scope, the less you're exposed to.

73

Delerius:
@accius
You’re right on scope. After rechecking my local history, I found the root cause: early on I had a fork/clone/origin misconfiguration, and I unintentionally carried along formatting/noise changes that shouldn’t have been in the PR. That’s on me.

I’m going to fix this by rebuilding the PR cleanly off current Staging and keeping it strictly scoped to the feature files only (WorldMap.jsx + dxClusterFilters.js, and any new DX-weather-specific util/component files if needed). I’ll drop the blanket formatting commit entirely and keep unrelated files out of the diff.

PR #591, opened 2026-02-23
Fix SSN display inconsistencies when value is 0 (Solar Panel vs Header) #586

### LINKED ISSUE #586: [BUG] Sun Spot Number Incorrect
## Describe the Bug

Sun Spot Number right now is zero on Ham Clock and on the web in general, however OHC is saying it is 113 11:38 UTC 23/02/2026. Showing above top of main map screen. This has not updated for over 24 hours.

## Expected Behavior

At least a daily update of SSN should take place.

## Environment

- **OS**: Windows 10 and Raspberry Pi Trixie
- **Browser**: Chrome on windows Chromium on Pi
- **Node.js Version**: No idea on Windows, v20 on Pi
- **OpenHamClock Version**: 15.6.2
- **Running As**: Web Browser

## For Raspberry Pi Users

- **Pi Model**: [e.g., Pi 4 4GB, Pi 3B+]
- **Pi OS Version**: [e.g., Bookworm 64-bit]
- **Display**: [e.g., Official 7" touchscreen, HDMI 1080p monitor]
- **Running in Kiosk Mode**: [Yes/No]

## Console Errors

If there are any errors in the browser console (F12), please paste them here:

```
Paste console errors here
```

## Additional Context

Add any other context about the problem here. Include your callsign if you're comfortable sharing it!

---

73!

echo-gravitas:
## Summary
This PR fixes an SSN display bug where valid `0` values were treated as “missing”, causing incorrect or inconsistent numbers across the UI. Fixes #586.

### What was wrong
The app used `||` fallbacks and truthy checks in several SSN render paths.  
Because `0` is falsy in JavaScript:

- `SSN = 0` was rendered as `--` in the Solar panel.
- In the header, `SSN = 0` fell through to a fallback field (`spaceWeather.data.sunspotNumber`), which can come from a different source/timescale (monthly NOAA dataset), so users could see a different number (e.g. `113`) at the same time.

### Root cause
Mixed data sources + falsy logic:
- `solarIndices.data.ssn.current` can legitimately be `0` (current/day-level source).
- Fallback fields can represent different cadence/semantics (e.g. monthly NOAA).
- `||` and `if (!v)` incorrectly treated `0` as absent.

### What changed
- Replaced SSN `||` display fallbacks with nullish coalescing `??` so `0` is preserved.
- Replaced `if (!v)` with `if (v == null)` where SSN labels are derived.
- Applied the same fix to all relevant SSN displays for consistency.

### Updated files
- `src/components/SolarPanel.jsx`
- `src/components/Header.jsx`
- `src/layouts/ClassicLayout.jsx`
- `src/components/SpaceWeatherPanel.jsx`

### Result
- SSN now correctly displays `0` instead of `--`.
- Header and Solar Panel no longer diverge due to falsy fallback behavior.
- SSN values are now consistent across components for the same underlying data.

### DIFF
--- a/src/components/APRSPanel.jsx
+++ b/src/components/APRSPanel.jsx
@@ -3,9 +3,8 @@
  * Displays real-time APRS station positions with watchlist group management.
  * Supports tagging callsigns into named groups for EmComm and public service tracking.
  */
-import React, { useState, useMemo, useCallback } from 'react';
+import { useState, useMemo, useCallback } from 'react';
 import CallsignLink from './CallsignLink.jsx';
-import { getBandColor } from '../utils/bandColors.js';
 
 const APRSPanel = ({ aprsData, showOnMap, onToggleMap, onSpotClick, onHoverSpot }) => {
   const {
@@ -14,7 +13,6 @@ const APRSPanel = ({ aprsData, showOnMap, onToggleMap, onSpotClick, onHoverSpot 
     connected,
     aprsEnabled,
     loading,
-    lastUpdate,
     watchlist = { groups: {}, activeGroup: 'all' },
     allWatchlistCalls = new Set(),
     addGroup,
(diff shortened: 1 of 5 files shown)


accius:
Looks good!

PR #592, opened 2026-02-23
feat(rig-control): support port=0 as “no port” and clarify setting labels

echo-gravitas:
This PR updates Rig Control port handling to support reverse-proxy setups and makes the behavior explicit in the UI.

### What changed

- **Rig URL construction**
  - `port = 0` now means: **do not append a port** (use host only).
  - Empty or invalid port values now fall back to **`5555`**.
- **Settings behavior**
  - Rig port values are normalized consistently on load/save.
  - Users can explicitly enter `0` to force host-only URLs.
- **Translations**
  - Updated `station.settings.rigControl.port` in all `src/lang/*.json` files to include:
    - `(0 = no port)`

## Why

Some deployments use public HTTPS endpoints (e.g. reverse proxy / router forwarding) where appending `:5555` breaks access. With this change:

- default/local use stays simple (`5555`),
- advanced users can explicitly opt into no-port mode via `0`.


accius:
clean PR thank you!

Issue #596, opened 2026-02-24
[FEATURE] Show DX Station Local Time in place of UTC time in DX Panel


ws5bsatrustee-tech:
## Use Case:  It is always beneficial to know what the time of day is when working another station.  Admittedly this is purely a convenience function, but would very nice to have.

How would this feature benefit amateur radio operators? Be specific about the use case:

- [X] General HF operation
- [X] VHF/UHF operation
- [ ] Satellite operation
- [ ] Contesting
- [X] POTA/SOTA
- [X] DXing
- [ ] Emergency communications
- [ ] Other: \***\*\_\_\_\*\***

## Would you be willing to help implement this?

- [ ] Yes, I can submit a PR
- [X] Yes, I can help test
- [ ] No, but I'd use it!

## Additional Context

Add any other context, screenshots, or mockups about the feature request here.

## Similar Features in Other Software:  Original HamClock shows LMT in the Map Display Menu dropdown when you hover or click on a location.  My suggestion as described above would, I believe, be a better implementation of the feature.

Thank you for your consideration and for all you have done to make this program excellent.
---
73!
WS5ADV
chumac@att.net or ws5bsatrustee@gmail.com

never-rain:
@ws5bsatrustee-tech - if you're using the Dockable Layout you can switch between UTC and Local Time in the DX Tagret Panel already. In the Modern Layout the the time isn't displayed at all yet.

<img width="452" height="322" alt="Image" src="https://github.com/user-attachments/assets/efd0e05d-10fc-4eed-b74d-4f54a5c299b3" />

ceotjoe:
@echo-gravitas i implemented that for all layouts. that time I guess this was removed by some conflict while latest updates probably. 

See here: https://github.com/accius/openhamclock/pull/494/changes

We just need to reimplement the changes to Modern layout.

never-rain:
@ceotjoe - will do

ws5bsatrustee-tech:
Thank you for the quick reply.  I hope you will do this for the modern layout. The legacy green display is a bit hard on my eyes:)73,Charles L. McBrideWS5ADVWS5BSA On Feb 24, 2026, at 9:30 AM, das Ralph ***@***.***> wrote:﻿echo-gravitas left a comment (accius/openhamclock#596)
@ceotjoe - will do

PR #603, opened 2026-02-24
Implemented parity with Dockable DX Target time behavior in Modern layout

echo-gravitas:
### What was changed
- Added UTC/local time toggle in the **DX Target** panel of the Modern layout, requested in #596, reimplemented what @ceotjoe already implemented in a291a36c53a046df2932efcc0d31c8ed1590076f and PR #494.
- Introduced local component state (`showDXLocalTime`) in `ModernLayout.jsx`.
- Added a clickable time label under the DX grid display:
  - Shows **UTC** by default.
  - Toggles to **local DX time** (`Local UTC±X`) based on DX longitude-derived offset.
  - Uses the same offset calculation and interaction pattern as Dockable (`Math.round(dxLocation.lon / 15)`).

### File modified
- ModernLayout.jsx

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -28,6 +28,7 @@ import {
   RigControlPanel,
   OnAirPanel,
   IDTimerPanel,
+  DXLocalTime,
 } from './components';
 
 import { loadLayout, saveLayout } from './store/layoutStore.js';
@@ -416,11 +417,7 @@ export const DockableApp = ({
         </div>
       </div>
 
-      <WeatherPanel
-        weatherData={localWeather}
-        units={config.units}
-        nodeId={nodeId}
-      />
+      <WeatherPanel weatherData={localWeather} units={config.units} nodeId={nodeId} />
     </div>
   );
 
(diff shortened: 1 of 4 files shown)


accius:
Love the idea — showing local time at the DX location is a great addition. A few things to address before we merge:

**Crash risk:** The IIFE accesses `dxLocation.lon` without a null guard. If no DX station is selected, this will throw. Wrap it in a check or bail out early.

**`currentTime` — is it in scope?** I don't see it coming from props or a hook in this diff. If it's not wired up, this will silently show `NaN:NaN`.

**Timezone approximation:** `Math.round(lon / 15)` gives solar time offsets, which is fine, but it'll be wrong for fractional-offset zones like India (+5:30) or Nepal (+5:45). Can you add a comment like "approximate solar time" so users don't expect civil timezone accuracy?

**Inline IIFE:** The `{(() => { ... })()}` block is hard to follow in JSX and can't be memoized. Could you extract it into a small `<DXLocalTime />` component or at least compute the values above the return statement?

**Scope:** This only touches ModernLayout — is that intentional? If we want it everywhere, might be worth making it a shared component from the start.

**Minor:** The `marginTop: 4px → 8px` change on the DE card coordinates looks like a separate spacing tweak — consider splitting that into its own commit so it doesn't get lost in the feature diff.

None of these are huge — the null guard is the only real blocker. Looking forward to v2! 🔧

ceotjoe:
@accius it is already in Dockable but was somehow removed in Modern. 

@echo-gravitas crash risk is also there in Dockable and should be correct there also then. I would use then same implementation everywhere. Probably the solar time hint could be implemented via ToolTip while hovering or so. I wanted to correct the lost implementation in Modern anyway, but you were faster. :-)

never-rain:
Thanks @accius and @ceotjoe, for the strong feedback.  
You’re absolutely right, and your suggestions were really good.

I updated the change accordingly:

- Extracted the inline time logic into a shared `DXLocalTime` component.
- Applied the same implementation in both `ModernLayout` and `DockableApp` for consistency.
- Added guards for missing/invalid `dxLocation.lon` and invalid `currentTime` to prevent crashes.
- Clarified via tooltip that the DX local time is **approximate solar time** (longitude-based), not a civil/IANA timezone conversion.
- Reverted the unrelated Modern DE spacing tweak so the diff stays focused.
- Exported the new component in `src/components/index.js`.

Could you please take another look and review this version?

accius:
nice work. merged to staging

PR #604, opened 2026-02-24
Minor changes to satellite layer 

creinemann:
## What does this PR do?

Minor Changes to Satellite Logic and Visuals add Signal Acquisition status that bilinks to highlite active sat, and narrowed the 'foot print' of the info window to take up less space on screen when a satellite is selected

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Click on Satellite -window will open with new features
2.
3.

## Checklist

old
change1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openhamclock",
-  "version": "15.5.8",
+  "version": "15.6.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "openhamclock",
-      "version": "15.5.8",
+      "version": "15.6.2",
       "license": "MIT",
       "dependencies": {
         "axios": "^1.6.2",
--- a/src/plugins/layers/useSatelliteLayer.js
+++ b/src/plugins/layers/useSatelliteLayer.js
@@ -5,7 +5,7 @@ import { replicatePoint, replicatePath } from '../../utils/geo.js';
 export const metadata = {
   id: 'satellites',
   name: 'Satellite Tracks',
-  description: 'Real-time satellite positions with multi-select footprints',
+  description: 'Real-time satellite positions with glow tracks and narrow info card',
   icon: '🛰',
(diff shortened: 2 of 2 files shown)


accius:
Nice cleanup — the narrower card, CSS injection, and killing that `location.reload()` on clear are all solid improvements. A few things before we merge:

**Blocker: Revert the version bump.** The `package-lock.json` change from 15.5.8 → 15.6.2 needs to come out of this PR — versioning is handled separately on my end. Please revert that file so it doesn't conflict.

**Variable naming:** The rename from `positionAndVelocity` → `pV`, `futureTime` → `fT`, `geodetic` → `g` etc. hurts readability for other contributors. I'd prefer keeping the descriptive names — saving a few characters isn't worth the maintenance cost in a shared codebase.

**"Signal Acquisition":** The old label said "Visible" for above-horizon sats. "Signal Acquisition" sounds cool but it's not accurate — being above the horizon doesn't mean you've acquired the signal. Let's stick with "Visible" or maybe "Above Horizon".

**Minor (non-blocking):**
- The injected `<style>` tag never gets cleaned up on unmount. Since it's idempotent it won't cause issues, but a cleanup return in the useEffect would be cleaner.
- `window.clearAllSats` is a nice addition over the old reload approach. Just note it re-binds every time `selectedSats` changes since it's inside the same useEffect — could be its own one-liner effect but not a big deal.

Revert the version bump and rename the variables back and this is good to go 👍

creinemann:
I'll have a look at your ideas, sounds good. Some of it I may remove as I was working on adding AOS and LOS by adding a user input near LAT and LON for ALT altitude to compute  them and down the times it would actually happen. I don't think I will be able to make the changes this week. If you want feel free an hop in and change this parts, otherwise it'll be a while for me

Cheers

creinemann:
Moved to new pull request

PR #606, opened 2026-02-24
UI: Custom Dockable Header + Tablet-Safe Drawer Activation

Delerius:
This PR replaces the native FlexLayout tab strip with a minimal custom header + hover drawer, while preserving existing layout behavior.

A follow-up tablet safety pass ensures the drawer works reliably on touch and hybrid devices.

Target branch: `upstream/Staging`

---

### Dockable Header Refactor
- Hide native FlexLayout tab buttons
- Implement compact custom header (~18px height)
- Entire header acts as:
  - Click = next tab
  - Shift+Click = previous tab
- Prevent default FlexLayout double-click maximize behavior

### Drawer
- Appears on header hover (desktop)
- Opens via focus on touch devices
- Semi-transparent background (no blur)
- Slight hover darkening for contrast
- Centered, uniform icon layout

Drawer controls:
- Panels
- A− / A+
- Maximize
- Close

---

## Touch / Tablet Improvements

- Added pointer event handling (instead of mouse-only)
- Enabled `:focus-within` activation so drawer works on tap
- Made header hit-area focusable
- Added coarse pointer CSS adjustments for better touch targets
- Desktop behavior unchanged

---

## Scope

- Modified:
  - `src/DockableApp.jsx`
  - `src/styles/flexlayout-openhamclock.css`
- No changes to layout model logic
- No changes to upstream FlexLayout internals
- Build passes
- No conflicts with `upstream/Staging`

---

## Behavior Notes

- Drawer activation remains hover-based on desktop
- Touch devices open drawer on tap
- Tab cycling only occurs when drawer is visible
- Accessibility maintained via focus handling

---

Tested on:
- Desktop (mouse)
- Hybrid / touch input

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -1026,63 +1026,6 @@ export const DockableApp = ({
         }
       };
 
-      // Header-as-button: when drawer is open (hover), let users click anywhere on the header to cycle tabs.
-      // Click = next tab, Shift+Click = previous tab.
-      // Z-index in CSS keeps drawer buttons clickable above this hit-area.
-      renderValues.stickyButtons.push(
-        <button
-          key="ohc-header-cycle"
-          type="button"
-          className="ohc-header-cycle-hitarea"
-          title="Click to cycle panels (Shift = previous)"
-          aria-label="Cycle panels (Shift = previous)"
-          onPointerDown={(e) => {
-            // Tablet/touch safety:
-            // First tap should OPEN the drawer (via :focus-within), not immediately cycle.
-            // Once the drawer is open (hover or focus-within), tap/click cycles.
-            e.preventDefault();
-            e.stopPropagation();
-
-            const outer = e.currentTarget.closest('.flexlayout__tabset_tabbar_outer');
-            const drawerOpen = !!outer && outer.matches(':hover, :focus-within');
-            if (!drawerOpen) {
-              try {
-                e.currentTarget.focus();
-              } catch {
-                /* noop */
-              }
-              return;
-            }
-
-            cycleTab(e.shiftKey ? -1 : 1);
-          }}
-          onKeyDown={(e) => {
-            const outer = e.currentTarget.closest('.flexlayout__tabset_tabbar_outer');
-            const drawerOpen = !!outer && outer.matches(':hover, :focus-within');
-            if (!drawerOpen) return;
-
-            const key = e.key;
-            if (key === 'Enter' || key === ' ') {
-              e.preventDefault();
-              e.stopPropagation();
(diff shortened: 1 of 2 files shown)


accius:
Hey @Delerius — appreciate the effort here, this is a big piece of work. The drawer animation is slick, the touch support via `pointerDown` is the right call, and force-selecting newly added tabs is a real usability win. A few things to work through before we can merge:

**Blocker: Hiding all tabs removes fundamental navigation.** The CSS blanket-hides every `.flexlayout__tab_button` with `display: none !important`. Users now have zero visual indication of which panels are stacked in a tabset, or that multiple panels even exist. Discovering them requires: hover header → see drawer → click Panels → read modal → click one. That's 3-4 steps for what a visible tab did in 1 click. Anyone who's built a multi-tab layout will think their panels vanished.

I'd suggest keeping a compact tab strip — even just abbreviated names or dots — as the primary navigation, with the drawer as supplemental controls (zoom, maximize, close). Tabs shouldn't disappear entirely.

**`CYCLE` button is cryptic.** A text label saying CYCLE doesn't communicate what it does, and the shift+click for reverse is a hidden gesture with no affordance. If we keep tab cycling, maybe prev/next arrow icons (`◀ ▶`) would be more intuitive.

**CSS fragility — 200+ lines of `!important`.** Nearly every rule fights FlexLayout's internals with `!important`. Any FlexLayout version bump could silently break things. The `.ohc-drawer-cycle` styles are also duplicated (once in the main section and again under the "FIX5" comment). Could we scope these more tightly or work with FlexLayout's API instead of overriding everything?

**`onRenderBorderTabSet={onRenderTabSet}`** — Applying the tabset renderer to border tabsets could break border panel behavior since borders have different layout semantics in FlexLayout. Has this been tested with border panels?

**Minor:**
- The `catch (err) { /* noop */ void err; }` pattern appears 12+ times. `catch (_)` or a single lint disable would be cleaner.
- The Switch view IIFE (`{(() => { try { ... } catch { return null } })()}`) is hard to follow — worth extracting to a component.
- The `allowMaximizeRef` pattern to intercept FlexLayout actions is clever but couples tightly to their internal action type strings.

**What's solid:**
- `pointerDown` over `click`/`mouseDown` — proper touch/hybrid support
- Force-selecting new tabs after add
- `handleSelectTab` as a proper callback
- Drawer animation timing (120ms) feels right
- The Panel picker Switch/Add tabs are a nice concept

The core ask is to find a middle ground where tabs stay visible (even in compact form) so users always know what's in each stack. The drawer can enhance that, but shouldn't replace it. Happy to brainstorm approaches if you want to hop in a discussion thread.

PR #609, opened 2026-02-25
DX Weather (local-only): hover bubble, popup weather & line-only path highlighting

Delerius:
DX Weather (Local-Only): Hover Bubble + Line-Only Path Highlighting

This PR restores and stabilizes the DX Weather enhancements on the World Map, scoped strictly to local installations only to prevent external API load concerns.

Features:

Live weather hover bubble for DX Cluster, POTA, WWFF, SOTA, and WWBOTA spots.

Detailed weather popup on click (temperature, wind, humidity, pressure, precipitation).

DX path highlights when:

Hovering a callsign in the DX Cluster panel.

Opening a DX popup on the map.

Highlighting affects only the DX path line — DX dots remain visually unchanged.

Movable DX marker behavior remains unchanged.

POTA / WWFF / SOTA / WWBOTA marker behavior preserved.

Local-Only Safety:

Feature gated behind isLocalInstall.

Toggle-controlled via local settings.

Weather responses are cached with TTL to prevent repeated calls.

No behavior changes for hosted deployments.

Implementation Notes:

Refactored DX interaction handling for stability.

Path styling updates applied via setStyle() (no layer rebuilds).

Callsign normalization ensures consistent highlight matching.

Hover vs click highlight logic separated to prevent popup interference.

Scope intentionally limited to src/components/WorldMap.jsx.

### DIFF
--- a/src/components/WorldMap.jsx
+++ b/src/components/WorldMap.jsx
@@ -239,6 +239,14 @@ export const WorldMap = ({
     [clearDXHighlight],
   );
 
+  // DX Cluster panel/map hover -> highlight matching DX path (unless locked by popup)
+  useEffect(() => {
+    if (dxHighlightLockedRef.current) return;
+    const key = normalizeCallsignKey(hoveredSpot?.dxCall || hoveredSpot?.call || hoveredSpot?.dx || '');
+    if (key) setDXHighlight(key);
+    else clearDXHighlight();
+  }, [hoveredSpot, setDXHighlight, clearDXHighlight]);
+
   // Expose DE location to window for plugins (e.g., RBN)
   useEffect(() => {
     if (deLocation?.lat && deLocation?.lon) {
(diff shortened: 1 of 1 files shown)


accius:
This is a really nice architectural improvement — removing `hoveredSpot` from the DX paths dependency array eliminates the full-rebuild-on-every-hover problem, and the `dxLineIndexRef` index with in-place restyling is the right approach. The popup lock pattern is clean, and getting hover + weather popups on all four activate spot types is a great addition.

Three small things before merging:

**SOTA popup lost summit/points info.** The old SOTA popup showed summit name and point value (`spot.summit`, `spot.points`), but the new `baseHtml` uses the same generic ref/location/freq template as POTA/WWFF/WWBOTA. Can you add those back? Something like:
```
`${spot.summit ? ` — ${esc(spot.summit)}` : ''}${spot.points ? ` <span style="color:#ff9632">(${esc(spot.points)}pt)</span>` : ''}`
```

**Empty catch blocks in `attachHoverHandlers`.** The three `catch {}` blocks should be `catch (_)` for consistency and clarity.

**Minor:** `attachHoverHandlers` is a plain function inside the component so it recreates every render. Since it's only called during marker setup in useEffects it's not a real issue — just noting it for if we ever refactor.

Fix the SOTA popup and the catch blocks and this is good to go 👍

PR #575, opened 2026-02-22
introduce AddOns ecosystem with APRS and Antenna tools (Userscripts)

frankenstein91:
## What does this PR do?

Idea thanks to: https://github.com/accius/openhamclock/pull/465#issuecomment-3939363234

This PR introduces a dedicated **AddOns** ecosystem for OpenHamClock via **Userscripts** (Tampermonkey/Greasemonkey). It allows users to add specialized, optional features without increasing the core bundle size or forcing API configurations on every user. 

Key additions:
- **HFJ-350M Calculator**: Antenna tool for the Comet multi-band antenna (DE/EN/JA).
- **APRS Newsfeed (Inbox)**: Integrated message center for incoming APRS messages via `aprs.fi`.
- **APRS Auto-Position**: Live coordinates update for mobile/portable users based on APRS beacons.
- **🧩 Shared AddOn Drawer**: A non-intrusive launcher system in the bottom right corner that reveals tools on click and allows layout rotation via right-click.
- **CI Workflow Optimization**: Updated `.github/workflows/*.yml` to skip execution on forks, reducing notification noise for community developers.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [X] Documentation
- [X] Translation
- [ ] Map layer plugin

## How to test

1. Install a Userscript Manager (e.g., Tampermonkey).
2. Load any script from the `AddOns/` directory (e.g., `AddOns/HFJ350M-Calculator/hfj350m_calculator.user.js`).
3. Open OpenHamClock and look for the **🧩 icon** in the bottom right corner.
4. Click to expand tools, and right-click to switch between horizontal/vertical layouts.
5. Verify that the tools adapt to your current theme (Dark, Light, Retro) and language.

## Checklist

- [X] App loads without console errors
- [X] Tested in **Dark**, **Light**, and **Retro** themes
- [X] Responsive at different screen sizes (desktop + mobile)
- [X] If touching `server.js`: N/A (Scripts are frontend-only)
- [X] If adding an API route: N/A
- [X] If adding a panel: Implemented as modular Userscript overlays to keep layouts clean.
- [X] No hardcoded colors — uses CSS variables (`var(--accent-cyan)`, etc.)
- [X] No `.bak`, `.old`, `console.log` debug lines, or test scripts included

## Screenshots (if visual change)
[...]

PR #611, opened 2026-02-25
[AddOns] position of the menu system

frankenstein91:
https://github.com/accius/openhamclock/pull/575#issuecomment-3946233395

Enhanced AddOn Drawer (UX Improvements)
   * Drag-and-Drop: The 🧩 launcher icon can now be moved freely across the screen (Middle-click to drag).
   * Persistent Position: The drawer's position and orientation (horizontal/vertical) are now saved in localStorage.
   * Smart Layout: Added a dynamic layout system that automatically adjusts the drawer's expansion direction (row/column) based on its screen quadrant to prevent icons from overlapping the edges.

### DIFF
--- a/AddOns/APRS-AutoPos/aprs_autopos.user.js
+++ b/AddOns/APRS-AutoPos/aprs_autopos.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         APRS Auto-Position for OpenHamClock
 // @namespace    http://tampermonkey.net/
-// @version      1.1
+// @version      1.2
 // @description  Automatically updates your station position in OpenHamClock based on APRS beacons
 // @author       DO3EET
 // @match        https://openhamclock.com/*
@@ -82,7 +82,7 @@
             transition: all 0.3s ease;
         }
         .ohc-addon-icon:hover { border-color: var(--accent-amber, #ffb432); transform: scale(1.1); }
-        #ohc-addon-launcher { background: var(--bg-tertiary, #1a2332); color: var(--accent-amber); cursor: move; }
+        #ohc-addon-launcher { background: var(--bg-tertiary, #1a2332); color: var(--accent-amber); cursor: move; transition: transform 0.3s ease; }
         .ohc-addon-item { display: none; }
 
         #ohc-autopos-container {
--- a/AddOns/APRS-Newsfeed/aprs_newsfeed.user.js
+++ b/AddOns/APRS-Newsfeed/aprs_newsfeed.user.js
@@ -1,8 +1,8 @@
 // ==UserScript==
 // @name         APRS Newsfeed (Inbox) for OpenHamClock
 // @namespace    http://tampermonkey.net/
-// @version      1.7
-// @description  Fetches and displays your latest APRS messages from aprs.fi
+// @version      1.8
+// @description  Fetches and displays your latest APRS messages (Inbox & Outbox) from aprs.fi
 // @author       DO3EET
 // @match        https://openhamclock.com/*
 // @grant        GM_xmlhttpRequest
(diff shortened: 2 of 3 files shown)


accius:
Nice upgrade — the draggable drawer with position persistence and smart edge-snapping is a solid improvement. A few things to address:

**Bug — missing newline in aprs_newsfeed.user.js:**
After the drawer `if` block closes, the `const toggleBtn` declaration starts on the same line as the closing `}`. JS will parse it, but it's clearly unintended:
```
            setTimeout(updateLayout, 100);
        }        const toggleBtn = document.createElement("div");
```
Needs a linebreak there.

**Code duplication:**
The ~90-line drag implementation (updateLayout, savedPos, isDragging, mousedown/mousemove/mouseup, edge snapping, localStorage persist) is copy-pasted identically into all three userscripts plus the dev guide — roughly 360 lines of the same code. Since the drawer uses a shared-element pattern (`getElementById('ohc-addon-drawer')` → create only if missing), whichever script loads first creates it and the duplicate code in the other two is dead code.

Would be worth extracting the drawer into a standalone shared userscript (e.g. `ohc-addon-drawer.user.js`) that the others can depend on. Not a blocker for this PR, but if the drag logic ever needs a bugfix, it currently has to be patched in four places.

**Middle-click drag discoverability:**
`e.button === 1` means trackpad and touch users have no way to drag the drawer. Left-drag with a small dead zone threshold (e.g. 5px before initiating drag vs treating as click) would be more universal. The tooltip "M: Drag" helps for mouse users, but it's still pretty hidden. Something to consider for a follow-up.

Everything else looks good — the formatting cleanup and markdown tweaks are fine. Fix the missing newline and this is ready to merge. 👍

frankenstein91:
@accius You're absolutely right. The ~90 lines of shared drawer and drag logic are currently duplicated in all three scripts and the guide to  ensure each AddOn remains fully self-contained and 'plug-and-play' for the end user without requiring a 'Core' script to be installed  first.

As you noted, the if (!drawer) check means that only the first script to load actually builds the UI, while the others effectively carry  'dead code' for that specific part. This was a pragmatic choice to get the shared-element pattern working, but I agree that it's not ideal  for long-term maintenance.

If we had a good stable URL... we could load the shared code from there

frankenstein91:
touch screen was already on my list... and the code should now work.

accius:
Great iteration — the extracted `startDrag`/`handleMove`/`stopDrag` helpers are much cleaner, `wasDragged` flag properly prevents click-after-drag, touch support with `{ passive: false }` and vibrate feedback is well done. Newline bug is fixed.

One thing to fix: **the Development_Guide.md example is out of sync with the actual scripts in two spots:**

1. The guide's `launcher.onclick` is missing the `wasDragged` check, so after a long-press drag the click will fire and toggle the drawer.

2. The guide's `onmousedown` for `button === 0` doesn't set `startX`/`startY` before starting the timer. The mousemove handler then checks `Math.abs(e.clientX - startX) > 5` against `undefined`, so the 5px cancel-on-move threshold won't work. All three actual scripts set these correctly.

Both are copy-paste oversights — just sync the guide's example with the pattern from the scripts and this is good to merge. 👍

frankenstein91:
I would like to apologize for including an update to the APRS Newsfeed script in this Pull Request.

While I tried to keep the scope focused, I discovered that the Newsfeed was only displaying classic text messages, completely missing  Mic-E/Position-based beacons with comments (which are very common in mobile APRS use). In my eyes, this was a functional bug that made the  inbox feel incomplete, so I decided to include a minimalist fix for it.

The Newsfeed (v1.8) now supports:
   * Inbox & Outbox: It shows both received and sent messages.
   * Beacon Comments: It now correctly displays status comments from position beacons.
   * Extended SSID Search: It covers a wider range of SSIDs to ensure no activity is missed.

PR #620, opened 2026-02-25
Feature: rig bridge introducing plugin architecture

ceotjoe:
## Summary
This PR brings comprehensive improvements to the rig-bridge, focusing on robust FT-991A support, configuration persistence, and a refined setup experience.
### Key Changes
- **Changed architecture**:
  - Plugin based approach to reduce conflicts while integrating new features
- **More Robust USB Radio Support (Mac & Yaesu)**:
  - Added a `stty` fix for macOS/Linux to prevent serial port hang-ups during hardware flow control.
  - Enabled **Auto-Info (AI1)** mode for Yaesu radios, providing near-instant updates for frequency and PTT changes.
- **Improved Configuration Management**:
  - Added and automatic copy logic to ensure the configuration is preserved during git updates.
  - Fixed a bug preventing auto-connection at startup by ensuring settings are shared correctly across modules.
- **Refined Setup UI**:
  - Integrated logging
  - Fixed a layout bug that could hide the "Save" button for non-USB plugins.
  - Prevented the UI from overwriting saved settings with radio-specific defaults on page load.
  - Added a `--debug` flag to keep the console clean while allowing verbose CAT logging when needed.
## Verification
- Verified on macOS with a Yaesu FT-991A.
- Tested auto-connect and config-persistence logic.
- Validated UI behavior across different radio types (`mock`, `flrig`, `rigctld`, and `usb`).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

As I only tested on Mac with my own rig it would need further testing on Linux and Windows. Especially the USB part. As this was really a pain, because of some bugs or inconsistencies in node serial implementation across OS boundaries.

This is the first PR to convert rig-bridge into a universal solution for bridging to rigs, rotators, UDP QSO collectors etc. as plugins  Probably it get's then OHC-bridge or something.

After this is working completely we ask users to switch from rig-daemon, rig-listener etc. and remove these then.

App is not affecting openhamclock itself only rig-bridge.

@accius I hope you're fine with that approach. 

1. npm start and configure via http://localhost:5555
2. Check if radio comms works on Windows, Linux and the rigs you have.
3.

## Checklist

### DIFF
--- a/rig-bridge/plugins/rigctld.js
+++ b/rig-bridge/plugins/rigctld.js
@@ -20,6 +20,7 @@ module.exports = {
     let pending = null;
     let pollTimer = null;
     let reconnectTimer = null;
+    let wasExplicitlyDisconnected = false;
 
     function stopPolling() {
       if (pollTimer) {
@@ -83,6 +84,7 @@ module.exports = {
 
     function connect() {
       if (socket) return;
+      wasExplicitlyDisconnected = false;
 
       const host = config.radio.rigctldHost || '127.0.0.1';
       const port = config.radio.rigctldPort || 4532;
@@ -105,22 +107,28 @@ module.exports = {
       });
 
       s.on('close', () => {
-        console.log('[Rigctld] Connection lost — retrying in 5 s…');
         updateState('connected', false);
         socket = null;
         stopPolling();
         pending = null;
         queue = [];
-        reconnectTimer = setTimeout(connect, 5000);
+
+        if (!wasExplicitlyDisconnected) {
+          console.log('[Rigctld] Connection lost — retrying in 5 s…');
+          reconnectTimer = setTimeout(connect, 5000);
+        }
       });
 
       s.on('error', (err) => {
-        console.error(`[Rigctld] Error: ${err.message}`);
+        if (!wasExplicitlyDisconnected) {
+          console.error(`[Rigctld] Error: ${err.message}`);
+        }
         s.destroy();
       });
     }
 
     function disconnect() {
+      wasExplicitlyDisconnected = true;
       if (reconnectTimer) {
         clearTimeout(reconnectTimer);
         reconnectTimer = null;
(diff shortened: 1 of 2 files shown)


ceotjoe:
@accius, @echo-gravitas would be nice if you could give it a try in your environment.

accius:
Great work on this refactor — the plugin architecture is a big step forward for maintainability. Reviewed the full diff and I'm merging this. A few notes for follow-up:
Highlights:

Clean decomposition from the monolith into core/ + plugins/. The plugin descriptor contract is simple and well-documented both in code and the README.
Yaesu AI1 auto-info mode is the right call — push beats polling. The 30s keepalive fallback is a nice safety net.
Mock plugin is a great addition for dev/testing without hardware.
macOS cu.* port synthesis and the stty termios fix for CP210x are solid platform-specific improvements that will save people a lot of headscratching.
Console log SSE stream with level filtering in the UI is a really useful debugging tool.

Follow-up items (non-blocking):

Yaesu rtscts is hardcoded true in plugins/usb/index.js regardless of user config. Makes sense for FT-991A but could bite users on other Yaesu models — we should at least log when we're overriding their setting, and consider making this a per-model default rather than a blanket override.
registerBuiltins() will crash on startup if any plugin file is missing (e.g. mock.js). Would be good to wrap each require() in a try/catch so a missing or broken plugin doesn't take down the whole bridge.
--debug flag is checked via process.argv.includes() in individual protocol files. Worth centralizing in config.js and exposing as config.debug for consistency.
Test endpoint swallows DTR/RTS set errors silently — minor, but worth a console.warn at minimum.

Merging as-is, we can address the above incrementally. 🚀

never-rain:
> @accius, @echo-gravitas would be nice if you could give it a try in your environment.

Will do as soon as I can. Sounds great! 🤩

PR #621, opened 2026-02-25
Feature/rig bridge plugin architecture

ceotjoe:
## What does this PR do?

I just fixed the issues mentioned in comments I got for https://github.com/accius/openhamclock/pull/620#issuecomment-3960707690. 

and a bit more meanwhile:

### Summary

This PR completes the rig-bridge plugin architecture refactor by addressing PR feedback, hardening the USB serial plugin, and adding the first non-rig integration plugin (WSJT-X Relay) with a dedicated Integrations tab in the config UI.

---

#### `782b157` — Address PR feedback: centralized debug, robust plugin loading, respect RTS/CTS settings, improve test endpoint logging

- **Centralized debug flag** — removed the module-level `const DEBUG = process.argv.includes('--debug')` from the USB plugin. All verbose CAT logging now reads from `config.debug`, which is set by `applyCliArgs()` in `config.js`. This ensures the flag is parsed once and consistently respected across all modules.
- **Robust plugin loading** — wrapped all `require()` calls in `registerBuiltins()` in try/catch blocks. A missing or broken plugin file now logs an error and continues rather than crashing the registry.
- **Respect RTS/CTS setting** — removed the Yaesu-specific override that forced `rtscts: true` regardless of user config. The port now opens with `rtscts: !!config.radio.rtscts` for all radio types, honouring the setting saved in the config file.
- **Test endpoint logging** — the `DTR/RTS` set call in `POST /api/test` now logs a warning if the set fails, instead of silently ignoring the error.

---

#### `102ad3a` — Final improvements: robust USB port release, standardized variable names, and improved error handling

- **Robust USB port release** — tracked `stabilizerTimer` and `initialPollTimer` as named variables so they are properly cleared in `disconnect()`. Previously, these anonymous `setTimeout` calls could fire after the plugin was stopped, causing writes to a closed port.
- **Race condition fix on disconnect** — if `disconnect()` is called while the serial port's `open()` callback is still pending, the port is now immediately closed and the reconnect loop is suppressed via the `wasExplicitlyDisconnected` flag.
- **Port-in-use error handling** — `startServer()` now captures the `server` reference and attaches an `error` listener. An `EADDRINUSE` error exits with a clear message rather than an unhandled exception.
- **`--help` flag** — added `--help` / `-h` support to `rig-bridge.js` with usage, options, and examples.
- **Standardized variable names** — aligned naming conventions across USB protocol files (Yaesu, Kenwood, Icom) for consistency.

---

#### `d5c9588` — Add wsjtx-relay integration plugin **Needs testing**, split config UI tabs, and add --version flag

- **`plugins/wsjtx-relay.js`** (new) — embeds the full WSJT-X relay logic as a rig-bridge integration plugin: WSJT-X binary protocol parser, batched HTTP sender with adaptive backoff, heartbeat, and health check. Exposes `GET /api/wsjtxrelay/status`. Version is read from `package.json` automatically.
- **Split config UI tabs** — the single "Configuration" tab is replaced by two tabs:
  - **Radio** — all existing rig connection settings, unchanged.
  - **Integrations** — non-rig plugins. Currently hosts WSJT-X Relay with enable toggle, server URL, relay key, session ID, UDP port, batch interval, and a live status indicator (polled every 5 s).
- **Integration plugin lifecycle** — `PluginRegistry` gains `connectIntegrations()` (starts all enabled integration plugins at startup) and `restartIntegration(id)` (hot-reloads a single integration when its config is saved via `POST /api/config`).
- **`wsjtxRelay` config section** — added to `DEFAULT_CONFIG`, merged in `loadConfig()`, and documented in `rig-bridge-config.example.json`.
- **`VERSION` constant** — single source of truth in `rig-bridge.js`. Replaces all previously hard-coded `'1.1.0'` strings across the startup banner, `--help` text, web UI `<title>`, page footer, and JSON health response.
- **`--version` / `-v` flag** — prints the version and exits.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [x] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist

### DIFF
--- a/rig-bridge/core/config.js
+++ b/rig-bridge/core/config.js
@@ -33,6 +33,15 @@ const DEFAULT_CONFIG = {
     flrigHost: '127.0.0.1',
     flrigPort: 12345,
   },
+  wsjtxRelay: {
+    enabled: false,
+    url: '', // OpenHamClock server URL (e.g. https://openhamclock.com)
+    key: '', // Relay authentication key
+    session: '', // Browser session ID for per-user isolation
+    udpPort: 2237, // UDP port to listen on for WSJT-X packets
+    batchInterval: 2000, // Batch send interval in ms
+    verbose: false, // Log all decoded messages
+  },
 };
 
 let config = JSON.parse(JSON.stringify(DEFAULT_CONFIG));
@@ -54,6 +63,7 @@ function loadConfig() {
         ...DEFAULT_CONFIG,
         ...raw,
         radio: { ...DEFAULT_CONFIG.radio, ...(raw.radio || {}) },
+        wsjtxRelay: { ...DEFAULT_CONFIG.wsjtxRelay, ...(raw.wsjtxRelay || {}) },
         // Coerce logging to boolean in case the stored value is a string
         logging: raw.logging !== undefined ? !!raw.logging : DEFAULT_CONFIG.logging,
       });
--- a/rig-bridge/core/plugin-registry.js
+++ b/rig-bridge/core/plugin-registry.js
@@ -27,8 +27,9 @@ class PluginRegistry {
     this._config = config;
     this._services = services; // { updateState, state, broadcast }
     this._descriptors = new Map(); // id → descriptor
-    this._instance = null; // current active plugin instance
+    this._instance = null; // current active rig plugin instance
     this._activeId = null;
+    this._integrations = new Map(); // id → running integration plugin instance
   }
 
   /**
@@ -45,13 +46,23 @@ class PluginRegistry {
       console.error(`[Registry] Failed to load USB plugins: ${e.message}`);
     }
 
-    // Single-export plugins
+    // Single-export rig plugins
     for (const file of ['rigctld', 'flrig', 'mock']) {
       try {
         const p = require(`../plugins/${file}`);
         this._descriptors.set(p.id, p);
       } catch (e) {
         console.error(`[Registry] Failed to load plugin "${file}": ${e.message}`);
+      }
+    }
+
+    // Integration plugins (non-rig, run in parallel alongside the rig plugin)
+    for (const file of ['wsjtx-relay']) {
+      try {
+        const p = require(`../plugins/${file}`);
+        this._descriptors.set(p.id, p);
+      } catch (e) {
+        console.error(`[Registry] Failed to load integration plugin "${file}": ${e.message}`);
       }
     }
   }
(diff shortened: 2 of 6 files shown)


ceotjoe:
@accius please do not remove rig-daemon and rig-listener already in next release I want to keep them as backup for a while. :-)

accius:
Please join the dev discord,
https://discord.gg/NTB6HKEkdU

Great follow-up — all four items from the v1.1.0 review are addressed and the WSJT-X integration is a clean addition. One thing needs fixing before merge.

🚫 Blocking: Yaesu MD mode parsing regression
protocol-yaesu.js — the MD case was simplified but breaks modern radios:
javascript// PR (broken)
const modeStr = data.substring(2);
const mode = MODES[modeStr] || getState('mode');
MODES keys are single characters ('1', 'C', etc.). Modern Yaesu radios (FT-991A, FT-710, FT-DX10, FT-DX101) send MD0C; — substring(2) yields "0C", MODES["0C"] is undefined, so mode silently never updates from MD responses. Since AI1 auto-info fires unsolicited MD changes frequently, this means mode only refreshes on the 30s IF keepalive — a significant responsiveness regression.
The original logic handled both formats correctly. Please restore:
javascriptcase 'MD': {
  const modeStr = data.substring(2);
  const modeDigit = modeStr.length >= 2 ? modeStr.charAt(1) : modeStr.charAt(0);
  const mode = MODES[modeDigit] || getState('mode');
  updateState('mode', mode);
  break;
}

✅ PR feedback items — all resolved

Centralized debug — config.debug set by applyCliArgs(), module-level DEBUG constants removed from both protocol-yaesu.js and usb/index.js. debug param threaded through parse() and handleData() signatures cleanly.
Robust plugin loading — every require() in registerBuiltins() wrapped in try/catch. Integration plugins get their own loop. 👍
RTS/CTS override removed — now rtscts: !!config.radio.rtscts for all radio types. Users with FT-991A just need to check the Hardware Flow box in the UI.
Test endpoint logging — DTR/RTS set callback logs warning on error instead of swallowing it.

✅ USB plugin hardening

stabilizerTimer and initialPollTimer tracked as named variables, cleared in disconnect() — no more ghost setTimeout fires after stop.
Race condition on open() callback handled: checks wasExplicitlyDisconnected and immediately closes if plugin was stopped mid-open.
removeAllListeners() before close prevents stale event handlers.
EADDRINUSE on the HTTP server exits with actionable message instead of unhandled exception.

✅ WSJT-X Relay plugin
Nice port from the standalone wsjtx-relay/relay.js into the plugin architecture. The binary protocol parser covers all 13 WSJT-X message types. Adaptive backoff on send failures (normal → 2× → 10s → 30s cap) is well thought out. _currentInstance module-level ref for route delegation is pragmatic — works since there's only ever one instance. connectIntegrations() / restartIntegration() fit naturally into the registry lifecycle. Config UI tab split (Radio / Integrations / Console Log) scales well for future plugins.
Non-blocking nits for follow-up

Version bump — VERSION is still '1.1.0'. This adds a new feature (WSJT-X relay) — should be 1.2.0.
URL validation — connect() doesn't validate the server URL before binding the UDP socket. A malformed URL would silently queue messages that never send. A quick new URL(cfg.url) check at the top of connect() would catch that early.
Queue cap — messageQueue can grow unbounded if the server is unreachable for an extended period. Consider capping at ~10k entries and dropping oldest to prevent memory creep on long-running instances.

Fix the MD regression and this is good to merge. 73!

( I am just going to merge this, and fix the yaesu protocol here so don't worry, just noting it so its fixed for future)

ceotjoe:
@accius thanks for fixing the Yaesu issue. I will take care then regarding the few things you mentioned about wsjtx plugin. 

For the wsjtx I already have something in the pipeline to configure the session ID and key automatically. You can generate a config URL within wsjtx panel and then paste this to web interface of the bridge and it will configure it automatically.

Issue #614, opened 2026-02-25
[BUG] Some WSJT-X positions are plotted incorrectly


timkirby:
This morning I was curious to see that I had heard a station in EM74 in the US on 10MHz at around 1000z. I clicked on the point and it showed the callsign J51A but in EM74. Now I had heard and worked J51A but they're not in EM74, but in West Africa. 

Looking at the decodes I heard, you'll see that I *heard* NE4S calling J51A and NE4S *IS* in EM74

<img width="1344" height="218" alt="Image" src="https://github.com/user-attachments/assets/5046584c-c9a3-4d1a-bcb1-5000d4315ef2" />

I think we've talked about some points being plotted incorrectly before and I wonder whether this sheds any light as to what's happening.

As ever - thanks for all the work - OHC is just brilliant :-)

alanhargreaves:
OK, so it sounds like the wsjt-x decoding is sometimes getting confused. Is that something you are seeing all the time or is it a one off?

alanhargreaves:
OK, I've just seen teh same thing here withe exchange:
```1281 OH3MB JI1DOM PM95```

<img width="221" height="147" alt="Image" src="https://github.com/user-attachments/assets/e0198e3b-14f0-4a7e-89dc-32ee6a87098a" />

We've got it logged against the wrong call. Now looking at the decode from the server.js I see
```
{
  "id":"WSJT-X-102415-1291-OH3MBJI1DOMPM95",
  "clientId":"WSJT-X",
  "isNew":true,
  "time":"10:24:15",
  "timeMs":37455000,
  "snr":-5,
  "dt":"-0.1",
  "freq":1291,
  "mode":"~",
  "message":"OH3MB JI1DOM PM95",
  "lowConfidence":false,
  "offAir":false,
  "dialFrequency":14074000,
  "band":"20m",
  "type":"QSO",
  "dxCall":"OH3MB",
  "deCall":"JI1DOM",
  "exchange":"PM95",
  "grid":"PM95",
  "timestamp":1772187867268,
  "lat":35.5,
  "lon":139
}
```
So, this looks OK to me. The issue must be up in the client.

I'm thinking in the WorldMap code for plotting and generating that popup.

Maybe not. Looking at the WorldMap code, it sets the caller up as spot.caller, or if that does not exist, spot.dxCall. In the case of a decoded remote QSO, this is wrong.

alanhargreaves:
[...]

PR #628, opened 2026-02-27
[BUG 614] Some WSJT-X positions are plotted incorrectly

alanhargreaves:
## What does this PR do?

Addresses the issue in https://github.com/accius/openhamclock/issues/614

Currently when generating the list of wsjx-t spots, Worldmap assumes that unless `spot.caller` is set, the station associated with the data should be `spot.dxCall`. This is _only_ true for spots that are QSOs between `caller` and a `spot.dxCall`. In the case of a decoded remote to remote, the transmitting station is `spot.deCall`.

The result of this incorrect assumption is that on the WorldMap, we get the receiving station shown in the transmitting stations location.

This fix adds a little more logic to address that.

An `npx prettier -c src/components/WorldMap.jsx` added the extra newline at the end of the file.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Look for an easily locatable station in the WSJX-T Decodes panel that is _not_ a CQ
2. Click the diamond on the map
3. Verify that it has the correct (ie second) callsign associated with it in the popup

## Checklist

### DIFF
--- a/server.js
+++ b/server.js
@@ -1630,12 +1630,7 @@ if (fs.existsSync(path.join(__dirname, '.git'))) {
     // Mark directory as safe for git — fixes "dubious ownership" errors when
     // the server runs as a different user than the repo owner (e.g. systemd
     // running as root, repo owned by 'pi')
-    execFile(
-      'git',
-      ['config', '--global', '--add', 'safe.directory', __dirname],
-      { cwd: __dirname },
-      () => {},
-    );
+    execFile('git', ['config', '--global', '--add', 'safe.directory', __dirname], { cwd: __dirname }, () => {});
   } catch {}
 }
 
@@ -11139,7 +11134,8 @@ function parseDecodeMessage(text) {
       // Cache grid — in exchange it typically belongs to the calling station (dxCall)
       const coords = gridToLatLon(result.grid);
       if (coords) {
-        wsjtxGridCache.set(result.dxCall.toUpperCase(), {
+        const call = (result.deCall == CONFIG.callsign ? result.dxCall : result.deCall).toUpperCase();
+        wsjtxGridCache.set(call, {
           grid: result.grid,
           lat: coords.latitude,
           lon: coords.longitude,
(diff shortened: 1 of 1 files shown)


alanhargreaves:
Changed the pull to draft, as I just had another incorrect one show up.

alanhargreaves:
I've not been able to Make it fail again. I think what we have is an improvement on what we had, so making it Ready for Review.

alanhargreaves:
and it failed again; I suspect the issue is actually in the server.js code where we determine lat/long for all messages regardless of whether or not they have a grid specified.

accius:
merging with notes...
PR Review — WSJT-X QSO Callsign Resolution Fix

Verdict: Approve with follow-up items

What this fixes:

In FT8 QSO exchanges (DXCALL DECALL EXCHANGE), the first call is the station being called and the second is the sender. The old code always used dxCall as the remote station for grid caching and map lookups. When someone calls you, dxCall is your callsign and deCall is theirs — so the code was caching your own grid under the remote station's key and plotting spots at your own location instead of theirs.

The fix correctly picks whichever call isn't CONFIG.callsign (server) / callsign (client), and applies this consistently across all five resolution paths: grid cache write, grid cache lookup, HamQTH cache lookup, prefix estimation, and WorldMap deduplication.

Non-blocking issues:

    Loose equality in all 5 comparisons — == should be ===. Both sides are always strings so it works, but doesn't match project conventions:
        server.js ~11143, ~11253, ~11267, ~11310
        WorldMap.jsx ~1694
    Stale comment — server.js ~11140 still says "Cache grid — in exchange it typically belongs to the calling station (dxCall)" which no longer describes what the code does.
    Parser purity — The grid cache write in parseDecodeMessage now references CONFIG.callsign, coupling a previously pure text parser to app config. The four lookups in handleWSJTXMessage already have that context. Consider moving the cache write there too in a future cleanup pass.

Issue #622, opened 2026-02-25
[FEATURE] Add the ability to select DX Target by entering a grid square directly.


W1WH:
## Use Case

How would this feature benefit amateur radio operators? Be specific about the use case:

- [x] General HF operation
- [x] VHF/UHF operation
- [ ] Satellite operation
- [x] Contesting
- [ ] POTA/SOTA
- [x] DXing
- [ ] Emergency communications
- [ ] Other: \***\*\_\_\_\*\***

## Would you be willing to help implement this?

- [ ] Yes, I can submit a PR
- [ ] Yes, I can help test
- [x] No, but I'd use it!

## Additional Context

Add any other context, screenshots, or mockups about the feature request here.

## Similar Features in Other Software

Are there similar features in other amateur radio software? If so, what do you like/dislike about their implementation?

---

73!

ceotjoe:
Taking a look into this.

ceotjoe:
@W1WH if I unterstand you correctly, you want something in DX panel, where you can enter a Maidenhead locator manually. Right?

ceotjoe:
Addressed by https://github.com/accius/openhamclock/pull/629

PR #629, opened 2026-02-27
feat: add manual Maidenhead locator input to DX panel

ceotjoe:
## Summary

- Add a manual Maidenhead locator text input to the DX target panel in all three layouts (Modern, Classic, Dockable)
- The existing DX grid square display is converted into an editable field — no extra space is consumed
- Typing a valid 4- or 6-character locator (e.g. `JN58` or `JN58sm`) and pressing **Enter** (or leaving the field) moves the DX map marker and updates all derived data: bearing, distance, and sunrise/sunset times
- The field is **read-only** when the DX lock 🔒 is active, and **reverts** on Escape or invalid input
- Map clicks and spot clicks continue to work as before and sync back into the field
- Addresses https://github.com/accius/openhamclock/issues/622 

## Changes

| File | What changed |
|------|-------------|
| `src/utils/geo.js` | Export `parseGridSquare` (grid → lat/lon); was previously a private duplicate inside SettingsPanel |
| `src/components/SettingsPanel.jsx` | Import shared `parseGridSquare` from geo.js; remove local duplicate |
| `src/components/DXGridInput.jsx` | **New** reusable editable locator component used by all three layouts |
| `src/layouts/ModernLayout.jsx` | Replace static `{dxGrid}` span with `<DXGridInput>` |
| `src/layouts/ClassicLayout.jsx` | Replace static `{dxGrid}` span with `<DXGridInput>` |
| `src/DockableApp.jsx` | Replace static `{dxGrid}` div with `<DXGridInput>` |
| `src/lang/*.json` | Add `app.dxLocation.gridInputTitle` / `gridInputTitleLocked` — translated for all 14 supported languages |

## Test plan

- [ ] Typing `JN58sm` moves the DX marker to the Munich area
- [ ] Typing a 4-char grid `JN58` also works (centers on the grid square)
- [ ] Bearing, distance and sun times update after input
- [ ] Invalid input (e.g. `XY`) reverts to the current grid on blur
- [ ] Escape reverts without updating
- [ ] Map click still updates the input field
- [ ] DX cluster spot click still updates the input field
- [ ] Lock active → input is read-only (cursor: not-allowed)
- [ ] Verified in all three layouts: Modern, Classic, Dockable
- [ ] No regressions in DE location display or SettingsPanel grid input

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -35,6 +35,7 @@ import { loadLayout, saveLayout } from './store/layoutStore.js';
 import { DockableLayoutProvider } from './contexts';
 import { useRig } from './contexts/RigContext.jsx';
 import { calculateBearing, calculateDistance, formatDistance } from './utils/geo.js';
+import { DXGridInput } from './components/DXGridInput.jsx';
 import './styles/flexlayout-openhamclock.css';
 import useMapLayers from './hooks/app/useMapLayers';
 import useRotator from './hooks/useRotator';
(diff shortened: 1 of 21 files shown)


accius:
Looks good — merging this. Nice cleanup extracting parseGridSquare into the shared utility, and the component handles all the edge cases well (escape revert, lock state, external sync).

A few things to tighten up in a follow-up: parseGridSquare doesn't validate that the field characters are in the Maidenhead range (A–R for fields, A–X for subsquares), so something like ZZ99 will produce coordinates outside ±90/±180 instead of returning null. Related, a 5-character input like JN58s slips through — it passes the length < 4 check but doesn't hit the length >= 6 subsquare branch, so the 5th character is silently ignored. Should probably require exactly 4 or 6.

On the cosmetic side, DXGridInput.jsx has an explicit import React that the rest of the codebase doesn't need (automatic JSX transform), and the outer <div> in DockableApp still has the amber/22px/700 styles that are now redundant since the same values get passed through the style prop to the input.

PR #646, opened 2026-03-01
feat: Keyboard shortcuts to toggle map overlays

### LINKED ISSUE #635: [FEATURE] Add keyboard shortcuts to toggle map layers on/off

It's annoying to have to open the settings menu to toggle "busy" layers on and off.

Create keybindings dynamically from all possible layers, and provide a floating and dockable panel to show current keybindings.

## Use Case

- [x] General HF operation
- [ ] VHF/UHF operation
- [x] Satellite operation
- [x] Contesting
- [x] POTA/SOTA
- [x] DXing

- [x] Yes, I can submit a PR
- [x] Yes, I can help test
- [ ] No, but I'd use it!

---

73!

KidVizious:
## What does this PR do?

Adds key bindings to toggle map layers on and off. It also adds both a floating panel and dockable panel with the current list of dynamically created keybinds.

Resolves #635 
## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Press "?" key, panel should appear with a list of keybindings
2. Switch to dockable layout and add a panel, "Keyboard Shortcuts"
3. Panels should respect user's theme choice
4. Press each key listed to toggle the respective layer on/off

## Checklist

Some sample screenshots are below. I didn't post every combo for brevity sake.

Screenshot 2026-02-28 200313
Screenshot 2026-02-28 200333
Screenshot 2026-02-28 200216

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -415,6 +415,7 @@
   "app.mapControls.calls.hide": "Hide Calls",
   "app.mapControls.calls.show": "Show Calls",
   "app.legend.moon": "Moon",
+  "propagation.heatmap.tooltip.stoplight"
   "keybindings.panel.title": "KEYBOARD SHORTCUTS",
   "keybindings.panel.description": "Press the following keys to toggle map layers:",
   "keybindings.panel.toggle": "Toggle this help panel",


accius:
Hey, nice work on this — keyboard shortcuts for layer toggling is a great addition and the dual-mode panel (modal + dockable) is well thought out. A few things I'd like addressed before merging:

**Bug — Indentation**
The closing `/>` on `<KeybindingsPanel>` in App.jsx (around line 531) uses a tab instead of spaces. Please fix to match the rest of the codebase.

**Bug — Removed translation key**
The `propagation.heatmap.tooltip.stoplight` key was dropped from `en.json`. Is that intentional? If it's still referenced anywhere that'll break. Can you grep for it and confirm?

**Concern — Shortcut stability**
The auto-assignment algorithm (first unique letter per layer name) is clever, but it means shortcuts silently reshuffle whenever a layer is added or renamed. A user who learns "G = Greyline" could find G suddenly toggles something else after an update. Would you be open to adding an optional `shortcut` property to the layer registry so we can pin important ones? The auto-assign can still be the fallback for layers that don't specify one.

**Minor — Global coupling**
The `window.hamclockLayerControls` usage works but is a bit fragile. Not a blocker for this PR, but something we should think about cleaning up down the road.

**Minor — Static deps on useMemo**
`layerShortcuts` has `[]` dependencies so it only computes on mount. That's fine if all layers are registered at startup, but worth noting — if we ever load layers dynamically this will need to update.

TL;DR: Fix the indentation and confirm the removed translation key, and this is good to go. The shortcut stability thing can be a follow-up issue. Thanks for putting this together!

accius:
I am closing this and adding in the fixes as I wouldn't mind having this in tomorrows prod release.

KidVizious:
@accius I implemented your fixes as well as some other improvements if you want to take another look.
- Layer names now sorted before assigning keys
- Only use layer.id instead of name. Seems like these are used sort of interchangeably between layers but id is usually more sane for keybindings
- Stoplight was in en.json twice
- Fixed tab indentation

Issue #601, opened 2026-02-24
[FEATURE] Replace Ctrl + Drag for widget repositioning with Shift + Drag


echo-gravitas:
## Is your feature request related to a problem?

On macOS, `Ctrl + Click` triggers a context menu. The application currently requires holding `Ctrl` to drag widgets. As a result, macOS users cannot drag widgets without triggering the context menu, effectively breaking the intended interaction.

## Describe the Solution You'd Like

Replace `Ctrl + Drag` for widget repositioning with `Shift + Drag`.

On macOS, `Ctrl + Click` is reserved for opening the context menu.
Using `Shift` avoids this platform conflict while preserving a modifier-based drag interaction.

`Shift` is currently not assigned to widget movement and does not trigger native OS behavior that interferes with dragging.

## Describe Alternatives You've Considered

none

## Use Case

How would this feature benefit amateur radio operators? Be specific about the use case:

- [x] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [ ] Contesting
- [ ] POTA/SOTA
- [ ] DXing
- [ ] Emergency communications
- [ ] Other: \***\*\_\_\_\*\***

## Would you be willing to help implement this?

- [x] Yes, I can submit a PR
- [x] Yes, I can help test
- [ ] No, but I'd use it!

## Additional Context

none

## Similar Features in Other Software

none

---

73!

Delerius:
You’re absolutely right that Ctrl + Click maps to the native context menu on macOS, which creates a platform conflict with our current Ctrl + Drag widget repositioning behavior. That’s definitely not ideal.

We do need to be careful with Shift, though. In recent work on the Rotator panel, Shift + Click on the map is used to send the rotator to a clicked bearing. Reassigning widget drag to Shift + Drag would create a new conflict there.

A couple possible approaches we could consider:

Platform-specific modifier

Keep Ctrl + Drag on Windows/Linux

Use Option (Alt) + Drag on macOS
This avoids native OS conflicts and keeps Shift free.

Allow either Ctrl or Alt

Accept multiple modifiers for drag

Avoid Shift to preserve rotator behavior

Make the modifier configurable in Settings

Advanced option, but future-proof

I agree we should avoid fighting native macOS behaviors. The cleanest solution may be using Option + Drag on macOS while leaving existing behavior unchanged elsewhere.

never-rain:
[...]

PR #648, opened 2026-03-01
Unify map overlay widget drag-and-toggle behavior

echo-gravitas:
**PR Summary**

This PR removes the old `Ctrl+drag` interaction for map overlay widgets and replaces it with direct drag-and-drop from the widget title bar, fixing the macOS `Ctrl+click` right-click conflict. Feature request #601.

It also standardizes the overlay widget chrome across affected panels:
- drag now happens from the title only
- minimize/maximize is handled by a dedicated toggle button
- widget titles now share the same font, size, and color
- toggle buttons now use a consistent MUF Map style
- drag behavior no longer lags on WSPR panels
- minimize actions no longer cause panels to drift or shift position
- VOACAP, MUF Map, and Lightning panel edge cases around drag/toggle interaction were fixed
- unused refs, constants, functions, and states were removed as part of the cleanup

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -249,3 +249,10 @@ VITE_PSTROTATOR_TARGET=http://192.168.1.43:50004
 # How long the server keeps logged QSOs in memory (minutes)
 # 1440 = 24 hours
 N3FJP_QSO_RETENTION_MINUTES=1440
+
+# ===========================================
+# DOCKER / HEALTH CHECK
+# ===========================================
+
+# Override the health check endpoint (useful behind a reverse proxy)
+# HEALTH_ENDPOINT=http://localhost:3000/api/health
--- a/.github/workflows/auto-in-progress.yml
+++ b/.github/workflows/auto-in-progress.yml
@@ -1,0 +1,65 @@
+name: Auto move to In Progress on assignment
+
+on:
+  issues:
+    types: [assigned]
+  pull_request:
+    types: [assigned]
+
+jobs:
+  move-to-in-progress:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.PROJECT_TOKEN }}
+          script: |
+            const projectNumber = 1;
+            const orgOrUser = 'accius';
+            const isOrg = false;
+
+            // Get project ID
+            const query = isOrg
+              ? `query { organization(login: "${orgOrUser}") { projectV2(number: ${projectNumber}) { id, field(name: "Status") { ... on ProjectV2SingleSelectField { id, options { id, name } } } } } }`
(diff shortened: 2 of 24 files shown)


accius:
Great work on this — the move from Ctrl+drag to direct title dragging is a much better UX. The dead zone, click suppression, and transition disabling during drag are nice touches. Panel styling standardization across all the layers looks clean. A few things to address before I can merge:

**Blocker — N3FJP panel lost minimize and drag**
The entire `addMinimizeToggle` function was deleted from `useN3FJPLoggedQSOs.js` and the `makeDraggable` import was removed, but I don't see a replacement. Was this intentional? If so, what's the plan for that panel? If not, it needs the same title-handle treatment as the other layers before this can go in.

**Suggestion — Shared `addMinimizeToggle` utility**
The same minimize toggle pattern (create button, wrap content, build title with drag handle, bind toggle) is now copy-pasted across Gray Line, Lightning, RBN, and WSPR with nearly identical code. This was a good opportunity to extract a shared utility alongside `makeDraggable.js`. Not blocking merge on this, but I'd love to see it as a follow-up.

**Minor — Satellite panel inconsistency**
The satellite data window uses its own drag system checking for `.sat-data-window-title` class rather than the `data-drag-handle` attribute that `makeDraggable.js` looks for. Makes sense since it has its own drag implementation, but worth a comment in the code noting why it diverges from the shared pattern.

**Minor — MUF/VOACAP `toggleBound` guard**
Using `dataset.toggleBound` to prevent duplicate event listeners works, but it's a bit fragile. An AbortController pattern (like `makeDraggable` uses) would be more robust. Not a blocker.

Please fix the N3FJP issue and I'm happy to merge. Everything else can be follow-up issues. Nice improvement overall — the Ctrl+drag was not discoverable at all, this is way better.

never-rain:
**N3FJP** - Good catch. That was not intentional. The N3FJP overlay panel had indeed lost both the draggable control and minimize toggle during cleanup. I restored the panel, re-added `makeDraggable`, and brought it in line with the other layer widgets using the same title-handle drag and dedicated minimize button pattern.

**Shared addMinimizeToggle utility** - I took that suggestion and extracted the duplicated minimize/title-handle logic into a shared utility alongside `makeDraggable.js`. All may layer widgets now use the same helper, so the panel behavior is centralized instead of copy-pasted.

**Satellite panel inconsistency** - For the satellite panel, I added an inline code comment explaining why it intentionally diverges from the shared `makeDraggable()` pattern: it uses its own top/right-based drag model relative to the map container.

**MUF/VOACAP `toggleBound` guard**  - For the MUF Map / VOACAP toggle binding, I replaced the `dataset.toggleBound` guard in the shared minimize helper with an AbortController-based listener lifecycle, which is more robust when panels are re-initialized.

accius:
nice work, all blockers resolved.. merging

Issue #553, opened 2026-02-21
[FEATURE] Suppress DX Icon


jlp78:
## Is your feature request related to a problem?

I have no interest in DX information at all, but if I click somewhere on the map, a DX icon shows up and there is no way to remove it.  Currently, clicking on the DX icon pops up a small info panel that includes an "x" dismiss button, but clicking that x only removes the popup, not the DX icon itself.

## Describe the Solution You'd Like

I'd like two things:  first, a way to remove the DX icon if one is being displayed.  Second, a way to prevent clicking from placing a DX icon at all.

## Describe Alternatives You've Considered

I have not found a solution or work around for this.

## Use Case

How would this feature benefit amateur radio operators? Be specific about the use case:

- [ ] General HF operation
- [X] VHF/UHF operation
- [ ] Satellite operation
- [ ] Contesting
- [ ] POTA/SOTA
- [ ] DXing
- [ ] Emergency communications
- [ ] Other: \***\*\_\_\_\*\***

## Would you be willing to help implement this?

- [ ] Yes, I can submit a PR
- [X] Yes, I can help test
- [ ] No, but I'd use it!

---

73!

accius:
I am not sure why you wouldn't want to see propagation between home and a specific point, it's sort of the main function of the dashboard.  whats stopping you from clicking really close to DE and locking it there and turning off every panel that has to do with propagation in the dashboard?  not being rude, I just don't see the point of taking away half or more of the features of the dashboard

jlp78:
Given that you have no idea why someone would not be interested in DX information at all, I'm not surprised that you don't see any utility to this feature request.  Note that I'm not asking for the DX functionality to be removed for everyone, just to have a knob to turn it off.

I trust you don't have a problem if I fork your project and work on adding the feature to my own version?  Would you entertain a pull request with this functionality?

never-rain:
Let’s try to keep this friendly. OpenHamClock is a community project, and that’s exactly what makes it stronger over time. Different use cases don’t weaken a product, they often reveal where it can become more flexible.

As a UX designer by profession, I’m always interested in understanding the underlying need behind a request. Not necessarily to agree with it immediately, but to understand the context in which the product is being used.

The original **HamClock** was primarily focused on visualizing propagation and helping operators assess DX conditions. That was its core identity.
OpenHamClock, however, has evolved significantly and continues to grow in scope. With that growth comes a broader audience, and inevitably, a wider range of expectations and usage patterns.

[...]

PR #653, opened 2026-03-02
Feature Request #553: add separate toggle for DE/DX markers in Map Layers

echo-gravitas:
This PR implements Feature Request #553.

Previously, the `DE` and `DX` markers on the map could not be hidden independently. This change adds a dedicated `DE/DX Markers` option in `Settings -> Map Layers`.

**What changed**

- Added a new map-layer toggle: `DE/DX Markers`.
- The `DE` and `DX` markers in `WorldMap` are now rendered only when that option is enabled.
- The setting is persisted alongside the other map-layer preferences in `openhamclock_mapLayers`.
- The new option was integrated into the existing `Map Overlays` section in the Settings panel.
- The `Map Overlays` section was moved to the top of the layer list.
- Toggling now takes effect immediately without requiring a manual page reload.
- Also fixed an existing `DX News Ticker` issue:
  - disabling it hid it correctly,
  - but re-enabling it did not always show it again immediately.
  - Root cause was missing synchronization between the ticker's internal visibility state and updated `mapLayers`.
  - `useMapLayers` now dispatches a `mapLayersChanged` event so the ticker updates immediately in the same tab.

**Note**

I had to run `npm i` because `package.json` and `package-lock.json` were out of sync. The lockfile was updated accordingly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -111,7 +111,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -467,7 +466,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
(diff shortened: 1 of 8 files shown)


accius:
Nice PR — clean implementation, merging now. One fix I'm applying post-merge:

In `src/components/SettingsPanel.jsx`, the category rendering loop allows the `overlay` category to render when `grouped['overlay']` is undefined (because `hasBuiltInOverlays` is true). The built-in overlay cards render fine, but then it hits:

```js
grouped[key].forEach((layer) => {
```

This throws `TypeError: Cannot read properties of undefined` if no plugin layers have `category: 'overlay'` loaded (fresh install, plugin load failure, etc.). Fixed with:

```js
(grouped[key] || []).forEach((layer) => {
```

Thanks for the contribution! 73

Issue #206, opened 2026-02-07
[FEATURE] Adding DXCC Prefix list for selecting DX on Map


Chuck-K0XM:
## Is your feature request related to a problem?
No

## Describe the Solution You'd Like
When click on the 'DX' portion of the map to set the DX desired, pop up a list of the DXCC countries to select from

## Use Case
How would this feature benefit amateur radio operators? Be specific about the use case:
- [ X] General HF operation
- [ X] VHF/UHF operation
- [X ] Contesting
- [ ] POTA/SOTA
- [X ] DXing

## Would you be willing to help implement this?
- [ ] Yes, I can submit a PR
- [X ] Yes, I can help test
- [ ] No, but I'd use it!

## Additional Context
Maybe right click can allow you to select actual Lat/Long, OR have a button to allow a DXCC list to choose from, or Prefix entry

---
73!

never-rain:
Hi @Chuck-K0XM, thanks for the suggestion. I’m trying to understand the exact use case before I implement anything.

When you say _click on the DX portion of the map to set the DX desired_, what do you use that _DX desired_ for today, and what should change after selecting a DXCC entity?

A few interpretations I can think of, can you confirm which one matches your workflow?

1. **DXCC hunting shortcut**
 You want to pick a DXCC country/entity (e.g., “Japan”) and the app automatically sets the DX target to that entity (e.g., a representative lat/long like the centroid or capital), so bearings / map focus / overlays use that as the DX point.

2. **Precision vs convenience**
 The map-click method is too imprecise, so you want a DXCC list or prefix entry as a deterministic way to set a target location.

Also, for the _right-click to select actual Lat/Long_:

- Do you mean you want to enter coordinates manually (numeric input), or pick an exact point on the map without snapping to anything?

If you can share one concrete example (_I’m chasing DXCC X, I want to do Y, and I expect the UI to show Z_), I can design the interaction so it’s actually useful and not just a random list pop-up.

73

Chuck-K0XM:
Option 1 would be perfect. You can choose a DXCC Country Like Bouvet which is coming up real soon, and then the DX location would be shown on the map along with VOACAP between you and the dx, etc...all the needed info.

never-rain:
Hey @Chuck-K0XM — this is my current implementation in the dev environment. Could you take a look?

![Image](https://github.com/user-attachments/assets/c8e5504a-f718-4b47-8db5-3cbaa249f066)

Does this meet your expectations?

[...]

PR #667, opened 2026-03-03
Add DXCC target selector for DX Target panel

echo-gravitas:
## Summary

This PR adds a DXCC entity selector to the DX Target panel, requested in feature request #206.

Users can now open a DXCC input next to the Maidenhead locator, search/select a DXCC entity, and move the DX target location to the corresponding entity coordinates. The selector also includes a clear button for the current input.

## What changed

- added a DXCC selector for the DX Target panel
- placed the DXCC toggle button directly next to the DX Maidenhead locator
- used existing `cty.dat` entity data already loaded by the app
- wired selection into the existing DX target location flow, so map/bearing/distance/propagation update automatically
- added a clear (`x`) button to reset the DXCC input field
- added the new i18n keys to all language files, with English defaults/fallback text

select-dxcc

## Notes

This implementation intentionally reuses the current DX target update mechanism instead of introducing a separate location path.

One technical note: `DE Location` and `DX Target` are currently implemented inline in multiple layouts rather than as shared components. These panels are starting to become heavier, and the dockable layout already shows some additional render overhead. It may make sense in a follow-up to extract them into dedicated components for better reuse, maintainability, and render isolation.


accius:
Nice feature — the component is well structured and the CTY data wiring is clean. A few things to address before merge:

**Unused translation keys** — `dxccShowButton` and `dxccHideButton` are defined in all 15 language files but never referenced anywhere in the code. The toggle button just renders the hardcoded string `DXCC`. Either wire up the translation keys or remove them.

**Untranslated strings in non-English files** — Only `de.json` has actual German translations for the new keys. The other 13 languages (es, fr, it, ja, ko, ms, nl, pt, sl, zh, ca, ka, ru) all have English text copy-pasted in. These should either be properly translated or removed from those files so i18next falls back to `en.json` naturally. Having English in a non-English file makes it look translated when it isn't, and makes missing translations harder to find later.

**`textAlign: 'right'` removed from DXGridInput** — This changes the grid square alignment for all users regardless of whether the DXCC selector is visible. Was this intentional? It's a separate visual change from the DXCC feature and might surprise people.

Non-blocking nits:
- Selecting from the datalist dropdown sets the input value but doesn't commit until blur or Enter. User picks "Japan (JA)" from the list and nothing visibly happens until they click away. Could auto-commit on exact match in onChange as a polish pass, but not a blocker.

Everything else looks good — `useId()` for the datalist, the CTY event listener with cleanup, optionMap memoization, and the locked/disabled states are all solid.

never-rain:
@accius - thank you for the feedback.

Since I’m not able to provide proper translations for all languages myself, I removed the untranslated translation keys from the non-English files. We could potentially use AI here, but I’d rather fall back to English than ship incorrect AI-generated translations.

I removed `textAlign: 'right'` in `DXGridInput` intentionally, because it created an awkward empty gap in the panel. With everything left-aligned, the layout looks cleaner and more consistent. See the screenshot from the current version on openhamclock.com:

Bildschirmfoto 2026-03-04 um 18 59 46

Lastly, I really liked your suggestion to update the DX Target immediately on `onChange` when selecting from the input. I’ve implemented that as well. It makes the feature feel complete and improves the UX nicely.

PR #669, opened 2026-03-03
feat: add inline text size controls to DX News ticker

ceotjoe:
### Summary

Adds **+** and **−** buttons to the right side of the DX News scrolling ticker banner,
allowing users to resize the text directly on the map overlay without going into settings.
Useful for smaller screens where the default 11 px ticker text is hard to read.

### Changes

**`src/components/DXNewsTicker.jsx`**
- Added `textScale` state (range `0.7×` – `2.0×`, step `0.1`, default `1.0`) persisted
  to `localStorage` (`openhamclock_dxNewsTextScale`)
- Replaced all hardcoded `fontSize` values with `BASE_LABEL_SIZE * textScale` /
  `BASE_TEXT_SIZE * textScale`
- Bar height in map overlay mode scales proportionally: `28 * textScale` px
- Animation duration recalculates when `textScale` changes so scroll speed stays
  consistent with wider/narrower text
- +/− buttons sit outside the scrolling viewport on the right, separated by a border —
  amber when active, grey/disabled at the limits

**`src/lang/*.json`** (all 15 locales)
- Added `app.dxNews.increaseTextSize` and `app.dxNews.decreaseTextSize` aria-label keys
- English values in `en.json`; English fallback in all other locales

### Test plan

- [ ] Open map view → DX News ticker visible at bottom of map
- [ ] Click **+** several times → text grows, bar height grows, buttons scale proportionally
- [ ] Click **−** → text shrinks; **−** greys out and disables at `0.7×`
- [ ] **+** greys out and disables at `2.0×`
- [ ] Refresh page → scale setting is preserved
- [ ] Switch to Classic layout → buttons appear and work in sidebar ticker too
- [ ] Click ticker text to pause/resume → still works after resizing
- [ ] Scroll speed adjusts correctly to the new content width after resize

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

Bildschirmfoto 2026-03-03 um 22 17 02
Bildschirmfoto 2026-03-03 um 22 16 45

### DIFF
--- a/src/components/DXNewsTicker.jsx
+++ b/src/components/DXNewsTicker.jsx
@@ -6,6 +6,11 @@
 import React, { useState, useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 
+// Base font sizes (px) — all sizes are derived by multiplying with textScale
+const BASE_LABEL_SIZE = 10; // "📰 DX NEWS" label, separator ◆
+const BASE_TEXT_SIZE = 11; // news titles and descriptions
+const BASE_HEIGHT = 28; // container height in map overlay mode (px)
+
 // Check if DX News is enabled (reads directly from localStorage as belt-and-suspenders)
 function isDXNewsEnabled() {
   try {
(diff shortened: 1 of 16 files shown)


accius:
Clean implementation — the base constants with a multiplier approach is the right way to do this, and the animation duration recalc on textScale change is a nice touch.

Two things:

**Untranslated strings in non-English files** — same issue flagged on the DXCC PR. All non-English lang files have English text for `increaseTextSize` / `decreaseTextSize`. These are aria-labels so sighted users won't notice, but screen reader users in other languages will hear English. Either translate them or remove from non-English files so i18next falls back to `en.json` naturally — having English in a non-English file masks missing translations.

**Container height only scales in overlay mode** — `BASE_HEIGHT * textScale` is inside the `!sidebar` block. When `sidebar = true` the height is unset and relies on flex, so at 2.0x scale the text could clip depending on sidebar height constraints. Probably fine since the sidebar is more flexible, but worth a quick visual check at max scale to confirm.

Everything else looks good — `toFixed(1)` avoids float drift, min/max clamping is solid, and the buttons are unobtrusive. LGTM pending the translation cleanup.

accius:
(i Merged it anyway)

PR #670, opened 2026-03-03
Update to reduce SOTA API load

vk3arr:
## What does this PR do?

The current implementation of SOTA Spots uses deprecated SOTA APIs and violates terms of service as the SOTA MT have not been contacted to sign-off on API usage.  This has resulted in substantial load increase on the SOTA servers.  This PR implements the correct usage of the SOTA API.  

It does *not* alter the code under /hooks/useSOTASpots.js, which I'll leave as an exercise for someone else to tackle as I don't know your internal spots format and a lot of what is going on there seems more OHC internal.  I will note that:
 - the ISO8601 hacks are not needed if using the modern Spot format, as this always returns a timestamp that is ISO8601 compliant - indeed that is why the old endpoints are deprecated and have been for years.
 - QRT spots are a specific type of spot and handled via a SpotType field, so you can usually find this without having to scan comments fields (which is unreliable)
 - Frequency is consistently a float/number value, so I imagine line 73 is ok, but the whole line appears to be a no-op to me at present based on the comments.

## Type of change

- [ X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Run the server, fetch the spots
2. Contact the SOTA MT via https://www.sota.org.uk/Contact to agree to terms of service and verify load is dropping.
3. ???
4. Profit.

## Checklist

- [ N/A] App loads without console errors
- [ N/A ] Tested in **Dark**, **Light**, and **Retro** themes
- [ N/A  ] Responsive at different screen sizes (desktop + mobile)
- [ X ] If touching `server.js`: caches have TTLs and size caps (we serve 2,000+ concurrent users)
- [ N/A ] If adding an API route: includes caching and error handling
- [ N/A ] If adding a panel: wired into Modern, Classic, and Dockable layouts
- [ N/A ] No hardcoded colors — uses CSS variables (`var(--accent-cyan)`, etc.)
- [ N/A ] No `.bak`, `.old`, `console.log` debug lines, or test scripts included

### DIFF
--- a/server.js
+++ b/server.js
@@ -2720,13 +2720,18 @@ app.get('/api/sota/spots', async (req, res) => {
       return res.json(sotaCache.data);
     }
 
-    const ep = await fetch('https://api-db2.sota.org.uk/api/spots/epoch');
-    const epoch = await ep.text();
-
-    // epoch is consistent with previous fetched spots, do not refetch
-    if (epoch == sotaEpoch) {
-      res.set('Cache-Control', 'no-store');
-      return res.json(sotaCache.data);
+    // Check epoch to avoid unnecessary refetch (wrapped in try/catch so
+    // a failing epoch endpoint doesn't 500 the whole spots route)
+    let epoch = '';
+    try {
+      const ep = await fetch('https://api-db2.sota.org.uk/api/spots/epoch');
+      epoch = await ep.text();
+      if (epoch === sotaEpoch && sotaCache.data) {
+        res.set('Cache-Control', 'no-store');
+        return res.json(sotaCache.data);
+      }
+    } catch (e) {
+      // Epoch check failed — fall through to normal spots fetch
     }
 
     checkSummitCache(); // Updates sotaSummits if required


accius:
Typically I would request that you point this to the staging branch, but because of what it solves I will ignore that and check the diff and push to main.

accius:
Nice optimization — the two-tier cache (TTL first, then epoch check) avoids unnecessary refetches. One thing to fix before merge:

**No error handling on the epoch fetch** — if the `/api/spots/epoch` endpoint is down or slow, `await ep.text()` throws and the entire `/api/sota/spots` route returns a 500, even when the existing cache is still perfectly usable. Wrap it in a try/catch so it falls through to the normal spots fetch on failure:

```js
let epoch = '';
try {
  const ep = await fetch('https://api-db2.sota.org.uk/api/spots/epoch');
  epoch = await ep.text();
  if (epoch === sotaEpoch && sotaCache.data) {
    res.set('Cache-Control', 'no-store');
    return res.json(sotaCache.data);
  }
} catch (e) {
  // Epoch check failed — fall through to normal fetch
}
```

Also minor: `epoch == sotaEpoch` should be `===` — both are strings so it works either way, but strict equality is consistent with the rest of the codebase.

alanhargreaves:
I feel a lot more comfortable with this.

vk3arr:
I can confirm we've seen ~40% drop in load over the weekend from the openhamclock requests.   Thanks for pushing quickly.

vk3arr:
Just a reminder for someone to please reach out to the SOTA MT at https://www.sota.org.uk/Contact as per our terms of use so we have a point of contact regarding OHC usage of the API.

accius:
@vk3arr thank you for the report. I just filled out the form.

Issue #660, opened 2026-03-03
[BUG] POTA & SOTA panels truncate information


trevlarson-ORE:
## Describe the Bug

In the Dockable layout, the POTA & SOTA panels appear to truncate information regardless of the width of the panel.

<img width="471" height="321" alt="Image" src="https://github.com/user-attachments/assets/9267c06a-5f28-4007-963c-386c87535479" />

<img width="469" height="320" alt="Image" src="https://github.com/user-attachments/assets/77f36cd1-3ed3-40d6-8cc1-db618e855430" />

## Environment

- **Browser**: Chromium 144.0.7559.10
- **OpenHamClock Version**: 15.6.4
- **Running As**: Web browser locally install on RPi.

---

73!
Trevor, N6TJL

alanhargreaves:
This was a deliberate choice. I wanted to ensure that the target (park, summit, bunker) was shown, but mousing over would give a popup with the target and its name.

alanhargreaves:
I'll investigate changing it back to just the target and having a specific mouse-over action to show the extra data.

trevlarson-ORE:
Ok, I don't get any pop-up or additional info when I mouse over.  And clicking on an entry moves the DX target, but there's still no additional info.  I'm actually not even sure what is truncated...  I see the park reference is there - I assume the truncated text is the full park name?

> This was a deliberate choice. I wanted to ensure that the target (park, summit, bunker) was shown, but mousing over would give a popup with the target and its name.

alanhargreaves:
I've tidied up the code so that the column only provides the reference and hovering the mouse over the reference gives the reference and name. See PR https://github.com/accius/openhamclock/pull/672

There are before and after screenshots in the PR.

alanhargreaves:
FYI for chromium, this is what google is telling me, ...

Check the "Show tab preview on hover" setting: In some Chromium versions, there is a setting related to tab hover behavior.
Navigate to chrome://settings/appearance in your address bar.
Look for an option like "Show tab preview on hover" and ensure it is enabled.

trevlarson-ORE:
Here are the two related options I see in the Appearance settings.  They are already turned on:

[Chromium 144.0.7559.10]
<img width="689" height="160" alt="Image" src="https://github.com/user-attachments/assets/fe3e9377-baa9-43b0-814a-67c89e2969a1" />

No pop-up on hover:
[...]

PR #672, opened 2026-03-04
[BUG 660] POTA & SOTA panels truncate information

alanhargreaves:
## What does this PR do?

Updates ActivatePanel to only show the spot.ref in the reference column, but populates the title with "spot.ref - spot.name".

There has been some confusion about how I previously implemented this by putting the reference and name in the displayed column, and letting it display an ellipsis so that mousing over would display the reference and name.

This is a cleaner way of doing it.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. View an ActivatePanel and verify that we only see the reference and no ellipsis in the reference column
2. Mouse over the reference
3. Verify that the popup displays the reference and name

## Checklist

Before:
image

After:
image

### DIFF
--- a/src/components/ActivatePanel.jsx
+++ b/src/components/ActivatePanel.jsx
@@ -175,8 +175,9 @@ export const ActivatePanel = ({
                     textOverflow: 'ellipsis',
                     whiteSpace: 'nowrap',
                   }}
+                  title={`${spot.ref} - ${spot.name}`}
                 >
-                  {`${spot.ref} - ${spot.name}`}
+                  {spot.ref}
                 </span>
                 <span style={{ color: 'var(--accent-cyan)', textAlign: 'right' }}>
                   {(() => {


trevlarson-ORE:
@alanhargreaves - Thanks for the additional work on this.  In Chromium (ver. 144.0.7559.109), mouse over isn't producing a pop-up with any information.  I guess I'll wait for this PR to get merged into main and test again.  Thanks!

alanhargreaves:
That is peculiar, I wonder if it is a browser option to enable the title field to pop up. I was considering modifying the Weather popup to contain this information, but the weather popup only works for local installs.

alanhargreaves:
Fixes: #660

PR #680, opened 2026-03-05
Feature/satellite popup enhancements

creinemann:
## What does this PR do?

Satellite popup now displays downlink, uplink, tone, arm tone, HRPT, GRB, and SD, etc, frequencies pulled from satconfig.json
Fixed stale setSatellites closure by moving state internal to the hook
Fixed tle1/tle2 property names (were incorrectly line1/line2)
Added lead and tail track to satellite orbit
Added a feature for the "blinking" green VISIBLE indicator
Naarrowed popup to take up less "real estate on screen, and moved "clear all" to top of window.
Added satconfig.json to src/satellites/
Only shows rows for fields that have data (no empty rows)
Added Lat Lon and Speed to sat data
Color coded Sats in JSON for later use
Left TLE handling unchanged in server.js, other than pulling satellite list from the new config JSON
Refactored the entire satellite list to be compatible for dispaying frequencies, tone, etc.
Only left console items that would show a satellite load failure, or a json error.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Load a satellite from map layer, click on satellite on map, pop-up will dispay information

## Checklist

- [V] Tested in **Dark**, **Light**, and **Retro** themes

Before:
Screenshot 2026-02-26 083105

After:
Screenshot 2026-03-05 164125
New sat trails:
Screenshot 2026-03-05 164252

### DIFF
--- a/server.js
+++ b/server.js
@@ -7492,32 +7492,9 @@ const TLE_CACHE_DURATION = 12 * 60 * 60 * 1000; // 12 hours — TLEs don't chang
 const TLE_STALE_SERVE_LIMIT = 48 * 60 * 60 * 1000; // Serve stale cache up to 48h while retrying
 let tleNegativeCache = 0; // Timestamp of last total failure
 const TLE_NEGATIVE_TTL = 30 * 60 * 1000; // 30 min backoff after all sources fail
-// OFFLINE BLOCK
-// Load local TLE fallback file on startup (used when all upstream sources fail)
-// Set TLE_LOCAL_FALLBACK=src/satellites/weather.tle in .env to enable
-(() => {
-  const fallbackPath = process.env.TLE_LOCAL_FALLBACK;
-  if (!fallbackPath) return;
-  try {
-    const fullPath = path.join(__dirname, fallbackPath);
-    if (!fs.existsSync(fullPath)) {
-      console.warn(`[Satellites] Local fallback file not found: ${fullPath}`);
-      return;
-    }
-    const raw = fs.readFileSync(fullPath, 'utf8');
-    const data = JSON.parse(raw);
-    if (!tleCache.data) {
-      tleCache = { data, timestamp: Date.now() };
-      console.log(
-        `[Satellites] Seeded TLE cache from local fallback: ${fallbackPath} (${Object.keys(data).length} satellites)`,
-      );
-    }
-  } catch (e) {
-    console.warn(`[Satellites] Failed to load local fallback TLE: ${e.message}`);
-  }
-})();
-// OFFLINE BLOCK
+
 // TLE data sources in priority order — automatic failover
+
 const TLE_SOURCES = {
   celestrak: {
     name: 'CelesTrak',
@@ -7587,7 +7564,11 @@ const TLE_SOURCE_ORDER = (process.env.TLE_SOURCES || 'celestrak,celestrak_legacy
 
 function parseTleText(text, tleData, group) {
   // Build NORAD lookup set for fast matching
-  const knownNorads = new Set(Object.values(HAM_SATELLITES).map((s) => s.norad));
+  const knownNorads = new Set(
+    Object.values(HAM_SATELLITES)
+      .filter((s) => s && typeof s === 'object' && s.norad)
+      .map((s) => s.norad),
+  );
 
   const lines = text.trim().split('\n');
   for (let i = 0; i < lines.length - 2; i += 3) {
@@ -7603,7 +7584,9 @@ function parseTleText(text, tleData, group) {
       const alreadyExists = Object.values(tleData).some((sat) => sat.norad === noradId);
       if (alreadyExists) continue;
 
-      const hamSat = Object.values(HAM_SATELLITES).find((s) => s.norad === noradId);
+      const hamSat = Object.values(HAM_SATELLITES)
+        .filter((s) => s && typeof s === 'object' && s.norad)
+        .find((s) => s.norad === noradId);
       if (hamSat) {
         const key = name.replace(/[^A-Z0-9\-]/g, '_').toUpperCase();
         tleData[key] = { ...hamSat, tle1: line1, tle2: line2 };
--- a/src/plugins/layers/useSatelliteLayer.js
+++ b/src/plugins/layers/useSatelliteLayer.js
@@ -18,9 +18,13 @@ export const metadata = {
   },
 };
 
-export const useLayer = ({ map, enabled, opacity, config, units }) => {
+export const useLayer = ({ map, enabled, satellites: propSatellites, setSatellites, opacity, config, allUnits }) => {
   const layerGroupRef = useRef(null);
-  const [satellites, setSatellites] = useState([]);
+  const [internalSatellites, setInternalSatellites] = useState([]);
+  // Always render from internalSatellites (has enriched radio/speed data).
+  // If parent passes filtered satellites, use their names to filter ours so filter UI still works.
+  const allowedNames = propSatellites && propSatellites.length > 0 ? new Set(propSatellites.map((s) => s.name)) : null;
+  const satellites = allowedNames ? internalSatellites.filter((s) => allowedNames.has(s.name)) : internalSatellites;
 
   const [selectedSats, setSelectedSats] = useState(() => {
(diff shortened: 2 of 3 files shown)


accius:
Really nice additions here — the external satconfig.json, frequency/telemetry display, tail tracks, and the TLE local fallback are all solid improvements. Two bugs and some missing satellites need fixing before merge though.

**Bug 1: `allUnits` → `units` breaks distance display**

The new hook destructures `units` but `PluginLayer.jsx` passes `allUnits` to all layer hooks. So `units` is always `undefined` and the imperial/metric conversion in the info window silently fails. Either change the destructured prop back to `allUnits`, or update `PluginLayer.jsx` to also pass `units`.

**Bug 2: Satellite filters disconnected — double fetch**

The hook now manages its own state internally with `useState` for satellites and fetches `/api/satellites/tle` independently. But `App.jsx` still runs `useSatellites()` → `useSatellitesFilters()` and passes `filteredSatellites` down through `PluginLayer`. The new hook ignores all of that — so the TLE endpoint gets fetched twice, and any satellite filter UI in the parent becomes dead code. Either remove the old `useSatellites` / filter system from App.jsx, or keep accepting filtered data as a prop.

**10 satellites silently dropped from satconfig.json**

Missing from the previous `HAM_SATELLITES` that were actively tracked:
- **SO-124** (62690) — active FM repeater, added in v15.6.4
- **EO-88** (42017) — Nayif-1, Linear/Telemetry
- **TEVEL 1-8** (51013, 51069, 50988, 51063, 50998, 50999, 51062, 50989) — all 8 TEVEL FM satellites

The new weather sats are a great addition, but dropping these without changelog mention will confuse users who had them selected.

**`_SECTION_` keys create server-side noise**

The server does `Object.values(HAM_SATELLITES).map(s => s.norad)` in `parseTleText` and the debug endpoint. Section header strings like `"── High Priority..."` end up in there, calling `.norad` on a string (returns `undefined`, puts `NaN` in the Set). Won't crash but pollutes the NORAD lookup and debug output. Add a filter:
```js
Object.values(HAM_SATELLITES).filter(s => s && typeof s === 'object' && s.norad)
```

**UX: Ctrl+drag replaces title-bar drag**

Every other floating panel in the app (Lightning, RBN, WSPR, MUF, Gray Line) uses title-bar drag. The satellite info window now requires Ctrl+click to drag, which is undiscoverable and inconsistent with the rest of the UI.

Minor nits (non-blocking):
- Typo in satconfig.json `_SECTION_7`: "Reeserved" → "Reserved"
- `// OFFLINE BLOCK` comments look like debug markers — consider removing or converting to proper JSDoc
- Extra blank `//` line in the satellite audit comment block

The satconfig.json structure, README documentation, frequency fields, speed/altitude telemetry, and the glowing tail track are all well done. Looking forward to the fixed version!

creinemann:
I'll have a look today or tomorrow
The Nayif-1 (EO-88) CubeSat re-entered Earth's atmosphere and burned up on July 18, 2023
HADES-R (SO-124) is no longer on orbit, and TEVEL 1-8 are no longer in orbit as well,  so putting them back in the list is not necessary.

creinemann:
For Change Log:
vx.x — Satellite Layer Enhancements
•	Added new handling in server.js for support to pull satellite configuration from /src/satellites/satconfig.json
•	Added radio frequency display in satellite popup (downlink, uplink, tone, arm tone, HRPT, GRB, SD frequencies) sourced from satconfig.json
•	Added real-time satellite speed display in km/h and mph
•	Added lat/lon as separate labeled rows in popup
•	Lead track rendered as fading red gradient (20 min)
•	Tail track rendered as fading cyan/white glow gradient (45 min)
•	Fixed tle1/tle2 property name mismatch causing orbit propagation failure
•	Fixed NORAD ID lookup to correctly match satellites to radio config
•	Fixed allUnits prop reference for correct imperial/metric distance display
•	Fixed satellite filter UI reconnected to parent — eliminates double TLE fetch
•	Fixed popup drag to use title bar consistent with all other panels
•	Fixed _SECTION_ header keys polluting NORAD lookup in server.js
•	Added local TLE fallback (weather.tle) for offline/rate-limited operation
•	Added update-tle.js utility to refresh TLEs from Space-Track.org
•	Removed 9 decayed satellites from satconfig.json (8 TEVEL, EO-88, HADES-R (SO-124)
•	Corrected GK-2A NORAD ID from 43822 to 43823
•	Added satconfig.README.md user guide for adding/removing satellites



**Bug 1** — allUnits vs units Fixed. Hook prop destructuring changed from units back to allUnits. Both conv/distUnit references in the popup and the useEffect dependency array updated to match.
**Bug 2** — Double fetch / disconnected filters Fixed. Hook now accepts satellites/setSatellites from the parent via props. When the parent passes filtered satellites they are used directly. Internal fetch is kept as a fallback only when the parent passes nothing. This eliminates the double TLE fetch and reconnects the filter UI in App.jsx.
**Bug 3** — 10 satellites silently dropped Corrected. The satellites were removed intentionally as part of a separate cleanup — 8 TEVEL satellites confirmed re-entered July 2024, EO-88 (Nayif-1) re-entered July 2023 HADES-R (SO-124) no longer in orbit. All removals are documented in server.js and listed here for inclusion in the changelog (above for inclusion ). The _SECTION_7 header is retained for future additions.
**Bug 4** — _SECTION_ keys polluting NORAD lookup Fixed in server.js. Both Object.values(HAM_SATELLITES) calls on lines 7566 and 7582 now include .filter(s => s && typeof s === 'object' && s.norad) to skip section header strings before mapping or finding by NORAD ID.
**Bug 5** — Ctrl+drag inconsistency Fixed. Drag now triggers on the title bar span (.sat-drag-handle) consistent with all other floating panels in the app. Ctrl+click requirement removed.
**Nits**
•	Typo fixed: _SECTION_7 "Reeserved" → "Reserved"

accius:
The external satconfig.json is a great improvement. Radio frequency display, tail tracks, and the README documentation are all well done. The server.js defensive filters for _SECTION_ keys are correct. But there are two issues that need fixing before merge.

Blocking:
1. Double TLE fetch not resolved
The hook still fetches /api/satellites/tle every 5 seconds via setInterval(fetchSatellites, 5000) at line 387. Meanwhile useSatellites() in App.jsx fetches the same endpoint every 6 hours. Both run simultaneously. The fix from the last round uses propSatellites only to build an allowedNames filter set, but always renders from internalSatellites populated by the hook's own fetch. The parent's satellite data is never actually used, just its names. Either remove useSatellites() from App.jsx and let the hook fully own the data lifecycle, or have the hook enrich the parent's data without re-fetching.
2. QO-100 name breaks onclick handler
onclick="window.toggleSat('${sat.name}')" generates broken JavaScript for any satellite name containing an apostrophe. QO-100 (Es'hail-2) in satconfig.json produces onclick="window.toggleSat('QO-100 (Es'hail-2)')" which is a syntax error. The close button has the same issue. This is pre-existing but the new satconfig guarantees it fires. Fix with onclick="window.toggleSat(${JSON.stringify(sat.name)})" or escape quotes before interpolation. Affects both the info window card and the close button.

Non-blocking:
3. Speed display ignores allUnits — Altitude correctly respects allUnits.dist for imperial/metric, but speed always renders both km/h / mph hardcoded side by side. Should follow the same unit preference.
4. Console spam — console.warn('[SAT] No config match...') fires every 5 seconds for each TLE entry not in satconfig. Any bonus satellites returned by CelesTrak that aren't in satconfig.json will generate constant log noise. Should be removed or moved behind logDebug.
5. Hardcoded colors — The info window uses #00ffff, #ff4444, #00ff00 throughout instead of CSS variables (var(--accent-cyan), etc.). Checklist says no hardcoded colors.
6. Satellite names not escaped in innerHTML — ${sat.name} goes directly into innerHTML in several places. Safe for the current satconfig entries, but any future name containing < or & would inject HTML. Same esc() pattern used in WorldMap.jsx would be appropriate here.

accius:
I thought this was already merged when I refactored the code... it was not, but I put most of this into the refactor.  I am going to close this, can you check the new version on staging when you get a chance? (its not up yet but will be by 11pm EST time

PR #684, opened 2026-03-06
Improve TCI plugin diagnostics and Thetis compatibility

ceotjoe:
## What does this PR do?

Integrating better diagnosis options in TCI plugin

- Warn explicitly when ws npm package is missing and native WebSocket fallback activates
- Log which WebSocket implementation (ws npm vs native) is selected at startup
- Disable perMessageDeflate on ws connections for compatibility with non-standard TCI servers (e.g. Thetis)
- Add specific error branch for Sec-WebSocket-Accept handshake failures with actionable guidance

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/rig-bridge/plugins/tci.js
+++ b/rig-bridge/plugins/tci.js
@@ -48,16 +48,21 @@ module.exports = {
     // Resolve WebSocket implementation: prefer 'ws' npm package (works inside
     // pkg snapshots), fall back to Node 21+ built-in WebSocket.
     let WS;
+    let wsSource;
     try {
       WS = require('ws');
+      wsSource = 'ws npm';
     } catch {
+      console.warn('[TCI] ws npm package not found (run: npm install) — falling back to native WebSocket');
       if (typeof globalThis.WebSocket !== 'undefined') {
         WS = globalThis.WebSocket;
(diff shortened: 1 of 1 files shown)


accius:
Clean and well-scoped. The `perMessageDeflate: false` fix is the most valuable change here — Thetis (and likely other SDR TCI servers) don't handle WebSocket extension negotiation properly, which causes cryptic handshake failures. Disabling it only for the `ws` npm path is the right approach since native WebSocket doesn't support that option anyway.

The Sec-WebSocket-Accept error branch with actionable guidance is a nice touch — that's one of those errors that's impossible to debug without knowing what it means.

One tiny nit: `wsSource` could technically be `undefined` in the error handler's `Active WebSocket implementation: ${wsSource}` line if something unexpected happened during WS resolution. Not a real risk since `connect()` bails early on `!WS`, but you could default it: `let wsSource = 'unknown';` instead of `let wsSource;` for completeness.

LGTM — nice contribution.

Issue #685, opened 2026-03-07
[FEATURE] Multicast support for WSJT-X


alanhargreaves:
## Is your feature request related to a problem?

There has been a bit of discussion about getting orc to support listening on a multicast address for wsjt-x packets.

## Describe the Solution You'd Like

After (minimal) research, it looks like we should need probably another ".env" variable to specify the multicast address to use and then a little extra code in server.js where we set up the listener. This code would detect that we have a multicast address set and do the appropriate multicast join requests.

## Describe Alternatives You've Considered

Not really any alternative if folks want to have multiple applications receiving wsjt-x packets.

## Use Case

How would this feature benefit amateur radio operators? Be specific about the use case:

- [X] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [X] Contesting
- [X] POTA/SOTA
- [X] DXing
- [ ] Emergency communications
- [ ] Other: \***\*\_\_\_\*\***

## Would you be willing to help implement this?

- [X] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, but I'd use it!

## Similar Features in Other Software

There appears to be a number of other applications already capable of listening on a multicast address for this data

---

73!

trevlarson-ORE:
Hi @alanhargreaves - I'm pretty sure the functionality you're describing is already available.  My OHC instance is listening to a multicast address to get WSJT-X data from a different PC.  See https://github.com/accius/openhamclock/issues/83 & https://github.com/accius/openhamclock/issues/228.

73,
- Trevor, N6TJL

alanhargreaves:
It wasn't quite. We needed it to join the multicast group to work correctly (and make sure that I configured my wsjt-x to send out of the right ethernet).

Actually, after I did teh work to ad this it looks ike you are right. I suspect that the issue that folks may be having is not setting the ethernet up correctly on wsjt-x.

I'll pull that PR back to draft for now.

ceotjoe:
@alanhargreaves if you change something on the wsjtx stuff would be nice to let me know, because I already integrated wsjtx as a plugin within rig-bridge (maybe in future ohc-bridge). Plan is to concentrate all external connections to OHC here, to get rid of the bit fragmented approach we have currently. 

I want to avoid loosing functionality compared to the wsjtx-relay.

alanhargreaves:
Looking more closely at using multicast, ... 

[...]

PR #686, opened 2026-03-07
[BUG 685] Add full multicast support for WSJT-X, so that multiple apps can listen

alanhargreaves:
## What does this PR do?

As per https://github.com/accius/openhamclock/issues/685, add support for configurations having wsjt-x use multicast for reporting.

In the majority of cases listening on port 2237 with address `0.0.0.0` works just fine. Problems arise when we have more than one application listening for the packets. In this case all applications need to properly configure themselves for multicast use.

If you have more than one application wanting to receive the wsjt-x packets via multicast, you will need to configure `WSJTX_MULTICAST_ADDRESS` in `.env` and restart; ensuring that you use the same address that wsjt-x is configured to multicast to (e.g. `224.0.0.1`)

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Uncomment the WSJTX_MULTICAST_ADDRESS entry in .env, ensuring it matches what wsjt-x is using
2. Restart the server
3. Note that we get a server log message stating that we are listening on multicast ```
[WSJT-X] UDP listener on 0.0.0.0:2237
[WSJT-X] Joined multicast group 224.0.0.1
```
5. Verify that packets are coming in by either watching the WSJXT panel or running ```
$ curl localhost:3000/api/wsjtx/decodes
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,7 +502,7 @@ Live decoded FT8, FT4, JT65, JT9, and WSPR messages from WSJT-X, JTDX, or any co
 
 While the above configuration works just fine in a majority of cases, if you are running more than one multicast listener on a host (e.g. OpenHamClock and something like GridTracker2), then OpenHamClock needs to configure itself properly as a multicast listener.
 
-Uncomment the WSJTX_MULTICAST_ADDRESS line in `.env`, and make sure that the multicast address there matches what you have set in WSJT-X. e.g. `224.0.0.1`
+Uncomment the `WSJTX_MULTICAST_ADDRESS` line in `.env`, and make sure that the multicast address there matches what you have set in WSJT-X. e.g. `224.0.0.1`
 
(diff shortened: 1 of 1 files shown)


accius:
Looks good, approving. One small thing for a follow-up: the catch on addMembership at line ~11904 is missing the [WSJT-X] log prefix and doesn't include the error reason. Should be:
javascriptconsole.error(`[WSJT-X] Failed to join multicast group ${WSJTX_MULTICAST_ADDRESS}: ${e.message}`);
Every other log line in this section uses the [WSJT-X] tag, and without e.message the operator has no idea why the join failed (bad interface, invalid address, permissions, etc).

Issue #662, opened 2026-03-03
[FEATURE] Relocate Layout Lock Button so it's not consuming vertical screen space


trevlarson-ORE:
## Is your feature request related to a problem?

This is very minor - but in the Dockable Layout, the Layout Lock/Unlock button consumes vertical space which reduces the space available for the map and other panels.

## Describe the Solution You'd Like

Relocate the Layout Lock/Unlock button.  Perhaps relocate it under the Hide/Show UI interface.

<img width="1737" height="146" alt="Image" src="https://github.com/user-attachments/assets/7c0ed231-7980-4b80-a1e8-21420bf66719" />

---

73!
Trevor, N6TJL

denete:
I would like to suggest that this element be moved up to live in the header area. It would be layout dependent, but it feels like it fits better with the other buttons (settings, full, etc.).

denete:
I did some research on this topic and I'd like to share what I found. The single button loading into its own real estate in the display is definitely an anti-pattern. In reviewing the full OHC interface, there are really only a few interactive control locations: settings dialog, panels, and these few at the top right of the application.

Where the issue raised in this request is to not lose the row of vertical space while using the dockable layout, let's not ignore the lost horizontal space in the header itself from these buttons. I thought moving the layout lock/unlock button up into the header would be a potential solution, but for some display widths, that just throws it down under the other header items and we've lost the space again. Some language translations are going to lose that space due to the content size, etc....and, we haven't even thought about adding another button or element in the future.

Here comes the pitch. Let's look at moving those buttons off-screen and use a control (like a hamburger menu) to bring them into view. With this approach, we've saved horizontal and vertical space. OHC is a dashboard. We really shouldn't have elements that take up space without providing value in return.

trevlarson-ORE:
Moving it to the **Show/Hide UI** menu seems like a good place, which would work well with the existing OHC interface.

<img width="679" height="473" alt="Image" src="https://github.com/user-attachments/assets/99418ab1-2c5e-4e07-90e5-5813267e2e53" />

denete:
Context is important. The controls that float over the map are specific to the map itself.  Where the button currently sits, is actually appropriate to its context. 

Maybe we can think about this differently. **The control to reset the panel layout is in the settings dialog. Perhaps that is also where locking the panel layout should be found.**

The user adjusts, adds, resizes, etc. until they like their layout and then they lock it from the settings.

denete:
[...]

PR #687, opened 2026-03-07
[FEATURE] Relocate Layout Lock Button so it's not consuming vertical screen space 

denete:
## What does this PR do?

 Moved lock button to render in a new left border tabset using FlexLayout. This panel can collapse. The lock layout control can not be deleted from the border space as we don't want the user to get "stuck" without being able to unlock it.

This addresses #662.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Select the dockable layout
2. Click the "Lock Layout" in the left border and then the "Lock Layout" button
3. Confirm that the button can lock and unlock the dockable layout

## Checklist

- [✔] App loads without console errors
- [✔] Tested in **Dark**, **Light**, and **Retro** themes
- [✔] Responsive at different screen sizes (desktop + mobile)
- [~] If adding a panel: wired into Modern, Classic, and Dockable layouts
- [✔] No hardcoded colors — uses CSS variables (`var(--accent-cyan)`, etc.)
- [✔] No `.bak`, `.old`, `console.log` debug lines, or test scripts included

Screenshot 2026-03-07 013154
Screenshot 2026-03-07 013208

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -955,7 +955,7 @@ export const DockableApp = ({
                   ? 'Unlock layout — allow drag, resize, and close'
                   : 'Lock layout — prevent accidental changes'
               }
-              className="panel-layout-lock-button"
+              className={`panel-layout-lock-button ${layoutLocked ? 'locked' : 'unlocked'}`}
             >
               {layoutLocked ? '🔒' : '🔓'} Layout {layoutLocked ? 'Locked' : 'Unlocked'}
             </button>
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -367,15 +367,21 @@ body::before {
   display: flex;
   align-items: center;
   gap: 4px;
-  color: var(--text-secondary);
+  color: var(--text-muted);
   background: var(--bg-tertiary);
-  border: 1px solid var(--accent-amber);
+  border: 1px solid var(--border-color);
   border-radius: 4px;
   padding: 3px 8px;
   font-size: 11px;
   font-family: 'JetBrains Mono', monospace;
   cursor: pointer;
   margin: 1em auto;
+}
+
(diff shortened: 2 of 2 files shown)


accius:
Nice idea moving the lock toggle into a border tab instead of eating header space. The FlexLayout border styling and CSS additions are clean.
A few things to sort out before merge:
Existing users won't see it. The border tab only exists in DEFAULT_LAYOUT. Anyone with a saved layout in localStorage gets their saved layout from loadLayout(), not the default. Since the header button was also removed, those users have no lock toggle at all. Either inject the border tab into saved layouts during load (check if the lock-layout-tab id is missing and add it), or keep it in the "Add Panel" registry so users can restore it themselves.
Empty header div left behind. The button was removed but the parent div with its padding and flex styling is still rendered as an empty element. Should be removed entirely now that there's nothing in it.
Button doesn't visually change on lock state. The old button swapped to an amber background and border when locked. The new .panel-layout-lock-button class is static regardless of layoutLocked. Either add a conditional class (.panel-layout-lock-button--locked) or inline the amber highlight like the old version did.
No way to recover if closed. If a user closes the border tab, it's gone from their saved layout with no way to get it back without a full layout reset. Worth either preventing close on this specific tab or making sure it's available through the Add Panel menu as a fallback.

denete:
All comments addressed.

- Added check for border containers in saved dockable layout and add the default if missing.
- Lock Layout is in the panel registry and can be added as desired.
- "Locked" the border panel so that the default lock layout can not be deleted.
- Removed empty header div.
- Restored styling to "locked" state of button.

Screenshot 2026-03-09 141137

PR #688, opened 2026-03-07
Add optional fixed DX cluster spotter coordinates via env vars

on6zq:
## Summary

This PR adds two optional environment variables:

- DX_CLUSTER_SPOTTER_LAT
- DX_CLUSTER_SPOTTER_LON

When both are set, /api/dxcluster/paths uses these fixed coordinates for the DX cluster spotter origin and reports spotterLocSource as "env-fixed".

When unset, existing spotter location behavior remains unchanged.

## Motivation

OpenHamClock currently derives the DX cluster spotter location from grid / HamQTH / prefix logic. In some installations, operators want all DX cluster path origins to use a known fixed station location.

## Changes

- Added dxClusterSpotterLat and dxClusterSpotterLon to config parsing
- Added a fixed-origin override in /api/dxcluster/paths
- Updated the related configuration/documentation files

## Testing

Tested with:

- DX_CLUSTER_SPOTTER_LAT=50.70573
- DX_CLUSTER_SPOTTER_LON=4.39205

Verified /api/dxcluster/paths returns:
- spotterLat: 50.70573
- spotterLon: 4.39205
- spotterLocSource: "env-fixed"

Also verified that when the variables are unset, behavior falls back to the existing automatic spotter location logic.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -142,8 +142,11 @@ LAYOUT=modern
 # DX_CLUSTER_HOST=dx.someserver.com
 # DX_CLUSTER_PORT=7300
 
-# When using DX_CLUSTER_HOST for a local skimmer, make sur the paths on map have the right origin location.
-# When set, DX cluster path origins use those fixed coordinates. Otherwise default lookup is used.
+# When using DX_CLUSTER_HOST for a local skimmer, make sure the paths on map have the right origin location.
+# When set, DX cluster path origins use those fixed coordinates. Only use with a local skimmer (source=custom),
+# otherwise all spotters on all spots will be pinned to one location.
+# If not set, default lookup is used.
+
 # DX_CLUSTER_SPOTTER_LAT=50.123
 # DX_CLUSTER_SPOTTER_LON=4.123
 


on6zq:
Demo on https://openhamclock.on6zq.be/

accius:
Please point this PR to Staging. all PRs go to staging first and tested before pushing to the main branch

on6zq:
> Please point this PR to Staging

Done.  Thank you.

accius:
Small fixes needed before merge.

Good addition for local skimmer setups. The Number.isFinite(parseFloat(...)) handling is solid, both vars must be set for the override to apply, and source: 'env-fixed' makes it traceable in debug output.

Typo in .env.example: "make sur" should be "make sure".

Worth noting in the comment: the override applies to all spotters on every spot, not just your own callsign. That's correct for source=custom pointed at a local skimmer where every spot originates from you. But if someone sets these vars while using the default proxy/auto source, every spotter on every DX cluster spot gets pinned to one location and the map paths all fan out from a single point. Consider adding a note like "Only use with a local skimmer (source=custom)" to the .env.example comment so people don't set it by accident on a shared cluster feed.

on6zq:
> Small fixes needed before merge.

Fixed:
- corrected the .env.example typo ("make sur" -> "make sure")
- added a note that the fixed spotter override should only be used with a local skimmer (source=custom), since otherwise all spotters on all spots would be pinned to one location

PR #696, opened 2026-03-08
[BUG 695] Lightning layer is not using settable distance units

alanhargreaves:
## What does this PR do?

Updated to use the global `allUnits.dist` for 'metric' vs 'imperlal'.

If `allUnits.dist` is 'imperial', make the proximity radius `20 miles` rather than `30km` which translates to `18.6411 miles` and just looks strange.
    
Updated documentation to reflect this.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Set distance units to Imperial in the settings
2. Bring up the lightning layer (hotkey 'L')
3. Verify distances are listed in miles

## Checklist

Before:
image

After:
image


alanhargreaves:
I've just found some other places where it does 30km. I'm also now concerned that if we change the units, that distances saved in existing spots will be in the previous units.

Needs more work.

alanhargreaves:
Withdrawing this one as I've a better solution in a different branch.

Issue #695, opened 2026-03-08
[BUG] Lightning layer is not using settable distance units


alanhargreaves:
## Describe the Bug

Regardless of what the distance unit is set to (imperial or metric), `src/plugins/layers/useLightning.js` appears to only report in metric.

## To Reproduce

Set distance units to imperial and view the lightning layer. It still reports in km.

## Expected Behaviour

It should report in the units appropriate to the selected distance units.

---

73!

PR #697, opened 2026-03-08
[BUG 695] Lightning layer is not using settable distance units

alanhargreaves:
## What does this PR do?

Updated to use the global `allUnits.dist` for 'metric' vs 'imperlal' for distance display.
    
All comparisons are still done in km, but we display in the appropriate unit.

I also had to update the panel header to list the strike radius in both units, in order to have this fit, i had to override the font size down to 11px.

Addresses https://github.com/accius/openhamclock/issues/695

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Set distance units to Imperial in the settings
2. Bring up the lightning layer (hotkey 'L')
3. Verify distances are listed in miles

## Checklist

Before:
image

After:
image

### DIFF
--- a/src/plugins/layers/useLightning.js
+++ b/src/plugins/layers/useLightning.js
@@ -45,15 +45,17 @@ function lzwDecode(compressed) {
 }
 
 // Haversine formula for distance calculation
-function calculateDistance(lat1, lon1, lat2, lon2, unit = 'km') {
-  const R = unit === 'km' ? 6371.14 : 3963.1; // Earth radius in km or miles
+function calculateDistance(lat1, lon1, lat2, lon2) {
+  // Calculate the distance in both km and miles, returning both
+  const Rkm = 6371.14; // Earth radius in km
+  const Rmiles = 3963.1; // Earth radius in miles
   const dLat = ((lat2 - lat1) * Math.PI) / 180;
   const dLon = ((lon2 - lon1) * Math.PI) / 180;
   const a =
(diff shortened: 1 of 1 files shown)


accius:
Approve with minor fixes. Good change overall, distances now respect unit preferences cleanly.

Three small things:

Line 835 still passes `'km'` to `calculateDistance` which no longer accepts a unit parameter. Silently ignored but it's dead code:

```javascript
// current
const distance = calculateDistance(stationLat, stationLon, strike.lat, strike.lon, 'km');
// should be
const distance = calculateDistance(stationLat, stationLon, strike.lat, strike.lon);
```

The nearby strikes header crams both units in and shrinks the font to compensate. Since `allUnits` is right there, just show the user's preference:

```javascript
// current
`📍 Nearby Strikes(30km/18.6miles)`
// cleaner
`📍 Nearby Strikes (${allUnits.dist === 'metric' ? '30 km' : '18.6 mi'})`
```

Same applies to the "No strikes within 30km (18.6 miles)" string further down. Also missing a space before the parenthesis in `Strikes(30km`.

alanhargreaves:
I tried that. It doesn't work as it's only evaluated when the Panel is created. If you change units, it doesn't change with you.

Issue #692, opened 2026-03-08
[BUG] Header elements do not flow predictably


denete:
## Describe the Bug

When different screen resolutions / window sizes / device orientation changes are observed, the elements in the header display do not adapt in a predictable manner.

- In the "Tablet" and "Compact" layouts, header items simply scroll off the right side of the display if the size is too narrow to display all items. They should wrap to the next row.
- In the "Modern" and "Dockable" layouts, the action buttons will wrap to the next row and then jump to the far right side of the row if another element wraps also.
- The "Tablet" layout is missing the "update" button for local installations.
- The "Modern" layout loses the "Update" and "Support" buttons at smaller viewport widths.
- At one point, the "Support" button changes content to only show the heart emoji. The emoji is not properly centered in the button.

## To Reproduce

1. For the scenarios above, select the layout in question
2. Change either device orientations or browser window dimensions as described above
3. Behold the errors

## Expected Behavior

Stated above in the description

## Screenshots

<img width="859" height="341" alt="Image" src="https://github.com/user-attachments/assets/de1807e0-2641-4d7d-8b05-89ab5212a66c" />
<img width="1484" height="260" alt="Image" src="https://github.com/user-attachments/assets/4466fbad-f4fe-4263-b13c-ef93986456d2" />
<img width="626" height="327" alt="Image" src="https://github.com/user-attachments/assets/6abc5494-4ae1-4c55-8d38-082adee67588" />
<img width="361" height="266" alt="Image" src="https://github.com/user-attachments/assets/f92abd79-6f8e-4193-b1a2-a40b31d70cee" />

---

73!

denete:
In order to have a predictable control that takes up as little space as possible, I would like to move the action buttons (support, full-screen, settings, and update) to their own dialog box and replace them with a single hamburger menu.

PR #702, opened 2026-03-09
Bug 692 - Header elements do not flow predictably

denete:
## What does this PR do?

When different screen resolutions / window sizes / device orientation changes are observed, the elements in the header display do not adapt in a predictable manner. This PR keeps the content visible, organized, and responsive, while moving the action buttons out of the header area. The action buttons have been moved to a panel which is triggered by a single hamburger menu button in the header. This minimizes the space taken from the header's content area.

The pop-over menu (hamburger menu) displays the project logo, version number running with info link, and the action buttons that were previously in the header.

## Type of change

- [X] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test
In each layout...
1. Note the content in the header
2. Resize the window / change device orientation
3. Note the content stays visible
4. Select the hamburger menu, interact with each button

## Checklist

Screenshot 2026-03-09 114814
Screenshot 2026-03-09 114741
Screenshot 2026-03-09 114734
Screenshot 2026-03-09 114717
Screenshot 2026-03-09 114623

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -13,6 +13,7 @@ import ModernLayout from './layouts/ModernLayout.jsx';
 
 import { resetLayout } from './store/layoutStore.js';
 import { RigProvider } from './contexts/RigContext.jsx';
+import { AppMenuProvider } from './contexts/AppMenuContext';
 
 import {
   useSpaceWeather,
@@ -49,6 +50,7 @@ import WhatsNew from './components/WhatsNew.jsx';
 import { initCtyLookup } from './utils/ctyLookup.js';
 import { getAllLayers } from './plugins/layerRegistry.js';
 import ActivateFilterManager from './components/ActivateFilterManager.jsx';
+import AppMenu from './components/menus/AppMenu';
 
 // Load DXCC entity database on app startup (non-blocking)
 initCtyLookup();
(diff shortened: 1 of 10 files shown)


denete:
This addresses #692.

accius:
Good idea consolidating the header actions into a cleaner UI, especially for mobile and tablet. But on desktop this adds an extra click to everything and hides controls that power users hit constantly. A few things to address before merge:
Desktop shouldn't use the hamburger. Settings, Fullscreen, and Donate were all one-click buttons visible in the header. Now they're all behind a two-click hamburger menu on every screen size. For users who open Settings dozens of times a session, that's a real usability regression. The hamburger makes sense on mobile and tablet where header space is tight, but on desktop the buttons should stay inline. Use the existing breakpoint prop to switch between direct buttons (desktop) and hamburger (mobile/tablet).
Duplicate menu code. The hamburger menu + modal is copy-pasted nearly identically three times: Header.jsx, Classic compact, and Classic full. All three have the same logo, version display, button list, overlay, close handler, and escape listener. Extract it into a shared HeaderMenu.jsx component that all layouts import.
Debug artifact in Header.jsx. There's a backgroundColor: 'lightgray' on the header-content-column div. That'll paint a light gray bar across the header in every theme including dark.
z-index mismatch. The menu overlay is zIndex: 100000 in Header.jsx but zIndex: 10000 in ClassicLayout.jsx. Should be consistent. The DonateButton was bumped to 100001 to sit above the menu but that only works against the 100000 value.
HeaderMenu defined inside component body. In ClassicLayout, HeaderMenu is defined as a function inside the component, so it's recreated every render. Either pull it out as a standalone component or wrap it in useCallback/useMemo.
Confirm logo in production builds. public/img/ohc-logo-254x114.png works in dev, but verify it's accessible at /img/ohc-logo-254x114.png in the Docker production build since only dist/ and public/ (from the builder stage) get copied.

denete:
@accius What is your recommendation regarding this?

I can just address the issues with elements jumping from the right to left with resizes / orientation changes, and the icon centering for the smaller buttons. The hamburger menu was just an idea to remove elements that didn't add "dashboard visibility". Totally understand the point about power users.

1. Do you want hamburger menu implemented for compact and tablet layouts?
2. Do you want it to remain as buttons for all other layouts?

I'm happy to make changes, just let me know.

denete:
@accius I've addressed your comments. The hamburger menu will show by default in 'tablet' and 'compact' layouts. For 'modern' and 'dockable' layouts, the buttons will show in the header as long as the breakpoint stays at "desktop". If the window is resized to change breakpoint to "tablet" or "mobile", the hamburger will show.

Screenshot 2026-03-11 120737
Screenshot 2026-03-11 123214
Screenshot 2026-03-11 123316
Screenshot 2026-03-11 123328

denete:
@accius If you'd like, I can use shorter button text for the header buttons to take up less space, and then verbose for the pop-up menu. It'd be an easy update.

accius:
I still am not sold on this, i think this issue just gets solved with a more aggressive media tag.  everything stays how it is currently on staging/main, if the screen gets too small where the spacing is crushed, then it turns the buttons into a hamburger menu.

denete:
@accius Did you want me to change something, or just leave it as is. If the header content is getting too compressed, shortening the button text and possibly upping the tablet breakpoint value would help

denete:
**Some alternatives shown at 1080px screen width...**
Screenshot 2026-03-12 101219
Screenshot 2026-03-12 100408


**And at my full screen width...**
Screenshot 2026-03-12 101243

denete:
Either way, this new approach keeps the buttons in a predictable location and out of the content that the user wants to see.

accius:
Lets hold off on this for a bit until I have proper time to think about UX. This is a much bigger issue for me than just the top right buttons, its a whole UI issue of how things are currently organized and how space is utilized. May want to move these into tabs on the left since we put the lock there and its now a whole side of useless space being used.  Give me a few days, I travel tomorrow and can start looking at this again over the weekend.

PR #704, opened 2026-03-09
[enhancement-703] rebranded the lock layout panel to just layout,

denete:
## What does this PR do?

- Change the "lock layout" panel to be just "layout".
- Add a "reset panel layout" button to that panel.
- Utilize a disabled state when the layout is locked.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Select the dockable layout
2. Click the "Layout" in the left border and then the "Lock Layout" button
3. Confirm that the button disables and enables the reset layout button
4. Confirm that the reset layout button resets the dockable layout

## Checklist

Screenshot 2026-03-09 181918
Screenshot 2026-03-09 182002

### DIFF
--- a/src/DockableApp.jsx
+++ b/src/DockableApp.jsx
@@ -954,7 +954,7 @@ export const DockableApp = ({
 
         case 'layout':
           content = (
-            <>
+            <div className="panel" style={{ height: '100%' }}>
               <button
                 onClick={toggleLayoutLock}
                 title={
@@ -979,7 +979,7 @@ export const DockableApp = ({
                 </svg>
                 {t('station.settings.layout.reset.button')}
               </button>
-            </>
+            </div>
           );
           break;
 
--- a/src/store/layoutStore.js
+++ b/src/store/layoutStore.js
@@ -135,6 +135,20 @@ export const loadLayout = () => {
         if (parsed.borders.length === 0) {
           parsed.borders = DEFAULT_LAYOUT.borders;
           saveLayout(parsed);
+        } else {
+          // Migrate lock-layout → layout rename
+          let migrated = false;
+          for (const border of parsed.borders) {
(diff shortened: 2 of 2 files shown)


accius:
Good progress from the previous round. `resetLayout` is implemented and exported, i18n keys are in all 14 languages, `enableClose: false` prevents tab loss, and the reset button properly disables when locked.

One remaining issue: the component rename from `lock-layout` to `layout` will break existing saved layouts. Users who already have the border tab in localStorage still have `component: 'lock-layout'` in their saved JSON. The migration in `loadLayout` only checks `parsed.borders.length === 0`, which won't trigger for them since they already have borders. They'll see "Unknown panel" in the border tab until they manually reset.

Add a rename migration in `loadLayout` after the empty borders check:

```javascript
} else {
  // Migrate lock-layout → layout rename
  let migrated = false;
  for (const border of parsed.borders) {
    for (const child of border.children || []) {
      if (child.component === 'lock-layout') {
        child.component = 'layout';
        child.id = 'layout-tab';
        child.name = 'Layout';
        migrated = true;
      }
    }
  }
  if (migrated) saveLayout(parsed);
}
```

Also the `DEFAULT_LAYOUT` in staging still has the old names (`component: 'lock-layout'`, `id: 'lock-layout-tab'`, `name: 'Lock Layout'`) — the diff renames them but it looks like the staging code hasn't been updated yet. Make sure both the default and the migration use the new `layout` naming consistently.

denete:
@accius This has been updated with your fix for the saved layout migration to the new tab and panel name. Also fixed a styling issue.

Issue #724, opened 2026-03-13
[BUG] Profile save may capture stale unit setting (Metric/Imperial) from localStorage instead of current Settings state


on6zq:
**Issue**

It looks like profiles may not correctly save the current Metric/Imperial selection.

If I switch to Metric, save a profile, and later reload it, the profile appears to come back as Imperial.

**Suspected cause**

This seems to happen because profile save snapshots localStorage, while the Settings dialog keeps unit changes in component state until the main Settings save/apply action is used.

So if a profile is saved before Settings are explicitly saved, the snapshot may still contain the previous unit config.

**Relevant code**

In SettingsPanel.jsx, unit changes are held in local state and only persisted when handleSave() runs.

The profile Save action calls saveProfile(...) directly.

In profiles.js, saveProfile() uses takeSnapshot(), and takeSnapshot() reads persisted values from localStorage.

That means profile save may capture the last saved config, not the current edited Settings values.

**Reproduction**

- Open Settings
- Change units from Imperial to Metric
- Save a profile without first saving/applying Settings
- Reload that profile

Units appear to revert to Imperial

**Expected behavior**

Saving a profile should include the current Settings values, including unit preferences.

**Suggestion**

Before taking a profile snapshot, either
- persist the current Settings state first, or
- pass the current in-memory config into saveProfile()

A UI warning or disabling profile save while Settings are unsaved could also help prevent confusion.

alanhargreaves:
That's a really interesting point. I hadn't considered that when I split them out. I would assume that we would have seen a similar issue if the previous overall "units" had changed.

alanhargreaves:
@saschabuehrle, could you assign yourself this bug using the slash assign command?

PR #725, opened 2026-03-13
fix: persist pending settings before profile save snapshots

saschabuehrle:
## Bug
https://github.com/accius/openhamclock/issues/724 — saving a profile could capture stale unit settings because profile snapshots read localStorage, while Settings edits stay in component state until Save is applied.

## Fix
- Added a `persistCurrentSettings()` helper in SettingsPanel that writes the current in-memory form state via `onSave` without closing the dialog
- Call this helper immediately before every profile snapshot save action (new profile save and profile update)
- Keep existing Save button behavior by reusing the same helper there

## Testing
- `npx prettier --check src/components/SettingsPanel.jsx`

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/src/components/SettingsPanel.jsx
+++ b/src/components/SettingsPanel.jsx
@@ -358,7 +358,7 @@ export const SettingsPanel = ({
     }
   };
 
-  const handleSave = () => {
+  const persistCurrentSettings = () => {
     const rigPortValue = String(rigPort ?? '').trim();
     let nextRigPort = 5555;
     if (rigPortValue === '0') {
@@ -396,6 +396,10 @@ export const SettingsPanel = ({
         autoMode,
       },
     });
+  };
+
+  const handleSave = () => {
+    persistCurrentSettings();
     onClose();
   };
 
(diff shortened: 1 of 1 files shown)


alanhargreaves:
I would also consider some kind of notification that we have done a persistent save without the user pressing "Save".

e.g. 
```
Notice

As settings have changed, they have been saved before saving the profile.
```
or perhaps with better wording.

Or maybe even a note on the profiles page noting that current settings will be saved before a profile save.

saschabuehrle:
Thanks for the suggestion — good point.\n\nI kept this PR focused on fixing state consistency before profile save, but I’m happy to add a follow-up UI notice if you’d like that in this PR instead of a separate one.\n\nGreetings, saschabuehrle

accius:
The fix is clean and correct. The root cause is well-identified — saveProfile() snapshots from localStorage, but Settings edits live in React component state until Save is clicked. Flushing state via persistCurrentSettings() before the snapshot ensures the profile captures what the user actually sees on screen.
The refactor is minimal and easy to follow: extract persistCurrentSettings() from handleSave, call it before all three profile save paths (new, overwrite, update), keep handleSave as persist + close. No side effects, no behavior change for the existing Save button.
Alan's suggestion about a user notification is a nice-to-have, happy to take that as a follow-up PR.
One thing before we can merge: this PR targets main, please retarget it to Staging. All PRs go through staging first for testing before they hit main/production. If you update the base branch we're good to go.
Thanks for the contribution, @saschabuehrle, solid fix.

saschabuehrle:
Thanks for the thorough review and for validating the root cause/fix details.\n\nAppreciate you taking the time to verify this so carefully.\n\nGreetings, saschabuehrle

PR #732, opened 2026-03-14
fix: coordinates at lat=0 or lon=0 silently rejected

### LINKED ISSUE #729: [BUG] Valid coordinates at latitude 0 and/or longitude 0 are treated as missing in multiple places
There appears to be a cross-cutting bug where valid coordinates on the equator (`lat = 0`) or prime meridian (`lon = 0`) are treated as absent because the code uses JavaScript truthiness checks instead of explicit null/number checks.

As a result, legitimate locations such as:

- `0, 10`
- `10, 0`
- `0, 0`

can fail in multiple parts of the app.

### Root cause

Several code paths use patterns like:

```js
if (lat && lon) { ... }
```

or:

```js
if (!location?.lat || !location?.lon) return;
```

Since `0` is falsy in JavaScript, these checks reject valid coordinates.

#### `src/hooks/app/useDXLocation.js`

```js
if (parsed.lat && parsed.lon) return parsed;
```

#### `src/hooks/useWeather.js`

```js
if (!location?.lat || !location?.lon) return;
```

#### `src/components/SettingsPanel.jsx`

```js
if (lat && lon) {
  const grid = calculateGridSquare(lat, lon);
  setGridSquare(grid);
  setConfigLocator(grid);
}
```

#### `src/App.jsx`

```js
return wsjtx.decodes.filter((d) => d.lat && d.lon && d.timestamp >= ageCutoff);
```

#### `src/DockableApp.jsx`

```js
if (spot.lat && spot.lon) {
  handleDXChange({ lat: spot.lat, lon: spot.lon });
}
```

#### `src/layouts/ModernLayout.jsx`

```js
if (r.lat && r.lon) handleDXChange({ lat: r.lat, lon: r.lon });
```

#### `src/components/PSKReporterPanel.jsx`

```js
cursor: report.lat && report.lon ? 'pointer' : 'default',
```

### Likely user-visible effects

Depending on which code path is hit, valid zero-valued coordinates may:

- fail to restore from saved DX location
- prevent weather from loading
- prevent locator/grid recalculation
- make spots/reports/decodes non-clickable
- prevent DX target updates
- cause inconsistent UI behavior for valid map/report positions on the equator or prime meridian

### Expected behavior

Latitude `0` and longitude `0` should be treated as valid numeric coordinates everywhere.

### Actual behavior

Any logic using truthiness treats `0` as missing, causing valid coordinates to be ignored.

### Suggested fix

Replace truthiness checks with explicit coordinate validation, for example:

```js
lat != null && lon != null
```

or, where numeric validation is appropriate:

```js
Number.isFinite(lat) && Number.isFinite(lon)
```

This seems to affect multiple files, so it may be worth doing a repo-wide audit for patterns such as:

- `if (lat && lon)`
- `if (!lat || !lon)`
- `if (location?.lat && location?.lon)`
- `if (!location?.lat || !location?.lon)`

ceotjoe:
## What does this PR do?

Fixes #729. JavaScript's truthy check treats `0` as falsy, causing valid coordinates on the equator (lat=0) or prime meridian (lon=0) to be ignored as if missing.

- Replace all `if (lat && lon)` / `if (!lat || !lon)` guards with explicit `!= null` / `== null` checks across 9 files (18 occurrences total)
- Affected areas: DX location restore from localStorage, weather fetching, grid square calculation, map spot rendering (WorldMap, AzimuthalMap), PSK Reporter UI, lightning layer

## Test plan

- [ ] Set DE or DX location to lat=0 or lon=0 and verify it is saved/restored correctly
- [ ] Confirm weather loads for a location on the equator or prime meridian
- [ ] Confirm spots at lat=0 or lon=0 appear on WorldMap and AzimuthalMap
- [ ] Verify grid square calculates correctly when lat or lon is 0

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -392,7 +392,7 @@ const App = () => {
 
     // Map all decodes with resolved coordinates (CQ, QSO exchanges, prefix estimates)
     // WorldMap deduplicates by callsign, keeping most recent
-    return wsjtx.decodes.filter((d) => d.lat && d.lon && d.timestamp >= ageCutoff);
+    return wsjtx.decodes.filter((d) => d.lat != null && d.lon != null && d.timestamp >= ageCutoff);
   }, [wsjtx.decodes]);
 
   // Map hover


alanhargreaves:
Good catch.

accius:
nice work

PR #740, opened 2026-03-16
[translation] weather forecast cross-reference language JSON for day-of-week

MichaelWheeley:
## What does this PR do?

Instead of displaying abbreviated English day-of-week for weather forecast, cross references JSON language files for localized day-of-week. Updated keys for all languages using English as master, also for several languages updated the names used.

(During development locally only temporarily set forecast number of days from 3 to 8 to validate that keys for all days returned by weather service are valid.)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [X] Translation
- [ ] Map layer plugin

## How to test

1. Have loaded and every language from Settings, no loading errors observed.
2. English no longer displays abbreviated name but rather full-name.
3. Also try French, Spanish, German, Japanese

## Checklist

Screenshot 2026-03-15 131144

NOTE THAT THIS SCREEN SHOT is only a validation that all key names are functional. The code used to make this graphic was not checked-in so it is still a three day forecast.
Screenshot 2026-03-15 134947


MichaelWheeley:
Japanese
image

French
image

PR #741, opened 2026-03-16
Make the SOTA code more error resilient and tidy up debug messages

alanhargreaves:
### Make the SOTA code more error resilient
- moves the creation of the summit array and to be a function within Papa.parse upon completion.
- tells Papa.parse to skip empty lines
- catches if there are any CSV parsing issues

### Tidy up debug messages
- specifically we really don't need to know about the sample fields for POTA, SOTA and WWFF any more.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Note that `server/routes/spot.js` is less chatty

## Checklist

### DIFF
--- a/server/routes/spots.js
+++ b/server/routes/spots.js
@@ -3,8 +3,6 @@
  * Lines ~2657-2830 of original server.js
  */
 
-const { logWarn } = require('../utils/logging');
-
 module.exports = function (app, ctx) {
   const { fetch, logDebug, logErrorOnce } = ctx;
 
@@ -26,7 +24,6 @@ module.exports = function (app, ctx) {
 
       // Log diagnostic info about the response
       if (Array.isArray(data) && data.length > 0) {
-        const sample = data[0];
         logDebug(`[POTA] API returned ${data.length} spots.`);
 
         // Count coordinate coverage
@@ -66,7 +63,6 @@ module.exports = function (app, ctx) {
 
       // Log diagnostic info about the response
       if (Array.isArray(data) && data.length > 0) {
-        const sample = data[0];
         logDebug(`[WWFF] API returned ${data.length} spots.`);
       }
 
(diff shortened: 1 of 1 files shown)


accius:
checking the diff and this is what was found

⏺ Issues found:
                                                                                                               
  Race condition / async bug in SOTA CSV parsing — Papa.parse is being used with a complete callback (async style), but
  the code doesn't await or return a promise. The sotaSummits cache is set inside the callback, but the outer function  
  continues immediately. Previously, Papa.parse was used synchronously (returning results directly), which was correct.
  This change means sotaSummits may not be populated by the time it's needed.                                           
                                                     
  The fix should either:
  - Keep it synchronous (the original approach was fine), or
  - Wrap it in a Promise and await it

  logWarn import may be unnecessary or wrong — logWarn is imported from ../utils/logging at the top of the file, but
  logDebug and logErrorOnce come from ctx (line 4). This is inconsistent. Check whether logWarn is available on ctx
  already, or whether ../utils/logging is the right source.

  logWarn call has a formatting issue — logWarn('[SOTA] Papa.parse returned error code ', results.errors) passes the
  errors array as a second argument. This will likely log [object Object] or similar. Should probably be:
  logWarn('[SOTA] Papa.parse returned errors:', JSON.stringify(results.errors));

  sample variable is unused on line 175 — const sample = data[0] is assigned but never read (it was previously used for
  logging Object.keys(sample)). Should be removed.

  Removed error handling for CSV parse — The original synchronous version would naturally throw on malformed data
  (caught by the outer try/catch). The callback version swallows errors silently if results.errors is non-empty — it
  just logs a warning and skips, but sotaSummits stays stale/empty. The else branch after the error check also means
  partial parse results (where results.errors.length > 0 but results.data has valid rows) are completely discarded.

  The debug log for empty summits cache (logDebug('[SOTA] Summits Cache empty')) will now fire more often due to the
  async bug above, potentially masking the real problem.
  
 Take a look and let me know

alanhargreaves:
I'll take it back out of the async function and do the other cleanup.No rush on this one.

PR #780, opened 2026-03-18
POTA, SOTA, WWFF & WWBOTA refactoring and cleaning up

alanhargreaves:
## What does this PR do?

- Consolidate common code for POTA, SOTA, WWFF & WWBOTA map placement into a new function
- Have the respective panel (POTAPanel, SOTAPanel, ...) record a number of defaults for use by `ActivatorPanel` and `WorldMap`
- `ActivatorPanel`
   - Removed the word `ACTIVATOR` as it takes up unnecessary space
   - Display the callsign in the colour associated with activation type
   - Display the frequency and mode when hovering over the frequency
- `WorldMap`
   - Imports the map defaults from the Panel files
   - Displays the activation type and symbol as a part of the popup
   - Displays the above in the colour associated with the activation type as defined in the map defaults

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Select a POTA, SOTA, WWBOTA or WWFF panel that has activators
   - note that the word `ACTIVATOR` is no longer present
   - hover the mouse over the frequency and note the popup displaying the frequency and mode
   - note that the callsign is displayed in the activation type colour
2. If the map icon is not lit, click to to display the activations on the `WorldMap`
3. Click on one of the icons displayed on the `WorldMap`
4. note that in the popup
   - We see the icon displayed
   - We see the activation type displayed
   - The icon, the activation type and the callsign are displayed in the activation type colour

## Checklist

- [X] Tested in **Dark**, **Light**, and **Retro** themes - In WWFF, teh callsigns can be a little hard to read in teh Panel, but mouseover changes the colour to be readable

## Screenshots

Panel Before (eg SOTA)
image

Panel After (eg SOTA)
image

WorldMap Before (eg SOTA)
image

WorldMap after (eg SOTA)
image

### DIFF
--- a/src/components/ActivatePanel.jsx
+++ b/src/components/ActivatePanel.jsx
@@ -181,7 +181,7 @@ export const ActivatePanel = ({
               >
                 <span
                   style={{
-                    color: `${mapDefs.color}`,
+                    color: mapDefs.color,
                     fontWeight: '600',
                     overflow: 'hidden',
                     textOverflow: 'ellipsis',


accius:
Issues                                                            
                                                                                                                        
  Popup color is hardcoded to POTA green. In placeSpots(), the popup HTML still uses:                                   
  style="color:#44cc44; cursor:pointer"  
  This was correct when it was only in the POTA block, but now SOTA popups will show green callsigns instead of orange  
  (#ff9632), WWBOTA will show green instead of purple (#8b7fff), etc. The popup color should come from                
  mapDefaults.background (or a separate popupColor field).                                                              
                                                                    
  Labels don't escape callsigns. In the label HTML:                                                                     
  html: `...${spot.call}...`                                                                                            
  The WWFF block previously used ${esc(spot.call)} in labels. The refactored version drops the esc() call. Should be
  ${esc(spot.call)} for consistency and XSS safety.                                                                     
                                                                                                                        
  placeSpots and activateMapDefaults are defined inside the component render. activateMapDefaults creates new L.divIcon 
  instances on every render. Since the icons are static, consider moving the defaults object outside the component (or  
  memoizing) so Leaflet doesn't recreate identical DOM elements each cycle. Not a blocker but worth noting for
  performance.                                                                                                          
                                                                    
  placeSpots references closure variables (mapInstanceRef, esc, replicatePoint, bandPassesMapFilter, normalizeBandKey,  
  bandFromAnyFrequency, onSpotClick) — this works since it's defined inside the component, but it means it can't be
  extracted to a separate file later without refactoring. Fine for now.

alanhargreaves:
I've made the following updates
 - Relocate ActivatePanel definitions to the specific POTA, SOTA., WWBOTA and WWFF files an import them for WorldMap, outside of the component render
 - Have the map popup show the shape and name associated with the spot
 - Have the map popup use the correct colours for the Spot callsign
 - Escape callsign in the label

alanhargreaves:
A little more tidy up for `ActivatePanel`
- use the same name for the defaults `mapDefs` and import it with a rename in `WorldMap`
- don't pass components of `mapDefs` to ActivatePanel, just pass `mapDefs`.
- update `ActivatePanel` to use the passed in `mapDefs`

Is this looking close, @accius ?

Issue #802, opened 2026-03-22
[FEATURE] multicast options in wsjtx relay download request


alanhargreaves:
## Is your feature request related to a problem?

There is currently no automatic way to get multicast options into the downloaded wsjtx-relay code, short of editing the download.

## Describe the Solution You'd Like

The screen where we select the OS type that we will be running the relay on currently looks like:

<img width="318" height="397" alt="Image" src="https://github.com/user-attachments/assets/c2881a26-1867-474a-a063-1241884a160a" />

In this screen we could have a checkbox option to listen to a multicast address and an input for the address (which defaults to 224.0.0.1). These two values could be passed to the server as a part of the download request, which will make the appropriate modifications to the script provided to start the server.

These values should also be saved in the local config, and appropriately pre-polulated.

This same interface could also be used when we switch over to using `rig-bridge`.

## Describe Alternatives You've Considered

Editing the download script manually before starting it. Not really user friendly.

## Use Case

How would this feature benefit amateur radio operators? Be specific about the use case:

- [X] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [ ] Contesting
- [X] POTA/SOTA
- [X] DXing
- [ ] Emergency communications
- [ ] Other: \***\*\_\_\_\*\***

Many hams currently already have applications listening on a multicast address for wsjt-x UDP packets. This simplifies getting a relay up and running for them.

## Would you be willing to help implement this?

- [X] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, but I'd use it!

## Additional Context

Add any other context, screenshots, or mockups about the feature request here.

## Similar Features in Other Software

There are already applications out there (e.g. GridTracker2) that listen on the multicast address for wsjt-x UDP packets.

---

73!

alanhargreaves:
It might actually be easier to have wsjtxMulticast (or whatever it ends up being called) as a config setting, and report it in that download screen

accius:
merged your PR

accius:
Reopening as a bug

accius:
<img width="1421" height="598" alt="Image" src="https://github.com/user-attachments/assets/1f16618b-8732-4899-b6ec-0a0a6776e6c4" />

PR #804, opened 2026-03-22
[FEATURE 802] multicast options in wsjtx relay download request 

alanhargreaves:
## What does this PR do?

Makes it easier to use multicast with the wsjtx-relay code addressing https://github.com/accius/openhamclock/issues/802

The change
- Adds a configuration setting `wsjtxRelayMulticast` with default `{ enabled: false, address: '224.0.0.1' }`
- Allows this to be edited on the Stations panel within settings
- Uses this information in `PSKReporter` to pass a multicast address to the server API if one is set
- Updates `server/routes/wsjtx`
   - to modify the downloaded start script to include the multicast address if it has been provided
   - fixes a bug that `mktemp` doesn't work on macOS the same way that it works on linux for the downloadable run script
- Updates `wsjtx-relay/relay.js`
   - replaces instances of 127.0.0.1 with the supplied multicast address if it was given one
- The PSKReporter panel for WSJT-X, when a relay is not running
   - shows the multicast address if it has been set
   - has a link to the Installation instructions (wsjtx-relay/README.md) in the main branch on GitHub
- Updated the installation instructions in wsjtx-relay/README.md
- addressed a markdown lint issue in the main README.md

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Note that I've only tested this on MacOS. It should be absolutely fine on Linux and Windows as the only thing I've done in the start script is to add an argument for multicast if it is necessary. It might be nice if someone could check Windows.

Assuming that you do not currently have wsjt-x being listened for in your server

1. Verify that the Multicast settings are present on the Station page of Settings
2. Check the checkbox to enable it
3. Go to the WSJT-X tab in the PSK Reporter Panel
4. You should see
   - the multicast address listed
   - A link to the Installation instructions. Clicking this should open a new tab. Note that before merge, this will be the unchanged `wsjtx-relay/README.md`.
5. Click the box with your OS to download the run script
6. Look at the downloaded script in your Downloads folder
7. If you are on MacOS or Linux, you need to add execute permission to the script.e.g.
```bash
chmod +x start-relay.sh
``` 
8. Run the script. e.g.
```bash
./start-relay.sh
```
9. Once the script is running, and you have wsjt-x configured to send to a multicast address, you should see the reporter screen start to populate.

It would probably also be useful to follow the instructions in `wsjtx-relay/README.md` to ensure their correctness.

## Checklist

WSJT-X Panel before:
image

WSJT-X Panel after
image

Station Settings Panel
image


alanhargreaves:
For the What's New, I think worth noting that our longer term goal is to replace wsjtx-relay with the rig-bridge code. This was done to simplify getting multicast working with the existing code in the interim.

accius:
Hey @alanhargreaves, thanks for this — multicast support is a great addition. Found a few issues during review:

  1. Operator precedence bug in Windows script (breaking)

  In server/routes/wsjtx.js, this line:
  'echo     Address: ' + multicastAddress ? safeMulticastAddress : '127.0.0.1' + '   Port: 2237,'
  Due to JS operator precedence, 'echo     Address: ' + multicastAddress evaluates first as a truthy string, so the
  ternary always picks safeMulticastAddress — even when multicast is disabled. Needs parens:
  'echo     Address: ' + (multicastAddress ? safeMulticastAddress : '127.0.0.1') + '   Port: 2237'
  Also there's a stray comma at the end of 2237, inside the string.

  2. Double multicast join in relay.js

  socket.addMembership(config.maddr) is called twice — once around line 518 (before bind) and again in the bind callback
   around line 650. The first one will fail since the socket isn't bound yet. The pre-bind call should be removed.

  3. Commented-out code in relay.js

  // Bind to all interfaces so WSJT-X can reach it from any address
  // socket.bind(config.port, '0.0.0.0');
  This dead code around line 516-517 should be cleaned up.

  4. No multicast address validation

  req.query.multicast is shell-sanitized but never validated as an actual multicast address. Should check that the first
   octet is 224-239:
  const parts = multicastAddress.split('.').map(Number);
  if (parts.length !== 4 || parts[0] < 224 || parts[0] > 239) {
    return res.status(400).json({ error: 'Invalid multicast address (must be 224.0.0.0–239.255.255.255)' });
  }

  5. Missing prop default in PSKReporterPanel

  The new wsjtxRelayMulticast prop doesn't have a default value. If a layout doesn't pass it,
  wsjtxRelayMulticast.enabled will throw. Add a default:
  wsjtxRelayMulticast = { enabled: false, address: '224.0.0.1' },

  Everything else looks good — the macOS mktemp fix, reuseAddr: true, shell sanitization, and the config/settings
  integration are all solid. Just need these fixes before we can merge. 73!

alanhargreaves:
1. Oops, I thought I caught all of those.
2. There is only one call to socket.addMembership() and that is on line 523. It succeeds as we don't get a caught error. The call on line 660 is to socket.setMulticastLoopback(). "The setMulticastLoopback(flag) method controls whether multicast packets sent by the local application are also received by local sockets on the same machine. This is useful if you have multiple processes or multiple sockets within the same process that need to receive the same multicast data."
3. Done
4. I had meant to do that. Good catch. It would be nice to do it in the settings page, but the way that the input works makes it difficult.
5. I did make sure that everything that currently uses PSKReporterPanel passes it, but your point is well taken

So basically, I've addressed all but item 2, which was not an issue.

alanhargreaves:
I've just merged the current staging in to this branch (no conflicts).

PR #827, opened 2026-03-24
Match ' QRT' in the comment as well as the existing QRT variants

alanhargreaves:
## What does this PR do?

Adds another variant of 'QRT' to match a spot comment in POTA, SOTA and WWFF.

..., and of course just after creating the PR, I notice someone in WWFF put 'qrt' in their comments.

I deliberately chose to match against '<space>QRT', so we would pickup comments that had an end of string, or some kind of punctation following 'QRT'.

I noticed a SOTA spot being shown when the comment string was "Now QRT many thanks chasers".

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

We don't see spots where the comment includes the string " QRT". 

## Checklist

### DIFF
--- a/src/hooks/useWWFFSpots.js
+++ b/src/hooks/useWWFFSpots.js
@@ -44,7 +44,7 @@ export const useWWFFSpots = () => {
           const validSpots = spots
             .filter((s) => {
               // Filter out QRT (operator signed off)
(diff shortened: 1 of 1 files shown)


alanhargreaves:
Good point. I also eliminated the extra variable.

PR #833, opened 2026-03-24
Fix contest QSO DX target updates

### LINKED ISSUE #832: Contest QSOs do not move the main DX target marker
## Summary
When the `Contest QSOs` map layer is enabled, newly received N1MM/DXLog QSOs are drawn as markers/arcs on the map, but the main DX target marker does not move to the latest worked station.

## Impact
This leaves the dashboard in an inconsistent state:
- the contest overlay shows the latest QSO location
- the main DX target marker still points to an older/manual target
- propagation, beam heading, and DX weather panels continue using the stale DX target

## Reproduction
1. Enable N1MM/DXLog QSO ingestion and log contest QSOs with valid lat/lon resolution.
2. Enable the `Contest QSOs` layer.
3. Log a new QSO.
4. Observe that the contest arc/marker appears, but the main DX target marker does not update.

## Expected behavior
The latest contest QSO with valid coordinates should update the main DX target, while still respecting the existing DX lock behavior.

## Notes
The current contest layer only renders overlay geometry. It does not have access to the app-level DX target update callback, so it cannot move the shared DX target state.

s53zo:
## Summary
Fixes the contest QSO map layer so the latest located contest contact updates the main DX target marker instead of only drawing overlay geometry.

Fixes #832.

## Problem
`/api/contest/qsos` data was being rendered by the `Contest QSOs` layer, but that plugin only drew markers/arcs. It did not have access to the shared DX target update callback, so the main DX target state stayed stale even when new QSOs were plotted successfully.

## Changes
- pass `onDXChange` through the plugin layer plumbing
- auto-target the latest contest QSO with valid coordinates
- avoid repeated scans and writes by tracking the last handled located QSO
- prevent locked contest QSOs from replaying and stealing the DX target after unlock
- drop unrelated formatting-only changes from this PR

## Verification
- `npm run build`

## Notes
DX lock is still enforced by the existing app-level `onDXChange` handler. This PR only addresses the DX target update behavior. It does not change `/api/contest/qsos` caching.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main, develop]
   pull_request:
-    branches: [main]
+    branches: [main, Staging]
 
 permissions:
   contents: read


accius:
also please point this at the staging branch not main

s53zo:
Addressed the review feedback in this branch update:

- retargeted the PR to `Staging`
- removed the unrelated formatting-only changes from this PR
- narrowed the plugin plumbing to pass only `onDXChange`
- removed `dxLocation`/`dxLocked` from the contest auto-target effect dependencies
- replaced `[...]qsos.reverse().find(...)` with a reverse index scan helper
- removed the redundant `alreadySelected` epsilon check
- track the last handled located QSO so locked contest QSOs are not replayed and steal the DX target after unlock

Verification:
- `npm run build`

PR #842, opened 2026-03-26
feat: optimistic UI, SSE push, long-poll relay and dev server fix for…

ceotjoe:
### Problem

Rig control panel and On Air panel updates felt sluggish and async — frequency and PTT changes sent via the cloud relay plugin took far longer than the 2 s push interval to appear in the UI. The root causes were:

1. **Client polled for state** — the browser polled `/relay/state` every 1–2 s, introducing a full poll cycle of latency before any change was visible
2. **Commands polled too** — the rig-bridge cloud-relay plugin fetched `/relay/commands` on a fixed timer, so a command sent just after a poll had to wait up to 2 s before rig-bridge even saw it
3. **No optimistic feedback** — the UI showed nothing until the round-trip completed; clicking PTT or tuning felt unresponsive
---

#### 1 · SSE push instead of polling (`RigContext.jsx`, `server/routes/rig-bridge.js`)
A new `GET /api/rig-bridge/relay/stream` Server-Sent Events endpoint fans out every rig state push from the cloud relay to all connected browser sessions instantly. The browser opens a persistent `EventSource` connection instead of polling. State changes appear in the UI as fast as the network allows — typically well under 100 ms from the time the rig-bridge device pushes.

- Exponential back-off reconnect (3 s → 60 s) on SSE error
- Automatic fallback to `/relay/state` polling (every 5 s, stops after 3 failures) while SSE reconnects, so updates are never completely lost during a blip
- Full cleanup on unmount (EventSource, retry timers, fallback interval)

#### 2 · Long-poll command delivery (`server/routes/rig-bridge.js`, `rig-bridge/plugins/cloud-relay.js`)
`GET /api/rig-bridge/relay/commands` now supports `?wait=<ms>` (max 30 s). When the rig-bridge plugin calls this endpoint with `?wait=25000`, the server **holds the connection open** until a command arrives, then resolves immediately — delivering the command in ~100 ms instead of up to 2 s. If no command arrives within the wait window, the server returns an empty list and the plugin immediately re-opens the connection.

- `notifyCommandWaiters()` resolves all waiting connections the instant a command is posted via `/relay/command`
- A 60 s TTL cleanup interval prevents leaked waiters from zombie sessions
- The old fixed-interval poll loop in the plugin is fully replaced

#### 3 · Optimistic UI updates with server-confirmed rollback (`RigContext.jsx`)
Frequency and mode changes in cloud relay mode now update the UI **immediately** when the user acts, before the relay round-trip completes. If SSE confirmation doesn't arrive within 3 s (e.g. the relay is unreachable), the display rolls back to the last server-confirmed value — preventing a stuck stale display without confusing the user with phantom values.

#### 4 · Graceful error handling — rig-bridge never harms the main app
- All server route handlers wrapped in `try/catch` — no unhandled exceptions can propagate to Express's default 500 handler
- All client-side fetch/SSE errors caught and swallowed at the rig-bridge boundary
- `useRotator` now applies exponential back-off after the first failure (doubles each attempt, caps at 30 s) — stops flooding the console when the rotator is offline
- The main application continues to function normally if any rig/rotator/relay endpoint is unavailable

#### 5 · Updated rig-bridge README
Full documentation of the cloud-relay plugin, long-poll command flow, all new and changed API endpoints, configuration keys, and a sequence diagram of the push/command cycle.

---

### Files changed

| File | Change |
|---|---|
| `src/contexts/RigContext.jsx` | SSE connection, optimistic updates, rollback timers, fallback poll |
| `src/hooks/useRotator.js` | Exponential back-off on repeated failures |
| `server/routes/rig-bridge.js` | `/relay/stream` SSE endpoint, long-poll `/relay/commands`, `notifyCommandWaiters`, try/catch hardening throughout |
| `rig-bridge/plugins/cloud-relay.js` | Long-poll command loop replaces fixed-interval poll |
| `rig-bridge/README.md` | Full documentation update |

---

### Testing

- [ ] PTT toggle reflects in UI within one push cycle (≤ 2 s) or near-instantly with SSE live
- [ ] Frequency tuning reflects immediately (optimistic), confirmed by SSE
- [ ] Killing the Express server mid-session: fallback poll activates, SSE reconnects with back-off, no console errors from other app features
- [ ] `useRotator` stops polling after rotator goes offline, resumes when it comes back

## Checklist

### DIFF
--- a/rig-bridge/plugins/cloud-relay.js
+++ b/rig-bridge/plugins/cloud-relay.js
@@ -22,7 +22,6 @@
  *   apiKey:         string   Authentication key for the relay
  *   session:        string   Browser session ID for per-user isolation
  *   pushInterval:   number   State push interval in ms (default: 2000)
- *   pollInterval:   number   Command poll interval in ms (default: 250)
  *   relayRig:       boolean  Relay rig state (default: true)
  *   relayWsjtx:     boolean  Relay WSJT-X decodes (default: true)
  *   relayAprs:      boolean  Relay APRS packets (default: false)
@@ -54,7 +53,6 @@ const descriptor = {
     const apiKey = cfg.apiKey || '';
     const session = cfg.session || '';
     const pushInterval = cfg.pushInterval || 2000; // Fallback interval for data batches
-    const pollInterval = cfg.pollInterval || 250; // Command poll — keep low for responsive rig control
     const { state, pluginBus, onStateChange, removeStateChangeListener } = services;
 
     let pushTimer = null;
@@ -171,7 +169,7 @@ const descriptor = {
 
     // Long-poll loop — holds the connection open for up to LONG_POLL_WAIT ms.
     // The server resolves the request immediately when a command is queued,
-    // so latency is ~network RTT rather than up to pollInterval.
+    // so latency is ~network RTT rather than the push interval.
     // On timeout (no commands) it restarts immediately. On network error it
     // waits 1 s before retrying to avoid hammering a temporarily unreachable server.
     const LONG_POLL_WAIT = 28000; // ms to hold open (server caps at 30 s)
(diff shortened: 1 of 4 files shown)


ceotjoe:
@accius I hope all findings are addressed.

Issue #859, opened 2026-04-01
[BUG] propagation band conditions do not match header display


navyitc:
## propagation band conditions do not match header display

Banner at top of window may not be properly updating. K shows "__" and SFI, A and other indexes do not match the indexes as shown in Propagation - Band Conditions frame leading to questionable Propagation status. The Banner appears to be correct based on other sources (WWV).

## Screenshots

<img width="628" height="894" alt="Image" src="https://github.com/user-attachments/assets/36844ed1-fea7-404a-abe3-3230cb78e096" />

## Environment

- **OS**: WIN11
- **Browser**: Edge

- **OpenHamClock Version**: 26.1.3
- **Running As**: Web browser

---

73!

ceotjoe:
Should be fixed with https://github.com/accius/openhamclock/pull/865

PR #865, opened 2026-04-02
Fix/propagation solar index display

ceotjoe:
## Summary

Fixes several issues (reported in https://github.com/accius/openhamclock/issues/859) with solar index display across the header banner and propagation/band-conditions panel, all triggered by two upstream NOAA API changes that went unnoticed.

- **K always showed `--` or `0`** — NOAA changed `noaa-planetary-k-index.json` and the forecast endpoint from an array-of-arrays format to an array-of-objects format. The old parsers read `d[0]`/`d[1]` on each row; with objects those return `undefined` → `NaN` → `0` (or `2` in the propagation route's hardcoded fallback). Fixed in all three independent places that fetched this endpoint: `space-weather.js`, `propagation.js`, and the browser-side `useSpaceWeather` hook.

- **SFI showed 112 in the panel footer instead of 140** — `f107_cm_flux.json` is not chronologically sorted; the last element in the array was a February reading rather than the current one. `getSolarData()` was blindly taking `data[last]`. Fixed by reducing over the array for the entry with the highest `time_tag`. The SFI history chart in `space-weather.js` was also sorting incorrectly for the same reason.

- **K showed `--` when Kp = 0** — The server used `|| null` when resolving `kp.current` from history, so a legitimately quiet geomagnetic day (Kp = 0) caused the value to be stored as `null` and rendered as `--`. Also fixed `|| '--'` → `?? '--'` on all numeric index display sites (Header, SpaceWeatherPanel, ClassicLayout) per project zero-value rules.

- **Band conditions panel could silently show indefinitely stale N0NBH data** — When `hamqsl.com` was unreachable the error fallback returned cached data without updating the cache timestamp, so every subsequent retry failed and kept serving the same arbitrarily old data. Added a 4-hour maximum stale TTL (503 beyond that) and a `⚠ stale (Nm)` badge in the BandConditions and PropagationPanel headers when fallback data is being served.

## Changed files

| File | Change |
|---|---|
| `server/routes/space-weather.js` | NOAA Kp object-format parser; SFI history sort; `?? null` zero-value fixes; N0NBH max-stale TTL + `fetchedAt`/`stale` fields |
| `server/routes/propagation.js` | NOAA Kp object-format parser; SFI latest-entry fix via `max(time_tag)` |
| `src/hooks/useSpaceWeather.js` | NOAA Kp object-format parser (`d.Kp` for new format) |
| `src/hooks/useBandConditions.js` | Expose `fetchedAt` and `stale` through `extras` |
| `src/components/Header.jsx` | `\|\|` → `??` for SFI fallback chain |
| `src/components/SpaceWeatherPanel.jsx` | `\|\|` → `??` for SFI and K-index display |
| `src/components/BandConditionsPanel.jsx` | Accept `extras` prop; show stale badge |
| `src/components/PropagationPanel.jsx` | Show stale badge in bands view header |
| `src/layouts/ClassicLayout.jsx` | `\|\|` → `??` for SFI (×3) and A-index; stale badge on inline band conditions header |

## Test plan

- [ ] Confirm header banner shows current K and SFI matching N0NBH / WWV
- [ ] Confirm propagation panel footer shows the same SFI and K as the header
- [ ] Verify K = 0 displays as `0` rather than `--` during a geomagnetically quiet period
- [ ] Simulate N0NBH unavailability: confirm stale badge appears and data older than 4 h returns 503
- [ ] Hard-refresh after server restart and confirm all panels agree

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist


ceotjoe:
@accius please check if these fixes are fine. Values are displayed correctly now from my perspective. But I don't want to break something by accident. :-)

alanhargreaves:
We've had people in FB asking about the kIndex not displaying. Looks like they changed the data format. Jörg addresses this in this PR.

PR #877, opened 2026-04-04
[Feature/satellite] styling visibility - other updates

### LINKED ISSUE #855: [IMPROVEMENT / QUESTION] satellites ? remove altitude and speed ?
I'm adding the satellite range, range rate and doppler factor which shows only when the satellite is above the horizon. As shown the window is getting pretty cluttered. Although they are interesting do the satellite altitude and speed have a useful purpose?

Image

### LINKED ISSUE #889: [IMPROVEMENT] satellite details colors
choice of colors in satellite windows needs attention

from https://github.com/accius/openhamclock/pull/877

> ### Hardcoded theme-unaware colors
> 
> Replacing `#hex` with `rgba()` equivalents doesn't solve the underlying theme issue — `rgba(48, 33, 21, .8)` and `rgba(0, 248, 0, .5)` are still dark-theme-only colors. I know inline HTML templates make CSS variables harder to use, but these section backgrounds will look off in Light and Retro. Could you at least use very low-opacity neutrals (e.g. `rgba(255,255,255,0.05)` / `rgba(0,0,0,0.1)`) that work across themes, or pass theme-aware colors from the component?

MichaelWheeley:
## What does this PR do?
- introduces station altitude and minimum elevation boxes in the satellite settings. Altitude used in calculations. Minimum elevation used rather than zero degrees to determine visibility of satellite
- changes satellite details to highlight sections satellite location, tracking/visibility, satellite miscellaneous
- highlights tracking details when satellite is visible

reference https://github.com/accius/openhamclock/issues/855

(miscellaneous tle1, tle2 passed - will use them in next iteration)

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist

station altitude and minimum elevation boxes
image

not visible
image

satellite above 0deg but below minimum threshold to be considered visible (I have it set to 5.1deg)
image

satellite visible, elevation shown as 5deg however internally it is greater than or equal to 5.1deg. Tried it but displaying decimal degrees here doesn't look good. I could have made the minimum elevation setting an integer but was happy with this arrangement.
image


accius (changes requested):
Nice feature — station altitude and min elevation are solid additions for sat tracking. A few things to address before merge:

### Must fix

**Broken HTML in satellite card** — The 3 inner `<table>` elements are nested directly inside the outer `<table>` without proper `<tr><td>` wrapping. This is invalid HTML and will render inconsistently across browsers. Either close the outer table before starting the sections, or wrap each inner table in `<tr><td colspan="2">`.

**Duplicate `style` attribute on position table:**
```html
<table style="background-color: #302115; width:100%; font-size:11px; style="color: #888; padding:2px 0;"">
```
Second `style=` is ignored — the `color: #888` never applies.

**`valueAsNumber` with `??`** — `e.target.valueAsNumber` returns `NaN` when empty, not `null`/`undefined`, so `?? 100` won't catch it. `NaN` will propagate to the saved config. Use `|| 100` or an explicit `isNaN` check instead.

**Hardcoded background colors** (`#302115`, `#233b46`, `#00f800`, `#252e17`) — these will look wrong in Light and Retro themes. Should use CSS variables or at least `rgba()` with low opacity over the existing background.

### Should fix

**`minElev` default mismatch** — `SettingsPanel` defaults to `5.0` but `useSatelliteLayer.js` uses `config?.satellite?.minElev || 0`. A user who hasn't opened settings will see different visibility behavior between the layer and the hook.

**Negative min elevation allowed** — The input allows `-89` which would mark a satellite as "visible" when below the horizon. Should probably clamp to `0–89`.

### Nit

Squash the merge commits on merge — 13 commits (4 are merge syncs) is noisy for the history.

MichaelWheeley:
> ### Must fix

> **Broken HTML in satellite card** — The 3 inner `<table>` elements are nested directly inside the outer `<table>` without proper `<tr><td>` wrapping. This is invalid HTML and will render inconsistently across browsers. Either close the outer table before starting the sections, or wrap each inner table in `<tr><td colspan="2">`.

***Handled by removal of inner tables with additional styling on <tr> elements.***

> **Duplicate `style` attribute on position table:**
> 
> ```html
> 
> <table style="background-color: #302115; width:100%; font-size:11px; style="color: #888; padding:2px 0;"">
> 
> ```
> 
> Second `style=` is ignored — the `color: #888` never applies.

***fixed duplicate style property***

> **`valueAsNumber` with `??`** — `e.target.valueAsNumber` returns `NaN` when empty, not `null`/`undefined`, so `?? 100` won't catch it. `NaN` will propagate to the saved config. Use `|| 100` or an explicit `isNaN` check instead.

***replaced ?? with || incase of NaN***

>**Hardcoded background colors** (`#302115`, `#233b46`, `#00f800`, `#252e17`) — these will look wrong in Light and Retro themes. Should use CSS variables or at least `rgba()` with low opacity over the existing background.

***removed all # hardcoded colors, replaced with rgba(), modify opacity on selected colors***

>### Should fix

>**`minElev` default mismatch** — `SettingsPanel` defaults to `5.0` but `useSatelliteLayer.js` uses `config?.satellite?.minElev || 0`. A user who hasn't opened settings will see different visibility behavior between the layer and the hook.

***minElev consistency default to 5.0***

> **Negative min elevation allowed** — The input allows `-89` which would mark a satellite as "visible" when below the horizon. Should probably clamp to `0–89`.

***Restricted minElev to -5 to 89 allowing for marginal below the horizon setting.
There was also a problem in the onChange functionality when the value was zero - which was getting set to 5. Fixed with isNaN().***

MichaelWheeley:
image
image

accius (changes requested):
Thanks for the updates Michael — good progress. A few things still need fixing before this can merge:

## Must fix

### Corrupted color strings
The hex-to-rgba find-and-replace left trailing characters on two lines in `useSatelliteLayer.js`:

```js
color: 'rgba(255, 255, 255, 1)fff',   // ~line 379 — trailing "fff"
color: 'rgba(255, 255, 255, 1)f00',   // ~line 408 — trailing "f00"
```

The first should be `'rgba(255, 255, 255, 1)'`. The second was originally `#ffff00` (yellow for lead track) — should be `'rgba(255, 255, 0, 1)'`, not white.

### `stationAlt` onChange still uses `??`
You fixed `minElev` to use `isNaN()` but `stationAlt` still has:

```js
onChange={(e) => setStationAlt(e.target.valueAsNumber ?? 100)}
```

`valueAsNumber` returns `NaN` when empty, not `null`/`undefined`, so `??` won't catch it. Should match the `minElev` pattern:

```js
onChange={(e) => setStationAlt(isNaN(e.target.valueAsNumber) ? 100 : e.target.valueAsNumber)}
```

### `satelliteConfig` not null-safe
In `useSatellites.js`, `elevation >= satelliteConfig.minElev` will throw if `config.satellite` is undefined (user hasn't opened settings yet). Needs a guard:

```js
const isVisible = elevation >= (satelliteConfig?.minElev || 5.0);
```

## Should fix

### Lat/lon default changed from `??` to `||`
The `SettingsPanel` lines for lat/lon were changed from `config?.location?.lat ?? 0` to `|| 0`. Since `0` is a valid coordinate, `??` was the correct operator — `||` treats `0` as falsy. It happens to work here because the fallback is also `0`, but it's semantically wrong and could mask bugs if the default ever changes. Please revert to `??`.

### Hardcoded theme-unaware colors
Replacing `#hex` with `rgba()` equivalents doesn't solve the underlying theme issue — `rgba(48, 33, 21, .8)` and `rgba(0, 248, 0, .5)` are still dark-theme-only colors. I know inline HTML templates make CSS variables harder to use, but these section backgrounds will look off in Light and Retro. Could you at least use very low-opacity neutrals (e.g. `rgba(255,255,255,0.05)` / `rgba(0,0,0,0.1)`) that work across themes, or pass theme-aware colors from the component?

MichaelWheeley:
thank-you for your patience !

> ## Must fix
>  
> ### Corrupted color strings
> 
> The hex-to-rgba find-and-replace left trailing characters on two lines in `useSatelliteLayer.js`:
>  
> ```js
> color: 'rgba(255, 255, 255, 1)fff',   // ~line 379 — trailing "fff"
> color: 'rgba(255, 255, 255, 1)f00',   // ~line 408 — trailing "f00"
> ```
>
> The first should be `'rgba(255, 255, 255, 1)'`. The second was originally `#ffff00` (yellow for lead track) — should be `'rgba(255, 255, 0, 1)'`, not white.

done

> ### `stationAlt` onChange still uses `??`
> 
> You fixed `minElev` to use `isNaN()` but `stationAlt` still has:
> 
> 
> 
> ```js
> 
> onChange={(e) => setStationAlt(e.target.valueAsNumber ?? 100)}
> 
> ```
> 
> 
> 
> `valueAsNumber` returns `NaN` when empty, not `null`/`undefined`, so `??` won't catch it. Should match the `minElev` pattern:
> 
> 
> 
> ```js
> 
> onChange={(e) => setStationAlt(isNaN(e.target.valueAsNumber) ? 100 : e.target.valueAsNumber)}
> 
> ```

done

> ### `satelliteConfig` not null-safe
> 
> In `useSatellites.js`, `elevation >= satelliteConfig.minElev` will throw if `config.satellite` is undefined (user hasn't opened settings yet). Needs a guard:
> 
> 
> 
> ```js
> 
> const isVisible = elevation >= (satelliteConfig?.minElev || 5.0);
> 
> ```

done

> ## Should fix
> 
> ### Lat/lon default changed from `??` to `||`
> 
> The `SettingsPanel` lines for lat/lon were changed from `config?.location?.lat ?? 0` to `|| 0`. Since `0` is a valid coordinate, `??` was the correct operator — `||` treats `0` as falsy. It happens to work here because the fallback is also `0`, but it's semantically wrong and could mask bugs if the default ever changes. Please revert to `??`.

Is that also true for lat / lon / sationAlt / minElev since all have valid values of zero? updated four lines

> ### Hardcoded theme-unaware colors
> 
> Replacing `#hex` with `rgba()` equivalents doesn't solve the underlying theme issue — `rgba(48, 33, 21, .8)` and `rgba(0, 248, 0, .5)` are still dark-theme-only colors. I know inline HTML templates make CSS variables harder to use, but these section backgrounds will look off in Light and Retro. Could you at least use very low-opacity neutrals (e.g. `rgba(255,255,255,0.05)` / `rgba(0,0,0,0.1)`) that work across themes, or pass theme-aware colors from the component?

If it's okay I'd like to split it into a separate task, I'd like to get the functionality checked-in and merged and then come back and address this. Please assign me to
https://github.com/accius/openhamclock/issues/889

mrw.

accius (commented):
## Review — Satellite styling + visibility

**Nice additions:** station altitude + minimum-elevation settings wired through to calculations are a good improvement and address #855.

**Blockers / concerns:**

1. **`??` → `||` regression** (`SettingsPanel.jsx` ~line 182-185):
   ```js
   setLat(config.location?.lat || 0);
   setLon(config.location?.lon || 0);
   ```
   This treats a legitimate `lat=0` (equator) or `lon=0` (prime meridian) as falsy. The previous `??` was correct — please revert. Same applies to `parseFloat(lat) || 0` on save.

2. **Hardcoded colors** in `useSatelliteLayer.js`: the PR checklist explicitly requires CSS variables. Converting `#00ffff` → `rgba(0,255,255,1)` is still a hardcoded color, and the new row background colors (`rgba(48,33,21,.8)`, `rgba(0,248,0,.5)`, `rgba(37,46,23,.8)`, `rgba(35,59,70,.8)`) will not respect Dark/Light/Retro themes. Please move these to CSS variables (or at minimum theme-aware classes) so the popup works across themes.

3. **Mixed concerns:** this PR bundles a functional feature (altitude/minElev) with a significant cosmetic rewrite of the satellite popup HTML. Splitting into two PRs would make review much easier and keep bisects clean.

4. **Inconsistent config nesting:**
   - `useSatellites.js` receives `satelliteConfig` as a separate prop (`satelliteConfig?.minElev`)
   - `useSatelliteLayer.js` reads `config?.satellite?.minElev` off a combined config
   
   Please verify both call sites actually resolve — otherwise `minElev` silently defaults to 5.0 in one of the two paths.

5. **Altitude default** changed from 260m → 100m in `useSatelliteLayer.js`. Intentional? Worth a note if so.

6. **No input bounds** on `stationAlt` — accepts negative or absurdly large values.

7. **Unused `tle1`/`tle2`** added to `positions` — the description notes these are for a future iteration. Recommend holding them until the iteration that uses them, to keep this PR tight.

**Testing:** checklist has theme testing unchecked — given the scope of color changes, please verify in Dark / Light / Retro before merge.

Overall direction is good — please address 1 (functional bug) and 2 (theme compliance) at minimum.

MichaelWheeley:
> 
> 1. **`??` → `||` regression** (`SettingsPanel.jsx` ~line 182-185):
> 
>    ```js
>    setLat(config.location?.lat || 0);
>    setLon(config.location?.lon || 0);
>    ```
> 
>    This treats a legitimate `lat=0` (equator) or `lon=0` (prime meridian) as falsy. The previous `??` was correct — please revert. Same applies to `parseFloat(lat) || 0` on save.

Done, assume the case for lat / lon / stationAlt / minElev since all have valid values of zero.

> 2. **Hardcoded colors** in `useSatelliteLayer.js`: the PR checklist explicitly requires CSS variables. Converting `#00ffff` → `rgba(0,255,255,1)` is still a hardcoded color, and the new row background colors (`rgba(48,33,21,.8)`, `rgba(0,248,0,.5)`, `rgba(37,46,23,.8)`, `rgba(35,59,70,.8)`) will not respect Dark/Light/Retro themes. Please move these to CSS variables (or at minimum theme-aware classes) so the popup works across themes.

See https://github.com/accius/openhamclock/issues/889. I was trying to keep those changes out of this PR for clarity.... anyhow I ported them in manually with BeyondCompare.
 
> 3. **Mixed concerns:** this PR bundles a functional feature (altitude/minElev) with a significant cosmetic rewrite of the satellite popup HTML. Splitting into two PRs would make review much easier and keep bisects clean.

There's more changes coming that I've already deliberately kept out of here.

> 
> 4. **Inconsistent config nesting:**
> 
>    - `useSatellites.js` receives `satelliteConfig` as a separate prop (`satelliteConfig?.minElev`)
> 
>    - `useSatelliteLayer.js` reads `config?.satellite?.minElev` off a combined config
> 
>    
> 
>    Please verify both call sites actually resolve — otherwise `minElev` silently defaults to 5.0 in one of the two paths.

Yes, both paths resolve correctly, did some temporary numerical checking with throws. I didn't want to modify the calling function for the plugins which is why satelliteConfig is derived from config rather than passed.

> 
> 5. **Altitude default** changed from 260m → 100m in `useSatelliteLayer.js`. Intentional? Worth a note if so.

Already has a note `// above sea level [km], stationAlt is [m], defaults to 100m`,
but removing a legacy hard-coded default location of Horicon WI at 280m ASL possibly doesn't need to be explained in the code comments!

> 
> 
> 6. **No input bounds** on `stationAlt` — accepts negative or absurdly large values.

Limited [-500, 9000]m.
> 
> 7. **Unused `tle1`/`tle2`** added to `positions` — the description notes these are for a future iteration. Recommend holding them until the iteration that uses them, to keep this PR tight.

removed, to be added back in on the next PR.
> 
> **Testing:** checklist has theme testing unchecked — given the scope of color changes, please verify in Dark / Light / Retro before merge.

Done.

accius (changes requested):
## Re-review — Satellite styling + visibility (updated)

Good progress — input validation, theme variables, and table structure are all improved. One prior blocker is partially regressed and one new concern turned up.

### Fixed since prior review

- **Item 1 (input side):** `config.location?.lat ?? 0`, `config.location?.stationAlt ?? 100`, `config.satellite?.minElev ?? 5.0` all now use `??`. `setMinElev(isNaN(e.target.valueAsNumber) ? 5.0 : e.target.valueAsNumber)` correctly preserves `0` on input. Save path mirrors this with `isNaN(parseFloat(minElev)) ? 5.0 : parseFloat(minElev)`.
- **Item 2 (hardcoded colors) — popup HTML:** sat-card text/backgrounds now use `var(--bg-tertiary)` / `var(--text-secondary)` / `var(--accent-red)` / `var(--text-primary)` / `var(--accent-blue)` / `var(--border-color)`. The `#000` on `var(--accent-green)` is reasonable (black on green is readable in all themes). Commit "remove # hardcoded colors, replace with rgba()" is partially done — see issue 3 below for the remaining rgba uses.
- **Item 5 (altitude default):** Unified to 100 m per the "consistency" commit.
- **Item 6 (input bounds):** `min="-500" max="9000"` on `stationAlt`, `min="-5.0" max="89.0"` on `minElev`. Good.
- **"Zero minElev reverts to 5" bug (commit Apr 16):** Input-side handling now preserves `0`. See issue 1 for the consumer side.

### Still blocking

**1. `|| 5.0` / `|| 100` on the consumer side undoes the input fix**

The Apr 16 commit fixed the `0 → 5.0` reset in the input handler, but the downstream consumers still use `||`:

- `src/hooks/useSatellites.js:85` — `const isVisible = elevation >= (satelliteConfig?.minElev || 5.0);`
- `src/plugins/layers/useSatelliteLayer.js:82` — `isVisible = el >= (config?.satellite?.minElev || 5.0);`
- `src/hooks/useSatellites.js:54` — `height: (observerLocation.stationAlt || 100) / 1000`
- `src/plugins/layers/useSatelliteLayer.js:58` — `height: (config?.stationAlt || 100) / 1000`

Result: a user who sets **minimum elevation = 0°** (a valid choice: "show me anything above the horizon") still gets the hardcoded 5° cutoff applied. Same for **station altitude = 0 m** (sea-level station) falling back to 100 m. The setting is silently ignored.

Fix: change all four sites from `||` to `??`. The preserved-`0` story has to be end-to-end.

**2. Layer-popup markers still use hardcoded colors**

The commit "remove # hardcoded colors, replace with rgba()" converted hex to equivalent rgba — but rgba with constant values is still hardcoded. A handful remain in `useSatelliteLayer.js`:

- `border-bottom: 1px solid rgba(0, 68, 68, 1)` (that's `#004444` — should be `var(--border-color)`)
- Leaflet layer colors (`rgba(0, 255, 255, 1)` for tracks, `rgba(255, 255, 0, 1)` for lead track, `rgba(0, 255, 0, 1)` / `rgba(0, 255, 255, 1)` for footprint) — these are map-domain data viz and reasonable to keep hardcoded, but for theme consistency consider reading from `getComputedStyle(document.documentElement).getPropertyValue('--accent-cyan')` once at render time
- Selected-marker shadow and label color inside `divIcon.html`: `drop-shadow(0 0 10px rgba(0, 255, 255, 1))`, `color: rgba(255, 255, 255, 1)` — inside injected HTML, will not respect theme

The sat-card `border-bottom` is the only hard requirement to fix — it's a non-map chrome element. The Leaflet layer colors are judgment-call.

### Concerns carried forward

**3. Config shape inconsistency still unresolved.** `useSatellites.js` takes `satelliteConfig` as its second parameter, while `useSatelliteLayer.js` reads `config?.satellite?.minElev` off the layer config. Please verify that the layer actually receives a `config.satellite.minElev` at runtime — if it's always `undefined`, the `|| 5.0` fallback masks the bug.

**4. Mixed concerns.** This still bundles the altitude/minElev feature with the popup cosmetic overhaul. Not a blocker at this stage, but future PRs of this size would benefit from splitting.

### Verdict

Please change the four `|| 5.0` / `|| 100` sites to `??`, and switch the sat-card `border-bottom` to `var(--border-color)`. Those are the two remaining blockers from the "preserve zero values" and "theme compliance" items. Happy to re-approve once those land.

MichaelWheeley:
### @accius, on 4/19 you said,

> ### Still blocking
> 
> 
> 
> **1. `|| 5.0` / `|| 100` on the consumer side undoes the input fix**
> 
> 
> 
> The Apr 16 commit fixed the `0 → 5.0` reset in the input handler, but the downstream consumers still use `||`:
> 
> 
> 
> - `src/hooks/useSatellites.js:85` — `const isVisible = elevation >= (satelliteConfig?.minElev || 5.0);`
> 
> - `src/plugins/layers/useSatelliteLayer.js:82` — `isVisible = el >= (config?.satellite?.minElev || 5.0);`
> 
> - `src/hooks/useSatellites.js:54` — `height: (observerLocation.stationAlt || 100) / 1000`
> 
> - `src/plugins/layers/useSatelliteLayer.js:58` — `height: (config?.stationAlt || 100) / 1000`
> 
> 
> 
> Result: a user who sets **minimum elevation = 0°** (a valid choice: "show me anything above the horizon") still gets the hardcoded 5° cutoff applied. Same for **station altitude = 0 m** (sea-level station) falling back to 100 m. The setting is silently ignored.
> 
> 
> 
> Fix: change all four sites from `||` to `??`. The preserved-`0` story has to be end-to-end.
> 
*Done.*
> 
> **2. Layer-popup markers still use hardcoded colors**
> 
> 
> 
> The commit "remove # hardcoded colors, replace with rgba()" converted hex to equivalent rgba — but rgba with constant values is still hardcoded. A handful remain in `useSatelliteLayer.js`:
> 
> 
> 
> - `border-bottom: 1px solid rgba(0, 68, 68, 1)` (that's `#004444` — should be `var(--border-color)`)

*Done.*
> 
> - Leaflet layer colors (`rgba(0, 255, 255, 1)` for tracks, `rgba(255, 255, 0, 1)` for lead track, `rgba(0, 255, 0, 1)` / `rgba(0, 255, 255, 1)` for footprint) — these are map-domain data viz and reasonable to keep hardcoded, but for theme consistency consider reading from `getComputedStyle(document.documentElement).getPropertyValue('--accent-cyan')` once at render time

*`--accent-cyan --accent-green --accent-amber`*

> 
> - Selected-marker shadow and label color inside `divIcon.html`: `drop-shadow(0 0 10px rgba(0, 255, 255, 1))`, `color: rgba(255, 255, 255, 1)` — inside injected HTML, will not respect theme

*Not modified.*
> 
> The sat-card `border-bottom` is the only hard requirement to fix — it's a non-map chrome element. The Leaflet layer colors are judgment-call.
> 
> 
> ### Concerns carried forward
> 
> 
> 
> **3. Config shape inconsistency still unresolved.** `useSatellites.js` takes `satelliteConfig` as its second parameter, while `useSatelliteLayer.js` reads `config?.satellite?.minElev` off the layer config. Please verify that the layer actually receives a `config.satellite.minElev` at runtime — if it's always `undefined`, the `|| 5.0` fallback masks the bug.
> 
*Previously verified.*
> 
> **4. Mixed concerns.** This still bundles the altitude/minElev feature with the popup cosmetic overhaul. Not a blocker at this stage, but future PRs of this size would benefit from splitting.
> 
*Heard.*

> 
> ### Verdict
> 
> 
> 
> Please change the four `|| 5.0` / `|| 100` sites to `??`, and switch the sat-card `border-bottom` to `var(--border-color)`. Those are the two remaining blockers from the "preserve zero values" and "theme compliance" items. Happy to re-approve once those land.

accius (changes requested):
## Re-review — Satellite styling + visibility (round 3)

Nearly there. The four call-sites flagged last round are all converted to `??`; the sat-card `border-bottom` now uses `var(--border-color)`; Leaflet track colors are read from CSS vars. One preserve-zero residual remains in the SettingsPanel Save path, and the `accentCyan` undo in `useSatellites.js` has a defensible rationale but leaves theme-compliance slightly inconsistent.

### Fixed since prior review

- **Item 1 (all four `|| 5.0` / `|| 100` sites) — FIXED.** Commit `b562a47` converts every call-site flagged last round:
  - `src/hooks/useSatellites.js:54` — `height: (observerLocation.stationAlt ?? 100) / 1000`
  - `src/hooks/useSatellites.js:85` — `const isVisible = elevation >= (satelliteConfig?.minElev ?? 5.0);`
  - `src/plugins/layers/useSatelliteLayer.js:58` — `height: (config?.stationAlt ?? 100) / 1000`
  - `src/plugins/layers/useSatelliteLayer.js:82` — `isVisible = el >= (config?.satellite?.minElev ?? 5.0);`
  Zero-preservation is now end-to-end for these paths.

- **Item 2 (sat-card `border-bottom`) — FIXED.** Commit `1976a24` at `src/plugins/layers/useSatelliteLayer.js:261`:
  `<div class="sat-card" style="border-bottom: 1px solid var(--border-color); ...">`

- **Leaflet layer colors — converted to CSS-var lookups.** Commit `bc6cbca` replaces hardcoded rgba with runtime lookups at `src/plugins/layers/useSatelliteLayer.js:387-389`:
  ```js
  const accentCyan = getComputedStyle(document.documentElement).getPropertyValue('--accent-cyan');
  const accentGreen = getComputedStyle(document.documentElement).getPropertyValue('--accent-green');
  const accentLeadTrack = getComputedStyle(document.documentElement).getPropertyValue('--accent-amber');
  ```
  and uses them for footprint fill, track polylines, and lead-track dashArray. The data-window panel `accentCyan` at line 124 is also from the lookup. All four referenced CSS variables (`--accent-cyan`, `--accent-green`, `--accent-amber`, `--border-color`) are defined in every theme block in `src/styles/themes.css` (dark / light / legacy / retro).

### Still open

**1. `parseInt(stationAlt) || 100` in the Save path — sea-level station still saves as 100 m.**

At `src/components/SettingsPanel.jsx:426`:
```js
location: { lat: parseFloat(lat) || 0, lon: parseFloat(lon) || 0, stationAlt: parseInt(stationAlt) || 100 },
satellite: { minElev: isNaN(parseFloat(minElev)) ? 5.0 : parseFloat(minElev) },
```

The `minElev` branch on line 427 uses the correct `isNaN(…) ? default : value` guard, but `stationAlt` on line 426 still uses `|| 100`. A user who legitimately sets stationAlt to `0` (sea-level station) will see the save silently rewrite it to 100. The input handler at line 3392 already handles this correctly (`setStationAlt(isNaN(e.target.valueAsNumber) ? 100 : e.target.valueAsNumber)`) — only the Save path needs the same pattern, e.g. `stationAlt: isNaN(parseInt(stationAlt)) ? 100 : parseInt(stationAlt)`.

The `lat` / `lon` `|| 0` fallbacks happen to be no-ops (`parseFloat("0") || 0 → 0`) so those are fine as written.

### Undo decision — reasonable

Commit `1fcc3a4` reverts the `useSatellites.js:145` color lookup from `tle.color || accentCyan` back to `tle.color || '#00ffff'`. Rationale given: "it makes no sense on a safety default usage." This is defensible: `getComputedStyle(document.documentElement).getPropertyValue('--accent-cyan')` can return `''` if the call lands before the stylesheet is applied (data hydration path runs earlier than map-render effects), and `tle.color || '' → ''` would yield an invisible polyline. The hardcoded `#00ffff` guarantees a visible default.

The same risk theoretically applies to the three `getPropertyValue` reads in `useSatelliteLayer.js:387-389`, but those run inside `renderSatellites`, which is gated on `map` being mounted — by then the theme CSS has been applied. Accept the asymmetry as intentional.

Soft note: the hardcoded `#00ffff` matches `--accent-cyan` only in dark + retro; in light mode `--accent-cyan: #0099bb`. So light-mode users get a slightly jarring pure-cyan default when `tle.color` is missing. Not a blocker — per-satellite color is configurable in `satellites.json`.

### New concerns

None that block. Two soft observations:

- `getPropertyValue(...)` returns the value with a leading space in most browsers. `'1px solid ' + accentCyan` (line 136) produces `'1px solid  #00ddff'` which CSS tolerates, and Leaflet's SVG stroke also tolerates it, so no visible bug — but a `.trim()` on the three reads at lines 124, 387-389 would be a small hardening.

- `useSatelliteLayer.js:465-466` still injects hardcoded white/cyan into Leaflet `divIcon` HTML (`drop-shadow(0 0 10px rgba(0, 255, 255, 1))`, `color: rgba(255, 255, 255, 1)`). These are inside a string passed to `divIcon({ html })` so CSS vars won't resolve. Carry-over from last review — still judgment-call, not a blocker.

### Verdict

CHANGES_REQUESTED — one-line fix: replace `parseInt(stationAlt) || 100` at `src/components/SettingsPanel.jsx:426` with the same `isNaN(parseInt(stationAlt)) ? 100 : parseInt(stationAlt)` guard used for `minElev` on the next line. After that, ready to approve.

accius (approved):
## Re-review — round 4

The one blocker from the last pass is fixed in `8ed7b0b` — `src/components/SettingsPanel.jsx:423-427` now matches the `minElev` pattern on the next line:

```js
stationAlt: isNaN(parseInt(stationAlt)) ? 100 : parseInt(stationAlt),
```

Sea-level (stationAlt=0) is preserved end-to-end now. Nothing else changed in this commit. All prior review items are either fixed (`|| → ??` across all four consumer sites, sat-card `border-bottom` to `var(--border-color)`, Leaflet track colors via CSS-var lookups) or explicitly accepted (the `'#00ffff'` fallback in `useSatellites.js` is the right call for early-hydration safety).

### Verdict

APPROVE — ready to merge.

PR #879, opened 2026-04-05
[Feature/sat] Satellite prediction

MichaelWheeley:
## What does this PR do?

- Adds satellite prediction modal.
- Fixes problem with inability to scroll satellite details window.
- SatelliteOrbit.js is imported with very minor modification from satvis project, MIT license.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. select many satellites in the details window such that it overflows the window - it should now be possible to mouse scroll up and down.
2. From the satellite details window choose the name of any listed satellite, it will open a modal dialog showing satellite prediction for the next 7 days. Press close or escape key to exit the modal.
3. Change the minimum elevation [deg] in the settings -> satellite menu, e.g. set it to some quite high like 70 degrees, then go back and open a new modal. Only those passes where the minimum is reached are shown, and the duration of the passes (that is the duration above the minimum threshold) will be shorter.
4. Leave the modal open, the times in the time from now column will decrement each minute.

## Checklist

- [X] If adding a panel: wired into Modern, Classic, and Dockable layouts

each satellite listed is now a PREDICT button - press it to get a prediction modal
image

image


accius (changes requested):
Good progress on satellite prediction — the pass table and auto-refresh are nice additions. A few things to address:

### Must fix

**XSS risk in onclick handler** — Satellite names and TLE lines are injected unescaped into an inline `onclick`:
```js
onclick="openSatellitePredict('${sat.name}', '${sat.tle1}', '${sat.tle2}')"
```
A satellite name containing `'` breaks the handler; malicious content is injectable. Use `data-` attributes with a delegated event listener, or escape the values.

**Interval leak** — `window.satellitePredictInterval` is overwritten if a second modal opens before the first interval fires. The old interval keeps running. Clear the previous one before assigning:
```js
if (window.satellitePredictInterval) clearInterval(window.satellitePredictInterval);
```

**Carries all #877 issues** — This PR includes #877's commits, so the hardcoded colors, broken nested `<table>` HTML, duplicate `style` attribute, and `valueAsNumber ?? fallback` bugs all apply here too. Fix those in #877 first.

### Should fix

**Commented-out code** — Large blocks in `useSatelliteLayer.js` (~30 lines) are commented out rather than removed. Either delete them now or open an issue to track it — don't ship dead code in comments.

**Hardcoded button colors** — `background: #440000; border: 1px solid #ff4444` on the satellite name button won't work in Light/Retro themes. Use CSS variables.

### Worth discussing

**`dayjs` dependency** — Does the project already have date utilities that could handle the `format()` and `diff()` calls? Adding a new dependency for a few date operations adds bundle weight. Not a blocker if the team is fine with it.

**60s full recomputation** — The interval recomputes all orbital passes every minute. Since pass times only shift meaningfully over hours, you could just update the "from now" column without rerunning the orbital math.

accius (commented):
## Review — Satellite prediction modal (draft)

Nice feature. A few things to address before exiting draft:

**Correctness / safety**

1. **String interpolation into `onclick`** is fragile and XSS-adjacent:
   ```js
   onclick="openSatellitePredict('${sat.name}', '${sat.tle1}', '${sat.tle2}')"
   ```
   A satellite name containing a single quote (or a TLE line containing backslash/newlines) will break the handler or allow HTML injection. TLE source is currently trusted, but please switch to a data-attribute + `addEventListener` pattern, or register the sat by an id lookup in a module-level map and pass only an index.

2. **Invalid HTML — nested `<table>` inside `<table>`**. The sat-card already wraps everything in one `<table>`, and the PR opens three more nested `<table>` elements inside its body, plus a stray trailing `</table>`. Browsers recover, but layout/paint is unpredictable and it will bite future styling changes. Use `<tbody>` sections with classes instead.

3. **setInterval memory leak on re-open:** `window.satellitePredictInterval` is overwritten without clearing the previous interval when the user clicks another satellite name while a modal is already open. Clear it at the top of `openSatellitePredict` unconditionally.

4. **keydown listener leak:** `handleKeyDown` only removes itself on Enter/Escape. Backdrop-click close and the "replace existing modal" path leave the listener attached, and each re-open adds another. Always remove in the cleanup paths.

5. **`window.openSatellitePredict` never cleaned up** in the useEffect return. On hot reload it's fine, but consider returning a cleanup that deletes the global.

6. **`generateModalContent(currentPasses)` uses the recomputed `currentPasses`**, but the initial render uses `passes` from an earlier `computePassesElevation` call. These two calls happen back-to-back with no TLE change — recompute once and use the single array.

**Code quality**

7. **Inconsistent config access:** the component already receives `config` via `useLayer` props but the PR adds `useAppConfig()` to grab `globalConfig`. Pick one path — mixing them means settings changes won't propagate consistently.

8. **Hardcoded colors** (`#302115`, `#233b46`, `#00f800`, `#252e17`) in the satellite card HTML break theme support. Please use CSS variables per the checklist.

9. **Orbit.js modification from upstream** — changing `0.5 → 0.05` in `orbitalPeriod` skip factor increases compute cost ~10x. Please add a short comment explaining the observed failure mode that motivated it, and consider whether we can keep the upstream default in the non-first-pass path.

10. **`computePassesSwath` imported from upstream but appears unused** in this PR. Drop dead code until needed.

11. **package.json** adds `dayjs`. The repo already uses plain `Date`/`Intl` in many places; adding a date lib just for formatting and `.add(7, 'day')` is heavyweight. Consider `Intl.DateTimeFormat` + `date.getTime() + 7*86400000`.

12. **`lastElevation`** is read but reassignment after loop entry is unused in the branch where `ongoingPass` is true (just noting for future cleanup).

**Dependency**

- Stacked on #877. Please mark this ready for review only after #877 lands (or rebase so it's self-contained).

**UX**

- Modal `width: 50vw` will be unusable on mobile — suggest `min-width` + `max-width` responsive handling. Checklist has "Responsive at different screen sizes" unchecked.
- Closing on Enter is surprising; Escape is standard. Suggest dropping Enter.

Direction is great — mostly safety and cleanup needed.

MichaelWheeley:
### still waiting on #877 before this one is ready for review

### @accius in https://github.com/accius/openhamclock/pull/879 you said April 5,

> ### Must fix
> 
> **XSS risk in onclick handler** — Satellite names and TLE lines are injected unescaped into an inline `onclick`:
> 
> ```js
> 
> onclick="openSatellitePredict('${sat.name}', '${sat.tle1}', '${sat.tle2}')"
> 
> ```
> 
> A satellite name containing `'` breaks the handler; malicious content is injectable. Use `data-` attributes with a delegated event listener, or escape the values.

*now handled in event listener*

> **Interval leak** — `window.satellitePredictInterval` is overwritten if a second modal opens before the first interval fires. The old interval keeps running. Clear the previous one before assigning:
> 
> ```js
> 
> if (window.satellitePredictInterval) clearInterval(window.satellitePredictInterval);
> 
> ```

*done*

> **Carries all #877 issues** — This PR includes #877's commits, so the hardcoded colors, broken nested `<table>` HTML, duplicate `style` attribute, and `valueAsNumber ?? fallback` bugs all apply here too. Fix those in #877 first.

*changes from previous PR merged*

> ### Should fix
> 
> **Commented-out code** — Large blocks in `useSatelliteLayer.js` (~30 lines) are commented out rather than removed. Either delete them now or open an issue to track it — don't ship dead code in comments.

*removed dead code in useSatelliteLayer.js, this appeared to be legacy code, unused because satellite location calculations are now done as a hook in useSatellites.js*

> **Hardcoded button colors** — `background: #440000; border: 1px solid #ff4444` on the satellite name button won't work in Light/Retro themes. Use CSS variables.

*Hard-coded colors where removed in earlier PR and merged into this one.*

> ### Worth discussing
> 
> **`dayjs` dependency** — Does the project already have date utilities that could handle the `format()` and `diff()` calls? Adding a new dependency for a few date operations adds bundle weight. Not a blocker if the team is fine with it.

*use of `dayjs` completely eliminated, package removed from `package.json`*

> **60s full recomputation** — The interval recomputes all orbital passes every minute. Since pass times only shift meaningfully over hours, you could just update the "from now" column without rerunning the orbital math.

*Modified, the intensive satellite calculation now only occurs once as the modal opens. Refresh time is every second and the 'Time From Now' is displayed as +hours:minutes:seconds. Displayed time is replaced with word 'VISIBLE' for a visible satellite, or the whole line is dropped from the table if the satellite has already passed.*



### @accius You also said April 14,

> 1. **String interpolation into `onclick`** is fragile and XSS-adjacent:....

*duplicate*

> 2. **Invalid HTML — nested `<table>` inside `<table>`**.....

*duplicate*

> 3. **setInterval memory leak on re-open:** `window.satellitePredictInterval` is overwritten without.....

*duplicate*

> 4. **keydown listener leak:** `handleKeyDown` only removes itself on Enter/Escape. Backdrop-click close and the "replace existing modal" path leave the listener attached, and each re-open adds another. Always remove in the cleanup paths.

*event listeners cleaned-up on exit from main window and from modal.*

> 5. **`window.openSatellitePredict` never cleaned up** in the useEffect return. On hot reload it's fine, but consider returning a cleanup that deletes the global.

*clean-up added to `window.openSatellitePredict`*

> 6. **`generateModalContent(currentPasses)` uses the recomputed `currentPasses`**, but the initial render uses `passes` from an earlier `computePassesElevation` call. These two calls happen back-to-back with no TLE change — recompute once and use the single array.

*duplicate*

> 7. **Inconsistent config access:** the component already receives `config` via `useLayer` props but the PR adds `useAppConfig()` to grab `globalConfig`. Pick one path — mixing them means settings changes won't propagate consistently.

*removed reference to `useAppConfig`*

> 8. **Hardcoded colors** (`#302115`, `#233b46`, `#00f800`, `#252e17`) in the satellite card HTML break theme support. Please use CSS variables per the checklist.

*duplicate*

> 9. **Orbit.js modification from upstream** — changing `0.5 → 0.05` in `orbitalPeriod` skip factor increases compute cost ~10x. Please add a short comment explaining the observed failure mode that motivated it, and consider whether we can keep the upstream default in the non-first-pass path.

*There was a bug in the imported algorithm that masked passes within 0.5 orbital periods of the current time. The bug has been fixed.*

> 10. **`computePassesSwath` imported from upstream but appears unused** in this PR. Drop dead code until needed.

*removed unused `computePassesSwath()` from imported code `orbit.js`*

> 11. **package.json** adds `dayjs`.... 

*duplicate*

> 12. **`lastElevation`** is read but reassignment after loop entry is unused in the branch where `ongoingPass` is true (just noting for future cleanup).

*has been cleaned up*

> - Modal `width: 50vw` will be unusable on mobile — suggest `min-width` + `max-width` responsive handling. Checklist has "Responsive at different screen sizes" unchecked.

*modal min/max height and width set as follows,*
```js
        min-width: 50vw;
        max-width: 95vw;
        min-height: 25vh;
        max-height: 90vh;
```

>
> - Closing on Enter is surprising; Escape is standard. Suggest dropping Enter.
>

*Keyboard 'Enter' key removed as option to exit modal.*

accius (commented):
## Re-review — Satellite prediction modal (still draft)

Substantial progress — all of the blocking bugs from the prior review are addressed. This is close to ready for un-draft.

### Fixed from prior review

- **Item 1 (XSS via `onclick` interpolation):** Replaced with `data-action="open-predict"` + `data-sat-name` / `data-tle1` / `data-tle2` attributes, a single delegated click listener on the map container, and `window.openSatellitePredict(name, tle1, tle2)` called with the unescaped values read via `dataset`. Safe from HTML injection. Same pattern applied to `clear-all-satellites`, `toggle-satellite`, and `close-predict-modal`.

- **Item 2 (nested `<table>` elements):** The sat-card now uses a single flat `<table>` with HTML comments marking sections ("section 1: position/motion", "section 2: visibility", "section 3: misc"). The PREDICT button is a `<tr>` inside that same table. No nesting, no stray closing tag. The modal's pass table is a separate standalone `<table>` — correctly scoped.

- **Item 3 (interval leak):** `if (window.satellitePredictInterval) clearInterval(...)` guards re-assignment. `closeModal` also clears it.

- **Item 4 (keydown listener leak):** `closeModal` removes `handleKeyDown` (Escape), `handleContentClick` (close-modal delegation), and `handleModalClick` (backdrop). Clean.

- **Item 5 (`window.openSatellitePredict` cleanup):** `useEffect` returns `delete window.openSatellitePredict`.

- **Item 7 (mixed `useAppConfig` / props `config`):** `useAppConfig` reference removed; component reads `config` from props consistently.

- **Item 8 (hardcoded colors):** All replaced with CSS variables except `color: '#000'` for text on `var(--accent-green)` background — that's a reasonable keep for contrast.

- **Item 9 (orbit.js `0.5 → 0.05`):** Rather than patching the upstream file, a fresh `src/utils/orbit.js` was forked; the skip factor is back to `0.5` with an `enableLogging` diagnostic switch when future tuning is needed.

- **Item 10 (`computePassesSwath` unused):** removed.

- **Item 11 (`dayjs` dependency):** eliminated. Time formatting now uses native `Date` + `padStart`.

- **UX — Enter / Escape:** Enter no longer closes (good — surprising behaviour); Escape still works.

- **Window drag-handle listener leak:** `winListenersRef` stores the handlers, both `win.remove()` branches clean them up before removing. Good.

### Remaining concerns

**1. Attribute interpolation escaping (defense-in-depth).**
```js
data-sat-name="${sat.name}" data-tle1="${sat.tle1}" data-tle2="${sat.tle2}"
```
Satellite names today are structured (`ISS (ZARYA)`, `AO-7`), so this works in practice. But a satellite name containing `"` would break the attribute. A tiny helper:
```js
const attrEscape = (s) => String(s ?? '').replace(/&/g, '&amp;').replace(/"/g, '&quot;');
```
…applied to the three data attributes plus `${sat.notes}` / `${sat.mode}` / `${sat.downlink}` / `${sat.uplink}` / `${sat.tone}` (which are inserted as text content) would harden this without changing behavior for valid data.

**2. `updatePasses` rebuilds `content.innerHTML` every 1 s.** The "from now" column needs periodic refresh, but recomputing all rows is wasteful and also means any user selection/scroll state in the table is lost each second. Consider targeted updates: render once, give each "from now" cell a stable `id` or `data-pass-id`, and only rewrite those cells from the interval. Not a blocker, more like an optimization.

**3. `enableLogging` flag in `orbit.js`.** If this is intended as a dev/diagnostic switch, wire it through a URL query param or `localStorage` flag so it can be toggled without editing source. If it's internal-only, consider removing the partially-built log messages (line `logMessage += …` but no `console.log` sink) before merge.

**4. Default station altitude `100` m.** Seems high as a world-wide default (mean sea level would be more conservative). But the range [−500, 9000] m is sensible; minor cosmetic choice.

**5. Responsive check unchecked.** The modal's `min/max width` are set, but please spot-check on a ≤ 375 px viewport before un-draft — the pass table has 9 columns and can get cramped.

### Verdict
No outstanding blockers. Address #1 (attribute escape is a 10-liner) and spot-check responsive before flipping out of draft. Approving on that.

accius (commented):
## Re-review — Satellite prediction modal (round 3, still draft)

Small, targeted cleanup round. The diff since the 2026-04-20 review is only 104 lines across two files (`src/hooks/useSatellites.js`, `src/plugins/layers/useSatelliteLayer.js`) — mostly theme-variable migration and one `||` → `??` hygiene pass. Nothing regressed; prior open items (attribute escaping, 1-Hz `innerHTML` rebuild, `enableLogging` wiring, station-altitude default, small-viewport spot-check) are unchanged.

### Fixed since prior review

- **`||` → `??` on safety defaults** (`b562a47`): `useSatellites.js:54` and `:85`, plus `useSatelliteLayer.js:58` and `:82`. Now `stationAlt = 0` and `minElev = 0` won't silently fall through to the hardcoded default. This wasn't an explicit prior concern, but it's a good defensive fix — the earlier `d71a5303` commit ("fix bug where zero minimum elevation is reverting to 5") was patching a symptom; `??` fixes the root cause.

- **Stray hardcoded border color**: `useSatelliteLayer.js:270` (`border-bottom: 1px solid rgba(0, 68, 68, 1)` on `.sat-card`) is now `var(--border-color)`. That was a leftover.

- **Sat-data panel and footprint/track colors promoted to theme vars** (`bc6cbca`): `useSatelliteLayer.js:98` introduces `const accentCyan = getComputedStyle(document.documentElement).getPropertyValue('--accent-cyan')` at the top of `updateInfoWindow`, and `:387–389` declares `accentCyan` / `accentGreen` / `accentLeadTrack` (=`--accent-amber`) at the top of `renderSatellites`. The four call sites previously hardcoded as `rgba(0,255,255,1)` / `rgba(0,255,0,1)` / `rgba(255,255,0,1)` now read from those variables. All four CSS custom properties (`--accent-cyan`, `--accent-green`, `--accent-amber`, `--border-color`) are defined for every theme in `src/styles/themes.css` (lines 10/14/16/20, 35/39/41/45, 60/64/66/70, 85/91/93/97).

- **Undo of `accentCyan` in `useSatellites.js`** (`1fcc3a4`): correct call. The reverted line was `color: tle.color || accentCyan` in the per-satellite position record. That `color` field is a piece of **data** that flows downstream; baking a theme-dependent value into data at propagation time means the color would never update when the theme changes, and the per-satellite `tle.color` from `satellites.json` is the authoritative source anyway. The fallback `'#00ffff'` is a cosmetic safety net for TLEs missing the optional `color` field — a hardcoded hex is actually the right choice there. Author's commit message ("makes no sense on a safety default usage") matches my reading.

### Still open / regressed

Carried over from prior review, unchanged in this round:

1. **Attribute-value escaping on the PREDICT button** (`useSatelliteLayer.js:345–347`):
   ```js
   data-sat-name="${sat.name}"
   data-tle1="${sat.tle1}"
   data-tle2="${sat.tle2}"
   ```
   Still unescaped. Works for all current data; breaks if any TLE metadata ever contains `"`. Same for `${sat.notes}` (line 364). A 3-line `attrEscape` helper would close this.

2. **`updatePasses` rebuilds `content.innerHTML` at 1 Hz** (`useSatelliteLayer.js:757–759` + `:784`). Still wasteful and still wipes user scroll/selection in the pass table each second.

3. **`enableLogging` in `orbit.js`** is a hardcoded `const enableLogging = false` with a real `console.log(logMessage)` sink at line 216 (sink was added — previously missing). Still not wired to a URL query param / `localStorage`, so toggling requires editing source.

4. **Default station altitude `100 m`** at `useSatellites.js:54` and `useSatelliteLayer.js:58`. Cosmetic.

5. **Small-viewport responsive check**: still cannot verify from source; 9-column pass table on ≤375 px viewport should be spot-checked before un-draft.

### New concerns

1. **`getComputedStyle(...).getPropertyValue(...)` is not `.trim()`ed**. `useSatelliteLayer.js:98, 387, 388, 389` all read raw values:
   ```js
   const accentCyan = getComputedStyle(document.documentElement).getPropertyValue('--accent-cyan');
   ```
   The existing codebase pattern (`src/plugins/layers/useIBPLayer.js:181–182`, `src/theme/themeUtils.js:19`) always calls `.trim()`. Per CSSOM spec the returned value can carry leading whitespace from the source declaration (`--accent-cyan: #00ddff;`). In practice Leaflet/CSS parsers swallow it, but `border: '1px solid ' + accentCyan` would produce `'1px solid  #00ddff'` (double space) and is inconsistent with the established convention. Low-impact but worth a `.trim()` on all four reads for consistency.

2. **Redundant `getComputedStyle` call at `:98`**. It's declared before the `if (!win)` guard but only consumed inside that branch (lines 108, 110), so the lookup runs on every update (hot path, ~1 Hz) even when the window already exists. Move it inside `if (!win)` or leave it — either way, not a correctness issue.

3. **Style-aware color is captured once at element creation time**. The panel `win.style.color` / `win.style.border` set at `:108/:110` are baked into inline styles when the DOM element is first created. If the user switches themes while the sat data panel is open, the panel keeps its old-theme cyan. The footprint/track colors rebuild each `renderSatellites()` call so they pick up new theme values naturally. Minor. A simple alternative would be to set `color: 'var(--accent-cyan)'` directly (CSS handles the indirection live) — which is the pattern used elsewhere in the same file (`:133` `backgroundColor: 'var(--bg-primary)'`). The only reason to materialise the value through JS is if the consumer (Leaflet) doesn't understand `var(...)`; here the consumer is inline `style`, which does. Consider reverting the `updateInfoWindow` changes to use `'var(--accent-cyan)'` string literals directly, like the sibling `backgroundColor`.

4. **Naming — `accentLeadTrack` vs `--accent-amber`**. At `:389`:
   ```js
   const accentLeadTrack = getComputedStyle(...).getPropertyValue('--accent-amber');
   ```
   The other two locals (`accentCyan`, `accentGreen`) match their CSS-var roots; this one doesn't. Fine if the lead-track color is intended to be semantically independent of "amber" (so a future theme could change `--lead-track` without touching amber), but then it should read a `--lead-track` (or similar) variable, not `--accent-amber`. As written, it's a misleading alias.

### Verdict

COMMENT — no new blockers introduced by these six commits. The round does what it says on the tin: theme-variable cleanup and a defensive `||`→`??` pass. Prior open items (attribute escape, 1-Hz innerHTML rebuild, `enableLogging` wiring, responsive spot-check) are still worth closing before flipping out of draft; the new trim inconsistency (#1 above) is a 4-character fix. The `accentCyan` undo in `useSatellites.js` is correctly reasoned.

PR #891, opened 2026-04-07
fix(propagation): use usedITURHFProp guard to prevent sparse-array nu…

ceotjoe:
## What does this PR do?

The crash path was:

1. ITURHFProp full hourly fetch fails → single-hour fetch succeeds → `predictions[band][14] = {...}` (sparse array, `usedITURHFProp` stays `false`)
2. Fallback guard `predictions['160m'].length === 0` is `false` (sparse array length is 15) → fallback skipped
3. Server JSON-serializes sparse array → holes become `null` → client receives `[null, null, ..., {hour:14,...}]`
4. `.find(h => h.hour === hour)` calls callback with `h = null` → **crash**

The fix uses `!usedITURHFProp` which is already the correct semantic: "did we get complete 24-hour ITURHFProp data?" If not, run the fallback to fill all 24 slots (the pre-seeded single-hour entry at `existing[currentHour]` is preserved by the `if (existing[hour])` check on line 440).
## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

Before:
Bildschirmfoto 2026-04-07 um 08 42 38

After, to be tested.

### DIFF
--- a/server/routes/propagation.js
+++ b/server/routes/propagation.js
@@ -425,8 +425,11 @@ module.exports = function (app, ctx) {
       }
 
       // ===== FALLBACK: Built-in calculations =====
-      // Used when ITURHFProp is unavailable (self-hosted without the service)
-      if (!predictions['160m'] || predictions['160m'].length === 0) {
+      // Used when ITURHFProp is unavailable (self-hosted without the service).
+      // Also fills in the remaining 23 hours when only a single-hour ITURHFProp
+      // result was used — without this the sparse array holes become JSON nulls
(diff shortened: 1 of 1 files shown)


alanhargreaves (approved):
Jörg and I are in agreement on the cause of this. I was looking at a client fix, but this addresses the underlying issue.

ceotjoe:
Probably I can test this with the guy who reported it in the afternoon. He seems same timezone. :-)

PR #895, opened 2026-04-07
Don't call fetchFull() on a useVisibilityRefresh() in useWSJTX()

alanhargreaves:
## What does this PR do?

The aim of calling `fetchFull()` in `useVisibilityRefresh()` is to quickly schedule a full next pull from the API (which we are not using). It also clears the current entries as it believes it's going to pull a full history, This is not the case with SSE.

Indeed if you move off a tab or close the browser and then come back, you end up with an empty panel for WSJT-X.

This change ensures that if we are using SSE that we don't make that call.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Display the WSJT-X Panel (I find teh CQ Oly sessions more useful here as they refresh slower)
2. Navigate away from the OHC tab and then back
3. Note that we retain history

## Checklist

### DIFF
--- a/src/hooks/useWSJTX.js
+++ b/src/hooks/useWSJTX.js
@@ -182,8 +182,10 @@ export function useWSJTX(enabled = true) {
   }, [enabled, fetchFull, pollDecodes]);
 
   // Refresh immediately when tab becomes visible (handles browser throttling)
+  // Don't do this if we are using SSE as fetchFull() flushes our history and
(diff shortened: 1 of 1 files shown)


ceotjoe:
Looks good to me.

PR #900, opened 2026-04-09
Feature/wsjtx sse enrichment

ceotjoe:
## Summary

This PR moves the WSJTX decode enrichment intelligence from the OHC server into the local `rig-bridge` process, so SSE consumers receive fully enriched events even when no server relay is configured.

- **Local enrichment library** (`rig-bridge/lib/wsjtx-enrich.js`): Maidenhead grid → lat/lon, Hz → band name, FT8/FT4 message parser (CQ/QSO type, caller, modifier, embedded grid), per-client grid cache (2 h TTL), and full enrichment helpers for DECODE, STATUS, QSO_LOGGED, and WSPR_DECODE messages.
- **HamQTH callsign lookup** (Phase 5, opt-in): fire-and-forget background resolution of callsign → lat/lon via `hamqth.com/dxcc.php`; 24 h cache, max 5 concurrent requests. On resolution a `decode-update` SSE event is emitted so the frontend can live-patch map pins already on screen. Toggled via a new checkbox in the rig-bridge Integrations tab.
- **End-to-end wiring** (`rig-bridge.js`): decode, status, qso, clear, wspr, and decode-update bus events are now all forwarded to SSE with their full enriched field sets (previously only a trimmed subset was forwarded, silently dropping all enrichment).
- **Frontend** (`useWSJTX.js`): handles new SSE event types — `clear` (per-client filtering), `wspr` (capped ring buffer), `decode-update` (async HamQTH patch). Band-change detection uses the `bandChanged` flag from enriched status; falls back to manual comparison on the server/relay path.
- **Bug fix**: `wsjtx-enrich` was emitting lowercase `'cq'`/`'qso'` type strings; changed to uppercase `'CQ'`/`'QSO'` to match the server convention and fix CQ filter + row colouring in PSKReporterPanel.
- **Bug fix**: `ModernLayout` was not passing `wsjtxWspr` to `PSKReporterPanel` — the WSPR tab now receives data in that layout.
- Enrichment on the **relay path is unchanged** — the rig-bridge relay queue still sends raw messages so the server can do its own enrichment without double-processing.

## Test plan

- [ ] Run rig-bridge in **SSE-only mode** (no server relay): confirm DECODE events in PSKReporterPanel show band, grid, lat/lon, CQ/QSO type and row colour
- [ ] Enable **HamQTH callsign lookup** in the Integrations tab: confirm callsigns without an embedded grid eventually get lat/lon resolved and map pins appear
- [ ] Confirm **WSPR tab** populates in ModernLayout (was broken before this PR)
- [ ] Confirm **CLEAR** from WSJT-X wipes only the correct client's decodes
- [ ] Run rig-bridge with a **server relay configured**: confirm server-side enrichment still works and no fields are duplicated or corrupted
- [ ] Check rig-bridge status endpoint reports `gridCacheSize`, `callsignCacheSize`, and `hamqthLookup` flag correctly
- [ ] Verify **MSHV / JTDX / JS8Call** (`digital-mode-base.js`) receive the same enrichment as WSJT-X

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-# Claude Code local config
+# Claude Code local config — AI assistant instructions, not part of the project source
 .claude/
 CLAUDE.md
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Claude Code local config
 .claude/
+CLAUDE.md
 
 # Dependencies
 node_modules/
--- a/rig-bridge/README.md
+++ b/rig-bridge/README.md
@@ -438,6 +438,17 @@ By default, WSJT-X sends its decoded packets only to one listener. If you want b
 2. In Rig Bridge → Plugins → WSJT-X Relay → tick **Enable Multicast**, group address `224.0.0.1`
 3. Click **Save**
 
+### HamQTH callsign lookup (optional)
+
+When **HamQTH callsign lookup** is enabled in the WSJT-X Relay settings, Rig Bridge resolves unknown callsigns to country-level coordinates via the public [HamQTH DXCC API](https://www.hamqth.com/dxcc.php). This places map pins for stations whose FT8 message did not include a grid square.
+
+**What to know before enabling it:**
+
+- Lookups use the unauthenticated `dxcc.php` endpoint, which is intended for lookup tools. Rig Bridge caps requests at 2 per second globally and waits at least 60 seconds before retrying any individual callsign, so the traffic volume is modest even on a busy 20m FT8 band.
+- Results are cached for 24 hours in `hamqth-cache.json` (stored alongside `rig-bridge-config.json`) and survive restarts, so each callsign is typically looked up only once per day.
+- Lookups require outbound internet access (port 443) from the machine running Rig Bridge.
+- If HamQTH is unreachable, decodes simply show no lat/lon for unresolved callsigns — no errors are shown and operation is otherwise unaffected.
+
 ---
 
 ## APRS via Local TNC _(Beta)_
--- a/rig-bridge/core/server.js
+++ b/rig-bridge/core/server.js
@@ -820,6 +820,16 @@ function buildSetupHtml(version, firstRunToken = null) {
             </div>
           </div>
 
+          <!-- HamQTH callsign lookup (always visible when enabled) -->
+          <div class="checkbox-row" style="margin-top:10px;">
+            <input type="checkbox" id="wsjtxHamqth">
+            <span>HamQTH callsign lookup</span>
+          </div>
+          <div class="help-text" style="margin-top:-8px; margin-bottom:10px;">
+            Resolve unknown callsigns to country-level coordinates via the HamQTH DXCC database.
+            Results are cached for 24 h. Requires outbound internet access from this machine.
+          </div>
+
           <!-- Server relay fields — only shown in relay mode -->
           <div id="wsjtxServerOpts" style="display:none; margin-top:4px; padding:10px 12px; background:#1a1f2e; border:1px solid #2d3548; border-radius:6px;">
             <label>OpenHamClock Server URL</label>
@@ -1130,6 +1140,7 @@ function buildSetupHtml(version, firstRunToken = null) {
       document.getElementById('wsjtxMulticast').checked = !!w.multicast;
       document.getElementById('wsjtxMulticastGroup').value = w.multicastGroup || '224.0.0.1';
       document.getElementById('wsjtxMulticastInterface').value = w.multicastInterface || '';
+      document.getElementById('wsjtxHamqth').checked = !!w.hamqthLookup;
       // Set delivery mode radio
       const relayMode = !!w.relayToServer;
       document.getElementById('wsjtxModeSSE').checked = !relayMode;
@@ -1200,6 +1211,7 @@ function buildSetupHtml(version, firstRunToken = null) {
         multicast: document.getElementById('wsjtxMulticast').checked,
         multicastGroup: document.getElementById('wsjtxMulticastGroup').value.trim() || '224.0.0.1',
         multicastInterface: document.getElementById('wsjtxMulticastInterface').value.trim(),
+        hamqthLookup: document.getElementById('wsjtxHamqth').checked,
       };
       try {
         const res = await fetch('/api/config', {
@@ -1231,11 +1243,14 @@ function buildSetupHtml(version, firstRunToken = null) {
           if (!data.running) {
             el.textContent = 'Not running';
             el.style.color = '#6b7280';
-          } else if (!data.serverReachable) {
+          } else if (data.relayToServer && !data.serverReachable) {
             el.textContent = 'Running — connecting to server...';
             el.style.color = '#f59e0b';
           } else {
-            el.textContent = 'Running — ' + data.decodeCount + ' decodes, ' + data.relayCount + ' relayed';
+            let status = 'Running — ' + data.decodeCount + ' decodes';
+            if (data.relayToServer) status += ', ' + data.relayCount + ' relayed';
+            if (data.hamqthLookup) status += ', ' + data.callsignCacheSize + ' callsigns cached';
+            el.textContent = status;
             el.style.color = '#22c55e';
           }
         } catch (e) {}
--- a/rig-bridge/lib/wsjtx-enrich.js
+++ b/rig-bridge/lib/wsjtx-enrich.js
@@ -53,7 +53,7 @@ function isGrid(s) {
   if (!s || s.length < 4) return false;
   const g = s.toUpperCase();
   if (FT8_TOKENS.has(g)) return false;
-  return /^[A-R]{2}\d{2}(?:[A-Xa-x]{2})?$/.test(s);
+  return /^[A-R]{2}\d{2}(?:[A-X]{2})?$/.test(g);
 }
 
 // ──────────────────────────────────────────────────────────────────────────────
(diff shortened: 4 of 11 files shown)


alanhargreaves:
I'm happy that it does what it says it does. I'm not seeing any errors from the rig-bridge node.js nor extras from the clients.

I checked out the debug stuff that was in there, and the api that had some stats on port 5555. Also played with adding in the gridSource to the spots on WorldMap to verify that we were getting grid locs from different places.

I feel I was a bit close to this one to approve it, and I'd also be more comfortable with you doing the best-practice javascript once over, given there is some claude code in there.

ceotjoe:
@accius if you don't like the .gitignore leave it. I just want to avoid that this thing (Claude.md) get's into the repo accidentally.

accius (commented):
## Review — WSJT-X SSE enrichment

Really solid work. Local enrichment makes the SSE-only mode genuinely useful, and the separation (raw → server relay, enriched → bus) is clean. Good test plan too. A handful of issues to fix before merge:

**Bugs**

1. **Duplicate `type` key in `rig-bridge.js` decode forwarder:**
   ```js
   type: msg.message.startsWith('CQ') ? 'CQ' : 'QSO',
   ...
   type: msg.type,
   ```
   Both keys are on the same object literal — the second overwrites the first and the first line is dead (and throws `TypeError` if `msg.message` is ever undefined, which is guarded elsewhere with `?? ''`). Remove the first.

2. **`clientStates[msg.id]?.deCall` / `deGrid` never populated.** In both `digital-mode-base.js` and `wsjtx-relay.js`, `clientStates[msg.id]` is set to `{ band, dialFrequency, mode }` only. The QSO_LOGGED handler falls back to `clientStates[msg.id]?.deCall || null`, which is always `undefined → null`. Either store `deCall`/`deGrid` from STATUS into `clientStates`, or drop the fallback and accept `myCall`/`myGrid` being null when WSJT-X doesn't supply them.

3. **`isGrid` case-sensitivity bug:** the regex `/^[A-R]{2}\d{2}(?:[A-Xa-x]{2})?$/` has no `i` flag and is tested against the *original* input (not the uppercased `g`). A 4-char lowercase grid (`em28`) returns false. FT8 messages are uppercase in practice, so low-impact, but trivial to fix — test against `g`, or add the `i` flag.

4. **`pluginBus.on('wspr', …)` forwards `msg` directly** as `data`, including the internal `source` field and no field whitelisting — every other handler carefully selects fields. Be consistent, and be careful not to leak fields the UI shouldn't see.

**Safety / robustness**

5. **HamQTH XML parsed by regex** (`<lat>([^<]+)<\/lat>`). Works for the current endpoint but brittle against whitespace/attributes. Consider a minimal XML parser or at least matching against `<lat[^>]*>(-?[0-9.]+)</lat>` to constrain to numerics.

6. **HamQTH public `dxcc.php`** — confirm this endpoint permits unauthenticated automated use at our expected volume (5 concurrent × many users). If ToS requires a session token, this breaks silently. Worth documenting in the README.

7. **No persistent HamQTH cache.** Every rig-bridge restart re-queries — not a correctness issue, but if we're asking users to enable this across the fleet we may want to persist a JSON cache file.

8. **Rate limiting only via `HAMQTH_MAX_CONCURRENT = 5`.** There is no overall QPS limit; on a busy band a burst of unknown calls could trigger dozens of requests per minute. Consider a short (~2s) per-callsign cool-down plus a global QPS cap.

9. **`SEEN_DECODE_MAX = 2000`** — reasonable, but check memory behavior on a very active 20m FT8 session over 24h. Should be fine with FIFO eviction.

**Minor**

10. `.gitignore` adds `CLAUDE.md` — harmless, but unrelated to this PR's topic. Happy to leave it.
11. `dt` is stored as a pre-formatted string (`toFixed(1)`) in `enrichDecode`, but `useWSJTX` and the UI may want the numeric form for sorting. Worth passing both or keeping numeric and formatting at render time.
12. Setup HTML adds hardcoded colors `#1a1f2e`, `#2d3548`, `#6b7280` — matches existing style in this file, so not a regression, but a note for a future cleanup.
13. `.gitignore`-ing `CLAUDE.md` with no comment — a one-line comment would help future readers.

**Checklist:** all UI-related items are unchecked — please verify at least the Modern layout WSPR tab and the CQ filter in PSKReporterPanel before merge.

Overall: fix 1–3 (real bugs) and this is ready.

ceotjoe:
Thanks for the thorough review — all points have been addressed across two follow-up commits.

---

### Bugs (1–4) + XML regex (5) — commit 2b19810

1. **Duplicate `type` key** — removed the dead `msg.message.startsWith('CQ')` line.
2. **`clientStates` missing `deCall`/`deGrid`** — fixed in both `wsjtx-relay.js` and `digital-mode-base.js`; STATUS now stores `deCall: msg.deCall ?? null` and `deGrid: msg.deGrid ?? null`, so the QSO_LOGGED fallback actually resolves.
3. **`isGrid` case bug** — regex now tests against the uppercased `g` instead of the original input.
4. **`wspr` bus handler** — switched to an explicit field whitelist consistent with all other handlers.
5. **HamQTH XML regex** — tightened to `/<lat[^>]*>(-?[0-9.]+)<\/lat>/` (and `lng` equivalent).

---

### Rate limiting (8) — commit a316a03

Two guards added inside `triggerHamqthLookup` in `wsjtx-enrich.js`:

- **Per-callsign cooldown** — an optional `lastAttemptedMap` (5th parameter, a `Map`) is set before the request is dispatched. Any subsequent call for the same callsign within 60 s is dropped immediately. The `wsjtx-relay` plugin creates and passes this map.
- **Global QPS cap** — a module-level sliding-window array (`_hamqthQpsWindow`) tracks timestamps of recent outbound requests. Any call that would push the rate above 2 req/s is silently dropped.

Together these keep traffic bounded even during a fast CQ burst on a busy 20 m FT8 session.

---

### Persistent cache (7) — commit a316a03

- Added `loadCallsignCache(filePath, cache)` and `saveCallsignCache(filePath, cache)` helpers to `wsjtx-enrich.js`. Save uses an atomic temp-file rename to avoid a corrupt file on crash mid-write.
- `createCallsignCache().set()` now accepts an optional `timestamp` argument so loaded entries preserve their original expiry time rather than getting a fresh 24 h TTL on every restart.
- `wsjtx-relay` derives the path `hamqth-cache.json` from `CONFIG_DIR` (alongside `rig-bridge-config.json`), loads the cache at `connect()`, does a debounced 30 s write after each successful lookup, and flushes immediately on `disconnect()`.

---

### `dt` as numeric (11) — commit a316a03

`enrichDecode` and `enrichWspr` now store `dt: msg.deltaTime ?? 0` (float). `PSKReporterPanel` formats it at render time with sign and one decimal place — `${Number(d.dt) >= 0 ? '+' : ''}${Number(d.dt).toFixed(1)}` — which also handles legacy string values from the server polling path gracefully via `Number()` coercion.

---

### HamQTH ToS note (6) — commit a316a03

Added a **HamQTH callsign lookup** section to `rig-bridge/README.md` covering: the unauthenticated endpoint, the rate limits we enforce, the 24 h cache and its persistence, network requirements, and graceful degradation when the service is unreachable.

---

### Minor notes

- **`.gitignore` comment (13)** — added `# Claude Code local config — AI assistant instructions, not part of the project source`.
- **Hardcoded colours in setup HTML (12)** — noted; left for a separate cleanup pass as they match the existing file style.
- **`SEEN_DECODE_MAX` (9)** — no change; FIFO eviction at 2000 entries is well within budget for a 24 h session.
- **`.gitignore` unrelated entry (10)** — happy to leave it as-is.

accius (approved):
## Re-review — WSJT-X SSE enrichment (updated)

Great work addressing the prior review. Every blocker and most of the nice-to-haves are in.

### Items 1–5 (bugs / safety) — fixed
- **Duplicate `type` key** — first line removed, second (`type: msg.type`) is the one that stays.
- **`clientStates[msg.id]` now stores `deCall`/`deGrid`** from STATUS (`wsjtx-protocol.js` confirms STATUS carries these as operator identity). QSO_LOGGED fallback `msg.myCall || clientStates[msg.id]?.deCall || null` now actually resolves.
- **`isGrid` case-sensitivity** — now tests against the uppercased `g`, so `em28` resolves.
- **`wspr` bus handler** — fields are whitelisted (`clientId`, `isNew`, `time`, `timeMs`, `snr`, `dt`, `frequency`, `band`, `drift`, `callsign`, `grid`, `power`, `offAir`, `lat`, `lon`, `timestamp`) consistent with the other bus handlers.
- **HamQTH XML regex** — tightened to `-?[0-9.]+` with attribute-tolerant opening tag.

### Items 6–8 (rate limiting / cache / docs) — fixed
- **`rig-bridge/README.md`** documents the HamQTH ToS context, rate limits, and cache behaviour.
- **Per-callsign 60 s cooldown** via `lastAttemptedMap` — set even on attempt, so failed lookups don't hammer.
- **Global 2 req/s sliding-window QPS cap** — module-level `_hamqthQpsWindow` shared across plugin instances.
- **Persistent cache** — `hamqth-cache.json` loaded on connect, saved on disconnect, with a debounced 30s save after successful lookups. Atomic temp-file+rename on write. Expiry respected on load.

### Item 11 — fixed
`dt` stored as float, formatted at render time in `PSKReporterPanel` with `+/-` and `.toFixed(1)`. Handles both legacy string values (server path) and numeric via `Number(d.dt)`, with a null guard for empty display.

### Items 10, 13 — addressed
`.gitignore` CLAUDE.md entry now has an explanatory one-line comment.

### Minor observations (non-blocking)
- **`package.json` bump to 2.1.0** — reasonable signal for the added-feature surface (HamQTH lookup, enrichment pipeline). No downstream consumers reading this version so no compatibility concerns.
- **`loadCallsignCache` swallows all errors** — appropriate for non-critical cache, but a single `console.warn` on JSON parse failure (vs. missing file) might help a user debugging an unexpectedly cold cache. Not required.
- **`hamqth-cache.json` size** — at a typical `CALLSIGN_CACHE_TTL` of 24 h and a busy FT8 session, this could grow to a few hundred entries per day. Reasonable; no bound needed.
- **QPS window array** — growing then shifting is fine for 2 req/s, but if the cap is ever raised to something like 50, switch to a ring buffer or count-based guard.

No blockers. Approving.

PR #902, opened 2026-04-10
fix(rbn): restore PORT in ctx destructuring so spotter-mode spots appear on map

### LINKED ISSUE #897: [BUG] RBN "What a skimmer is hearing" not quite working yet?
## Describe the Bug

I appreciate that my suggestion of adding the ability to use RBN as a propagation aide, to display *any* calls that a particular skimmer is hearing.  However, it doesn't look like it's quite working yet.  I can setup the the panel and put my desired skimmer in, but I don't see any of the stations it is hearing appearing on the map.

## To Reproduce

Add the RBN panel to the Map
Change mode to "What does a skimmer hear?"
Add callsign of a desired skimmer (i.e. KD2OGR, W2NNN, WS2C, etc..)
Note the number of Spot heard, reported at the top of the panel, but...
None of these spots show up on the map.

## Expected Behavior

I would expect the Spots heard by the selected skimmer to show up on the map, with paths, much the same way as the DX Cluster spots are shown on the map.

## Screenshots

Image

## Environment

Windows 11, Chrome browser

73,
Thanks,
Alan

ceotjoe:
## Summary

Fixes #897 — RBN "What does a skimmer hear?" spots show a correct count in the panel but never appear on the map.

- **Root cause:** `PORT` was present in the shared `ctx` object passed to `rbn.js` but was not included in the module's destructuring. Every internal callsign-lookup call (`enrichSpotWithLocation`, `enrichSpotWithDXLocation`, `/api/rbn/location/`) referenced the undeclared variable inside a `try/catch`, so the resulting `ReferenceError` was silently swallowed, a `_failed` marker was cached for 10 minutes, and **no spot ever received coordinates**. The stats counter read from the raw spot count (before enrichment), so it looked correct while the map stayed empty.
- **Fix:** Add `PORT` to the `ctx` destructuring at the top of `server/routes/rbn.js` (one line). This unblocks all three location-lookup paths and allows `enrichSpotWithDXLocation` to resolve DX station coordinates via the `/api/callsign/` endpoint (which itself falls back to prefix estimation, so virtually any valid callsign resolves).
- **Coding compliance:** Three falsy `&&` checks on `skimmerLat`/`skimmerLon` in `useRBN.js` were replaced with `!= null` guards, preventing silent breakage for stations on the equator (lat 0) or prime meridian (lon 0).
- **Debug logging:** Spotter mode now logs `dxLat`/`dxLon`/`dxGrid` per spot instead of the irrelevant skimmer coordinates, making future diagnosis possible without a code change.

## Test plan

- [ ] Add the RBN panel to the map and set mode to **"What does a skimmer hear?"**
- [ ] Enter a known-active skimmer callsign (e.g. `KD2OGR`, `W2NNN`, `WS2C`)
- [ ] Verify the spot count in the panel header matches the number of markers and paths drawn on the map
- [ ] Confirm the **default DX mode** ("who hears me?") also draws skimmer markers and paths correctly
- [ ] Confirm no regression on band/SNR/time-window filtering in both modes
- [ ] Test with a skimmer near lat=0 or lon=0 if possible (equator/prime meridian edge case)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Claude Code local config
 .claude/
+CLAUDE.md
 
 # Dependencies
 node_modules/
--- a/server/routes/rbn.js
+++ b/server/routes/rbn.js
@@ -10,6 +10,7 @@ module.exports = function (app, ctx) {
   const {
     fetch,
     CONFIG,
+    PORT,
     logDebug,
     logInfo,
     logWarn,
--- a/src/plugins/layers/useRBN.js
+++ b/src/plugins/layers/useRBN.js
@@ -282,16 +282,22 @@ export function useLayer({
         // Log spot details
         if (mySpots.length > 0) {
           mySpots.forEach((spot, idx) => {
-            console.log(
-              `  ${idx + 1}. Skimmer: ${spot.callsign}, Freq: ${spot.freqMHz} MHz, SNR: ${spot.snr} dB, Band: ${spot.band}, Grid: ${spot.grid || 'MISSING'}, Lat: ${spot.skimmerLat || '?'}, Lon: ${spot.skimmerLon || '?'}`,
-            );
+            if (queryMode === 'spotter') {
+              console.log(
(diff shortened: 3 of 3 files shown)


ceotjoe:
@accius I'm not very familiar with the code in this area and not sure about how to test this properly. But the fix seems easy and straight forward. Do you have someone at hand who can check this?

PR #904, opened 2026-04-12
header, temperature metric or US customary

MichaelWheeley:
## What does this PR do?

uses global temperature units selection to show C or F in header but not both

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist

WAS
was Screenshot 2026-04-12 093034

BECOMES
becomes F Screenshot 2026-04-12 093547

OR
become celcius Screenshot 2026-04-12 093457

### DIFF
--- a/src/components/Header.jsx
+++ b/src/components/Header.jsx
@@ -8,6 +8,7 @@ import DonateButton from './DonateButton.jsx';
 import { QRZToggle } from './CallsignLink.jsx';
 import { ctyLookup, isCtyLoaded } from '../utils/ctyLookup';
 import { getFlagUrl } from '../utils/countryFlags';
+import { formatTemperature, formatTemperatureBoth } from '../utils/formatWeather.js';
 
 export const Header = ({
   config,
@@ -37,9 +38,6 @@ export const Header = ({
   const clockSize = `${(isMobile ? 16 : 24) * scale}px`;
   const statsSize = `${(isMobile ? 10 : 13) * scale}px`;
   const labelSize = `${(isMobile ? 10 : 13) * scale}px`;
-
-  const allUnits = config.allUnits || {};
-  const isMetricTemp = allUnits.temp === 'metric';
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


ceotjoe:
I'm not sure about this. On one hand it is adjusting the metrics like we do in other areas. On the other hand it is the only reference to both metrics systems on the dashboard. PR itself is fine, it's a general UX design question. Do we want to keep reference to the other metrics system within the header? @accius @alanhargreaves

alanhargreaves:
@MichaelWheeley indicated that the change was for consistency and I think I can live with that. I **do**  like the idea of adding both temperatures to the mouseover display. If you could add that to this PR, I think we have a winner.

accius (commented):
## Review — Header temperature unit

Small, clean change. Honors the user's `allUnits.temp` preference instead of always showing both units.

**Minor suggestions (non-blocking):**

1. **Defensive defaulting:** `const allUnits = config.allUnits || {};` is good — consider making it `config?.allUnits || {}` in case `config` is ever undefined on first render (other spots in this file destructure it from props so it should be set, but the guard is cheap).

2. **Reuse elsewhere:** other places in the app (e.g. weather panel, dx weather) already handle this by passing `allUnits` down; consider a tiny `formatTempC(rawC, allUnits.temp)` helper in `utils/` to avoid the inline ternary scattering if this pattern is repeated.

3. **Loss of dual display:** some users may actually prefer both. Worth confirming with the wider user base — but I think following the global unit setting is the right default.

4. **Checklist** — please tick the "tested in Dark/Light/Retro + mobile" boxes; this is a header change so it's quick to verify.

Approved in spirit — happy to merge once the quick test-checks are confirmed.

MichaelWheeley:
@alanhargreaves
> @MichaelWheeley indicated that the change was for consistency and I think I can live with that. I **do** like the idea of adding both temperatures to the mouseover display. If you could add that to this PR, I think we have a winner.

done.
image

MichaelWheeley:
@accius 
> **Minor suggestions (non-blocking):**
> 
> 
> 
> 1. **Defensive defaulting:** `const allUnits = config.allUnits || {};` is good — consider making it `config?.allUnits || {}` in case `config` is ever undefined on first render (other spots in this file destructure it from props so it should be set, but the guard is cheap).

Noted, removed temperature formatting to utils/formatWeather.js and passed config?.allUnits
> 2. **Reuse elsewhere:** other places in the app (e.g. weather panel, dx weather) already handle this by passing `allUnits` down; consider a tiny `formatTempC(rawC, allUnits.temp)` helper in `utils/` to avoid the inline ternary scattering if this pattern is repeated.
 
Done with addtional exported formatters, please review. 
> 3. **Loss of dual display:** some users may actually prefer both. Worth confirming with the wider user base — but I think following the global unit setting is the right default.

Per separate discussion added dual temperature to the mouse-over.

> 4. **Checklist** — please tick the "tested in Dark/Light/Retro + mobile" boxes; this is a header change so it's quick to verify.

Checked and done!

PR #907, opened 2026-04-13
[Bug 898] UDP (local) spots can show an incorrect location of the spotter

### LINKED ISSUE #898: [BUG] UDP (local) spots can show an incorrect location of the spotter
## Describe the Bug

When delivering spots to OHC in UDP (local) mode, using a local app (like SDC's SKM Server or a logger program), the spotter QTH can be wrong.  This can happen with vanity callsigns (like mine, K3CDY), where the spotter appears to be in Maryland, when in fact I'm in Denver.  This can also happen when an operator moves to a different QTH, but keeps the same callsign.

The map below shows local CW spots captured by SDC at my QTH in Denver, but the spotter appears in Maryland, based upon "traditional" QTH determination.

Image

## To Reproduce

1. In SDC's SKM Server settings, set up the spotter field using a vanity callsign (K3CDY, for example):

Image

2. Let SDC do its job, and send spots to OHC.
3. Witness that the spotter "lines" point  to Maryland, even though I'm near Denver.

## Expected Behavior

When using UDP (local) mode, the spotter's QTH should be based upon the operator's **gridsquare** as entered in OHC - and NOT rely on outdated FCC conventions that determine QTH.

## Environment

- **OS**: Ubuntu 22.04
- **Browser**: Safari  26.4
- **Node.js Version**: 23.6.0
- **OpenHamClock Version**: 26.2.1
- **Running As**: Local Linux network PC build

## Console Errors

N/A

## Additional Context

I  am the author/contributor of the UDP spots feature in OHC.  I would be happy to provide the fix if you like.

---

73 de K3CDY

alanhargreaves:
## What does this PR do?

When using Local UDP for a DX Cluster source, some programs do not list a lat/lon or grid square. When this happens, `s.spotterGrid` is set to `null` and the hooks code tries to find the location based purely on the callsign prefix, which can be problematical.

This fix uses `CONFIG.gridSquare` instead of just defaulting to `null` if no gridSquare is offered.

We should also **not** offer "UDP Spots(local)" as an option for DX Cluster Source **unless** `isLocalInstall` is true. Otherwise we could have random people setting up UDP listeners on servers like OHC.

If the Settings panel notes that we are not local **and** `dxClusterSource` is set to 'udp' it resets it to 'auto'.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. If using Local UDP for DX Cluster, `curl  localhost:3000/api/dxcluster/paths`
2. Verify that the `spotterGrid` is set to the local grid square

## Checklist


accius (commented):
## Review — UDP spotter location fallback (draft)

Minimal, targeted fix. The one-line change is sensible for the local-UDP case: the "spotter" in a local UDP feed is effectively the operator themselves, so falling back to the server's `CONFIG.gridSquare` (which in self-hosting mode = the operator's grid) is reasonable when the upstream program doesn't supply one.

**Concerns / questions**

1. **`CONFIG.gridSquare` semantics in a multi-tenant deployment.** On the public server at `openhamclock.com`, `CONFIG.gridSquare` is the server operator's grid, not the individual user's — if this code path is ever reached in the hosted deployment, we'd be tagging every UDP spot with the server's location. Can you confirm this `udpSession` code path is only hit in self-hosted mode? If there's any chance it runs for hosted users, please guard with an `isLocal` / `isSelfHosted` check or pass the user's own grid in via the session.

2. **`CONFIG` in scope:** please double-check that `CONFIG` is imported/available at the top of this route, and that `CONFIG.gridSquare` is the field name (vs `CONFIG.locator` — the frontend config uses `locator`). A grep for other `CONFIG.gridSquare` references in `server/` would be reassuring.

3. **Nit:** consider falling back to `CONFIG.locator` if `CONFIG.gridSquare` isn't set — same source, different name on the frontend config, and either could be the canonical field depending on how the server reads the user's config.

4. **PR description typos** ("teh", "preoblematical", "gris") — tidy up before marking ready.

Happy to re-review when out of draft.

alanhargreaves:
Hmmm, I was under the apparently mistaken impression that this could only run for a local session.

OK, a better approach might be to where we have the setting selection for this to have it also request a default grid square where it defaults to the users DE, and pass that when we configure the session.

I can see this will need a bit more work and thought.

Thank you for picking up on that.

alanhargreaves:
as to locator, it looks like we set it in config.js, but read it in to gridSquare

```text
alan@Alans-MacBook-Pro openhamclock % find server -type f -exec grep -i config.locator {} +
server/config.js:  gridSquare: locator || jsonConfig.locator || '',
alan@Alans-MacBook-Pro openhamclock %
```

If I just look for locator I get

```text
alan@Alans-MacBook-Pro openhamclock % find server -type f -exec grep  locator {} +
server/config.js:  console.log('[Config] ⚠️  Please edit .env with your callsign and locator, then restart');
server/config.js:// Convert Maidenhead grid locator to lat/lon (used only during config init)
server/config.js:// Get locator from env (support both LOCATOR and GRID_SQUARE)
server/config.js:const locator = process.env.LOCATOR || process.env.GRID_SQUARE || '';
server/config.js:// Calculate lat/lon from locator if not explicitly set
server/config.js:if ((!Number.isFinite(stationLat) || !Number.isFinite(stationLon)) && locator) {
server/config.js:  const coords = gridToLatLon(locator);
server/config.js:  gridSquare: locator || jsonConfig.locator || '',
server/routes/callsign.js:  // Convert Maidenhead grid locator to lat/lon (center of grid square)
server/routes/callsign.js:  // Try to extract grid locators from a comment string
server/routes/config-routes.js:      locator: CONFIG.gridSquare,
server/routes/pskreporter.js:  // Topic format: pskr/filter/v2/{band}/{mode}/{sendercall}/{receivercall}/{senderlocator}/{receiverlocator}/{sendercountry}/{receivercountry}
server/routes/pskreporter.js:      // Validate grid: 4 or 6 character Maidenhead locator
server/routes/rbn.js:        const url = `https://retrieve.pskreporter.info/query?mode=WSPR&flowStartSeconds=${flowStartSeconds}&rronly=1&nolocator=0&appcontact=openhamclock&rptlimit=2000`;
server/utils/grid.js: * Grid locator (Maidenhead) and frequency/band utilities.
server/utils/grid.js: * Convert Maidenhead grid locator to lat/lon.
server/utils/grid.js: * @param {string} grid - Grid locator (e.g., 'DM79' or 'DM79lv')
server/utils/grid.js: * Convert lat/lon to Maidenhead grid locator (6-character).
```

alanhargreaves:
I think that with making it so you can't set the dxcluster type to 'udp' unless `isLocalInstall` is true, that I've addressed your concerns.

alanhargreaves:
It just occurred to me, ...

In the settings, should I check to see if it's not a local install **and** the "udp" option is selected, and change it to the recommended?

I'm also happy to say that it's an option they should not be using on a non local install so they just get to select a new one without default code.

accius (commented):
## Re-review — UDP spotter location (updated)

Two changes since the last pass:

### 1. `CONFIG.gridSquare` fallback for missing `spotterGrid` — good
Replacing `s.spotterGrid || null` with `s.spotterGrid || CONFIG.gridSquare || null` is the right call. UDP spots come from the station's own local network, so when the source doesn't include a grid the station's configured grid is the correct attribution.

Edge to consider: if `CONFIG.gridSquare` is unset or still `N0CALL`-era default, the fallback is still `null` — which is the previous behavior. Safe.

### 2. Gating the "UDP Spots (Local Network)" option behind `isLocalInstall` — good
UDP ingestion requires the server to be on the same LAN as the rig/WSJT-X. Hosted users couldn't use it anyway, so hiding it removes a confusing non-functional option. `isLocalInstall` is already imported in `SettingsPanel` via `useLocalInstall()` hook (line 83), so no wiring changes needed.

**One follow-up to verify:** if a hosted user has previously set `dxClusterSource = 'udp'` in their saved config (e.g., from a prior local-install session that got migrated, or from direct JSON edit), the `<select>` will now render with no matching `<option>`. Browsers typically show a blank value in that case, and the saved state isn't corrupted — but it's worth confirming the dropdown gracefully falls back to `auto` rather than leaving a dangling value. If it's a concern, you could coerce to `auto` on mount when `!isLocalInstall && source === 'udp'`.

No blockers — ready to merge from my side once you're happy with the fallback-on-migration question.

alanhargreaves:
I've added the following lines to `SettingsPanel()`

```javascript
  // Set a sane default if we are not a local installation and we have 'udp' set as the dxClusterSource.
  if (!isLocalInstall && dxClusterSource === 'udp') setDxClusterSource('auto');
```

I don't have any dx cluster software to test it, but the SettingsPanel loads without error for me.

PR #910, opened 2026-04-14
fix: reject invalid server timezone (e.g. Etc/Unknown) before it reac…

### LINKED ISSUE #908: [BUG] RangeError: invalid time zone: Etc/Unknown
## Describe the Bug

I've been deploying the Openhamclock via Portainer as a stack so that it is always up-to-date locally. However, after recent updates, an error pops up: "OpenHamClock hit an error. Something went wrong rendering the dashboard. This usually fixes itself after a reload. If it keeps happening, try clearing your browser cache." It happens with and without the callsign and locator being set properly. 

> Error details
> 
> RangeError: invalid time zone: Etc/Unknown
> 
> hw@https://openhamclock.mydomain.com/assets/index-tx2xjj17.js:1265:110
> Dg@https://openhamclock.mydomain.com/assets/index-tx2xjj17.js:17:10993
> mw@https://openhamclock.mydomain.com/assets/index-tx2xjj17.js:1265:12053

## To Reproduce

1. Deploy Openhamclock from https://github.com/accius/openhamclock.git via Portainer (Stack) 
2. Clear your browser cache and and try to load it.

## Expected Behavior

Even without setting the variables callsign and locator, I'd expect to be able to load it without issues. 

## Environment

- **OS**: OMV - 7.7.24-7 (Sandworm)
- **Browser**: Firefox 149.0.2
- **OpenHamClock Version**: [26.1.3]
- **Running As**: [Docker/Portainer stack]

---

73!

ceotjoe:
## Summary

- On minimal Linux Docker containers without a `TZ` environment variable set, Node's `Intl.DateTimeFormat().resolvedOptions().timeZone` returns `"Etc/Unknown"`. This string is accepted by Node but rejected by browsers with `RangeError: invalid time zone: Etc/Unknown`, crashing the entire dashboard.
- **Server fix** (`server/routes/config-routes.js`): validate the resolved timezone with `Intl.DateTimeFormat` in a try/catch before sending it to the client. Any invalid value (including `"Etc/Unknown"`) is replaced with `''`, which tells the client to fall back to the browser's own timezone.
- **Client fix** (`src/hooks/app/useTimeState.js`): apply the same validation guard before passing the timezone to `toLocaleTimeString`/`toLocaleDateString`, so a bad value can never crash the dashboard even if it somehow slips through in the future.

## Reproduction

1. Deploy via Docker/Portainer **without** a `TZ` environment variable set
2. Open the app — dashboard fails immediately with `RangeError: invalid time zone: Etc/Unknown`

## Workaround for existing deployments

Add `TZ=Europe/Berlin` (or your local IANA timezone) to the Portainer stack environment variables. This fix makes that optional rather than required.

## Test plan

- [ ] Deploy via Docker without `TZ` set — dashboard loads, local time shows browser timezone
- [ ] Deploy via Docker with `TZ=America/New_York` — local time reflects that timezone correctly
- [ ] Open Settings → Timezone, manually select a timezone — still works as before
- [ ] Verify no regression on bare-metal / non-Docker deployments

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/docs/DOCKER.md
+++ b/docs/DOCKER.md
@@ -26,6 +26,10 @@ LOCATOR=FN20
 TZ=America/New_York
 ```
 
+> **`TZ` is optional** — if omitted, each visitor's browser timezone is used for
+> the local-time display. Setting it is still recommended so that server-side
+> timestamps (logs, cache TTLs, etc.) match your local time.
+
 Restart to apply:
 
 ```bash
--- a/server/routes/config-routes.js
+++ b/server/routes/config-routes.js
@@ -180,9 +180,13 @@ module.exports = function (app, ctx) {
         const tz = process.env.TZ || Intl.DateTimeFormat().resolvedOptions().timeZone || '';
         if (!tz) return '';
         try {
-          Intl.DateTimeFormat('en-US', { timeZone: tz });
+          new Intl.DateTimeFormat(undefined, { timeZone: tz });
           return tz;
         } catch (e) {
+          console.warn(
+            '[config] Invalid resolved timezone "%s" — falling back to empty (client will use browser TZ). Set TZ env var to silence.',
+            tz,
+          );
           return '';
         }
       })(),
--- a/server/routes/config-routes.js
+++ b/server/routes/config-routes.js
@@ -173,8 +173,19 @@ module.exports = function (app, ctx) {
       // Whether config is incomplete (show setup wizard)
       configIncomplete: CONFIG.callsign === 'N0CALL' || !CONFIG.gridSquare,
 
-      // Server timezone (from TZ env var or system)
-      timezone: process.env.TZ || Intl.DateTimeFormat().resolvedOptions().timeZone || '',
+      // Server timezone (from TZ env var or system), validated.
+      // On minimal Linux containers without TZ set, Intl can return "Etc/Unknown"
+      // which browsers reject with RangeError. Validate before sending.
+      timezone: (() => {
+        const tz = process.env.TZ || Intl.DateTimeFormat().resolvedOptions().timeZone || '';
+        if (!tz) return '';
+        try {
+          Intl.DateTimeFormat('en-US', { timeZone: tz });
+          return tz;
+        } catch (e) {
+          return '';
+        }
+      })(),
 
       // Feature availability
       features: {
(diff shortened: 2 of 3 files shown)


accius (commented):
## Review — reject invalid server timezone

Clean, targeted bug fix with defense-in-depth (validate on the server, validate again on the client). Correct diagnosis of the `Etc/Unknown` case and well-scoped.

**Suggestions (non-blocking):**

1. **Memoize the client-side validation.** `useTimeState` runs its `try { Intl.DateTimeFormat(...) } catch` on every render. Not expensive, but trivially improved by:
   ```js
   const safeTimezone = useMemo(() => {
     if (!timezone) return '';
     try { new Intl.DateTimeFormat(undefined, { timeZone: timezone }); return timezone; }
     catch { return ''; }
   }, [timezone]);
   ```
   and then use `safeTimezone` below.

2. **Log at least once server-side** when the fallback kicks in, so self-hosters see something useful in the logs rather than silently getting browser-local time. A single `console.warn('[config] Invalid resolved timezone "%s" — falling back to empty (client will use browser TZ). Set TZ env var to silence.', tz)` is enough.

3. **`Intl.DateTimeFormat('en-US', …)`** — the locale is irrelevant for the validation; `new Intl.DateTimeFormat(undefined, { timeZone: tz })` reads a little more clearly.

4. **Consider documenting in the Docker/self-hosting README** that `TZ` is optional after this fix but recommended for accurate server-side time stamps (logs, cache TTLs, etc.).

**Test plan** items are not ticked — please run at least the "Docker without TZ" repro and confirm before merge.

Approved as-is if the `useMemo` nit is out of scope; happy to merge.

ceotjoe:
Thanks for the thorough review — all four points addressed in the follow-up commit (d045c41).

- **Memoized validation** — wrapped in `useMemo([timezone])` so the `Intl.DateTimeFormat` probe only fires when the timezone value actually changes.
- **`console.warn` on fallback** — server now logs the bad value with a hint to set `TZ`, so self-hosters have something actionable in their container logs.
- **Locale removed** — both the server and client now use `new Intl.DateTimeFormat(undefined, …)`.
- **Docs** — added a note to `docs/DOCKER.md` that `TZ` is optional after this fix but recommended for server-side timestamp accuracy (logs, cache TTLs).

Docker without TZ runs as expected in my case.

PR #912, opened 2026-04-15
[BUG 911] isDaytime in PropagationPanel should check local sunrise/set rather than 06:00/18:00 UTC

### LINKED ISSUE #911: [BUG] isDaytime in PropagationPanel should check local time rather than UTC
## Describe the Bug

In `src/components/PropagationPanel.jsx`  we have

```javascript
  const isDaytime = new Date().getUTCHours() >= 6 && new Date().getUTCHours() <= 18;
```

This should be based of the time of the DE station.

In addition, we actually KNOW exactly when sunrise and sunset are, we should use that information.

Going further down the rabbit hole, it appears we are displaying sunrise/sunset (at least in Dockable) in UTC rather than local time.

## To Reproduce

1. Go to 'BAND CONDITIONS', when day/night local does not match UTC
2. Note that it shows daylight based on UTC

## Expected Behaviour

It **should** be showing daylight based on the location of the DE station.

So, if we change the above javascript to use `currentTime` (why go and look it up again when we already have it), AND compare with `deSunTimes`, that should address the first issue.

The display of the sunrise/sunset in local times should "just" need a conversion to local time. We could simplify this, doing it in one place by adding a `local` element to `deSunTimes`, and do similar for `dxSunTimes` and use that in the display of the dx station information.

## Screenshots

Image

## Environment

All.

---

73!

alanhargreaves:
## What does this PR do?

Pass currentTime and deSunTimes to `PropagationPanel` and use them for determining if it is daytime. Correct `isDaytime` to account for sunrise < sunset if we are on the other side of the world from UTC as we are doing a local sunrise/sunset in UTC.

Added `safeTimeZone` to the `deSunTimes` memo dependencies, so that if the timezone changes we update the local times in it.

Display sunrise/sunset in local time, ensuring we also display the timezone. For DX locations we display in UTC, explicitly state 'UTC'.

In order to display the timezones for the sunrise/sunset in the Classic layout, I had to decrease the point size to make it fit without wrapping (and looking ugly).

The DE and DX displays also now will not try to display a range if sunrise contains either 'Polar night' or 'Midnight sun'

Taking the opportunity to correct the extra seconds display in Classic
image

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. In the propagation panel compare the daytime setting to what you see out the window
2. Note that in the DE information panel we see the correct sunrise/sunset with the timezone displayed
3. Note that in the DX information panel that the sunrise/sunset times are listed as UTC
4. Change your DE location to something like RQ29mv
5. Note that 'Midnight Sun' is displayed without trying to do a range for the sunrise/sunset in each of Dockable, Classic and Modern.
6. Change your timezone to one that should now be in Summer time (eg America/Los_Angeles).
7. Note that the displayed sunrise/sunset times have changed to that timezone
8. Move your DE to somewhere just about to have the terminator cross it into night time
9. Verify that the sunset time is close to the current local time
10. Verify that as the time hits sunset, that the propagation panel changes to show night time.

## Checklist

- [X] Tested in Modern, Classic and Dockable layouts


denete:
Shipit.

accius (changes requested):
## Review — isDaytime local sunrise/sunset (#911)

Right idea, but the implementation has a few blockers that will prevent it from working as intended.

### Blockers

**1. `deSunTimes.sunset()` called as a function — will throw at polar locations**

In `PropagationPanel.jsx:136`:

```js
currentTime.getUTCHours() <= deSunTimes.sunset()
```

`deSunTimes.sunset` is a string (e.g. `"19:45"`), not a function. At mid-latitudes you're saved by short-circuit: the left side is always `false` (see #2) so the call is never reached. But at polar locations where `deSunTimes.sunset = ''` (empty string), `"" > "Polar night"` is false, so the ternary takes the **else** branch — and `deSunTimes.sunrise()` throws `TypeError`.

**2. Number-vs-string comparison — `isDaytime` is effectively always `false` at non-polar locations**

`currentTime.getUTCHours()` returns a number. `deSunTimes.sunrise` is `"06:23"`. Comparing `15 >= "06:23"` coerces `"06:23"` → `NaN`, and every comparison against `NaN` returns `false`. So the `isDaytime` flag never turns true under the "normal" branch. Users always see the night label.

To fix both #1 and #2, convert sunrise/sunset to minutes-since-midnight and compare against `currentTime.getUTCHours() * 60 + currentTime.getUTCMinutes()`:

```js
const toMin = (hhmm) => {
  const [h, m] = hhmm.split(':').map((s) => parseInt(s, 10));
  return Number.isFinite(h) && Number.isFinite(m) ? h * 60 + m : null;
};
const sr = toMin(deSunTimes.sunrise);
const ss = toMin(deSunTimes.sunset);
const cur = currentTime.getUTCHours() * 60 + currentTime.getUTCMinutes();
const isDaytime =
  sr == null || ss == null
    ? currentTime.getUTCHours() >= 6 && currentTime.getUTCHours() <= 18 // polar fallback
    : ss > sr
      ? cur >= sr && cur <= ss
      : cur >= sr || cur <= ss;
```

### Issues

**3. "UTC" fallback label is dead code**

In `ClassicLayout.jsx`, `ModernLayout.jsx`, `DockableApp.jsx`:

```js
<span>{deSunTimes.local !== null ? config.timezone : 'UTC'}</span>
```

`convertTimeUTCtoLocal` always returns an object, so `deSunTimes.local !== null` is always truthy — the `'UTC'` fallback never shows. At polar locations where sunrise/sunset are strings like `'Polar night'` / `''`, `convertTimeUTCtoLocal` would produce `"NaN:NaN"` instead of falling back to UTC.

**4. Timezone offset mismatch**

`convertTimeUTCtoLocal` uses `new Date().getTimezoneOffset()` — the **browser's** timezone. But the adjacent label shows `config.timezone` (the app-configured timezone). If a user has set `config.timezone` to anything other than their browser TZ (e.g., remote operator, travelling, displaying a remote station's local time), the displayed hours are wrong for the label shown.

Given the hook already receives `timezone` as a parameter and computes `safeTimezone`, sun times should be formatted through `Intl.DateTimeFormat` using `safeTimezone` rather than re-doing offset math with the browser clock:

```js
const fmtInTz = (utcDate, tz) =>
  new Intl.DateTimeFormat('en-GB', { hour: '2-digit', minute: '2-digit', hour12: false, timeZone: tz }).format(utcDate);
```

Build a `Date` for each UTC sunrise/sunset and format with the configured TZ — avoids the browser-TZ assumption and handles DST naturally.

**5. Polar edge cases unhandled**

`calculateSunTimes` returns `{ sunrise: 'Polar night', sunset: '' }` or `{ sunrise: 'Midnight sun', sunset: '' }`. `convertTimeUTCtoLocal` tries to `.split(':')` these and `parseInt()` the parts → `NaN` → display shows `"NaN:NaN"`. Guard at the top of `convertTimeUTCtoLocal` and return the original `sunTimes` unchanged (or don't set `.local`) when the values aren't in `HH:MM` form.

### Minor

- **Font-size change** (`16px → 12px`) in `ClassicLayout.jsx` lines 864 and 889 is unrelated to the bug fix and makes the sun-time text visibly smaller. If intentional, mention it in the PR description; otherwise revert.
- **`var` usage** in `convertTimeUTCtoLocal` — the rest of the file uses `const`/`let`.
- **`parseInt` without radix** — `parseInt(rise.hr, 10)` is safer.
- **`convertTimeUTCtoLocal` defined inside the hook body** means it's rebuilt every render. Move it out of the hook (it's pure).
- **DX sun times stay UTC**, and the label is now explicit "UTC" — good.

### Testing suggestions

- Verify `isDaytime` toggles correctly at your location at sunrise + sunset (the bug makes it stick on night).
- Test at a polar-night / midnight-sun location (e.g., lat 80°) to confirm no crashes or `NaN:NaN` display.
- Test with `config.timezone` set to a non-browser TZ to confirm displayed sun times match the label.

alanhargreaves:
1. Corrected
2. I took a different approach. in `useTimeState` where I calculate `deSunTimes`, I am using a different method, similar to what you suggested), to get the string "local"times, and saving sunriseMin and sunSetMin in local time in deSunTimeslocal. I also needed to move the definition of `safeTimezone` so I could use it. `convertTimeUTCtoLocal` is now also aware of te possibility of 'Midnight sun' and 'Polar night'. Then I change `isDayTime` in `PropagationPanel` to be

   ```javascript
     const currentLocalMin = (currentTime) => {
       [hr,mn] = currentTime.toLocaleString("en-US", {
         timeZone: timeZone,
         hour12: false,
         hour: "numeric",
         minute: "numeric"
       }).split(':').map(Number);
       return (hr % 24 * 60) + mn;
     }
     const isDaytime = 
       (deSunTimes.sunset === '')
       ? deSunTimes.sunrise === 'Midnight sun'
       : currentLocalMin >= deSunTimes.local.sunriseMin && currentMin <= deSunTimes.local.sunsetMin;   ```
       which will
      - return true - if Midnight sun
      - return false - if Polar night
      - return the appropriate value otherwise based on sunrise/sunset
  
3. Removed the dead code and Polar night/Midnight sun addressed above. I also made the DE and DX panels aware of 'Midnight sun' and 'Polar night'.
4. Addressed in 2 above
5. Addressed in 3 above

Moved `convertTimeUTCtoLocal()` out of the hook.

We no longer user parseInt, using `.map(Number)` off the `split(':')` instead.

Now when fixing the various layouts I also noticed in classic that we do this
<img width="289" height="160" alt="image" src="https://github.com/user-attachments/assets/0e969923-de41-4009-b7c4-69974aff93dc" />
Now it looks like we explicitly

```javascript
             <div
                style={{
                  fontSize: '44px',
                  fontWeight: '700',
                  color: '#00ff00',
                  fontFamily: 'JetBrains Mono, monospace',
                  lineHeight: 1,
                }}
              >
                {utcTime}
                <span style={{ fontSize: '26px', color: '#00cc00' }}>:{seconds}</span>
              </div>
              <div style={{ fontSize: '18px', color: '#00aa00', marginTop: '4px' }}>
                {utcDate} <span style={{ color: '#888' }}>UTC</span>
              </div>
```

Is there a reason for this? or should we delete that last span?

Discussing it with @ceotjoe we agreed that the {seconds} span should be deleted (done).

I have not tested with
- `isDaytime` changing
- different timezone to local

I can do these tomorrow.

alanhargreaves:
Right, I've run the extra testing and fond a couple of problems which have now been addressed. It's looking a lot better now and probably ready for another review.

accius (changes requested):
## Re-review — isDaytime local sunrise/sunset (round 2)

Big improvement over round 1 — the two hard blockers (the `sunset()` call and the number-vs-string comparison) are both fixed, and the polar branch is now explicitly guarded. A few smaller items from the last pass are still open, and two new issues popped up while reviewing the reworked hook.

### Fixed since prior review

- **Blocker 1 — `deSunTimes.sunset()` call.** Gone. `isDaytime` no longer invokes `sunrise`/`sunset` as functions (`src/components/PropagationPanel.jsx:150-155`). Polar case is short-circuited at the top of the ternary via `deSunTimes.sunset === ''` so the problematic expression is never reached.
- **Blocker 2 — number-vs-string comparison.** Fixed. The hook now writes `local.sunriseMin` / `local.sunsetMin` as integers (minutes-since-midnight), and `isDaytime` compares them against a newly-computed `currentLocalMin` (`src/components/PropagationPanel.jsx:137-155`, `src/hooks/app/useTimeState.js:48-53`). Wrapping-across-midnight is handled with the `sunsetMin > sunriseMin` check on both sides.
- **Issue 3 — dead "UTC" fallback label.** Removed from `ClassicLayout.jsx`, `ModernLayout.jsx`, `DockableApp.jsx`. The label now either shows `config.timezone` or the entire sunset segment is hidden at polar locations (`src/layouts/ClassicLayout.jsx:863-872`, `src/layouts/ModernLayout.jsx:221-233`, `src/DockableApp.jsx:478-491`).
- **Issue 4 — browser-TZ vs app-TZ mismatch.** `convertTimeUTCtoLocal` now takes an explicit `tz` argument and formats via `Date.toLocaleString('en-US', { timeZone: tz, ... })`, which handles DST correctly. The caller passes `safeTimezone` (`src/hooks/app/useTimeState.js:6-55, 117-122`).
- **Issue 5 — polar strings produced `"NaN:NaN"`.** `convertTimeUTCtoLocal` now bails out at the top when `sunTimes.sunset === ''` and returns the original object, so `'Polar night'` / `'Midnight sun'` flow through untouched and `fmtSunTime` renders them as `--:--` in the UI (`src/hooks/app/useTimeState.js:9-11`).
- **Minor — `var` → `let`, `parseInt` → `.map(Number)`, function moved out of the hook body.** All done (`src/hooks/app/useTimeState.js:6-55`).

### Still broken / regressed

- **Font-size change in `ClassicLayout.jsx`** (`16px → 12px` at lines 863 and 892) is still bundled in with the bug fix. Low-stakes, but still scope creep for a bug PR — if deliberate, call it out in the PR description; otherwise revert.

### New concerns

- **`deSunTimes` useMemo dep array is missing `safeTimezone`, and `dxSunTimes` has it backwards.** In `src/hooks/app/useTimeState.js:117-127`:

  ```js
  const deSunTimes = useMemo(() => {
    let sunTimes = calculateSunTimes(configLocation.lat, configLocation.lon, currentTime);
    sunTimes.local = convertTimeUTCtoLocal(sunTimes, safeTimezone, currentTime);
    return sunTimes;
  }, [configLocation, currentTime]);            // <-- uses safeTimezone, missing from deps
  const dxSunTimes = useMemo(
    () => calculateSunTimes(dxLocation.lat, dxLocation.lon, currentTime),
    [dxLocation, currentTime, safeTimezone],    // <-- doesn't use safeTimezone
  );
  ```

  The two arrays look accidentally swapped. If a user changes their timezone in settings, `deSunTimes.local` will tick over on the next minute because `currentTime` is also in the deps, so the practical impact is small — but the pair is still wrong and `eslint-plugin-react-hooks/exhaustive-deps` should flag both.

- **Only one of four `PropagationPanel` instances in `DockableApp.jsx` gets `timeZone`.** In `src/DockableApp.jsx:773-830` the `case 'propagation'` branch adds `timeZone={config.timezone}` (line 785), but `'propagation-chart'`, `'propagation-bars'`, and `'band-conditions'` do not. `isDaytime` is computed at the top of `PropagationPanel` regardless of `viewMode`, and it's actually *rendered* in the bands view (`src/components/PropagationPanel.jsx:369-374`), which is the forced mode behind `case 'band-conditions'`. So users who dock the band-conditions panel get `currentTime.toLocaleString('en-US', { timeZone: undefined, ... })`, i.e. browser TZ — the exact regression we were trying to avoid for the DE panel label. Easiest fix: add `timeZone={config.timezone}` to all four call sites (and to the two call sites in `ModernLayout.jsx` already do this, which is nice).

- **Invalid `config.timezone` will crash the `deSunTimes` memo.** `Date.toLocaleString('en-US', { timeZone: '' })` throws `RangeError: Invalid time zone specified`. `safeTimezone` returns `''` when `timezone` is falsy or invalid (`src/hooks/app/useTimeState.js:107-115`), and `convertTimeUTCtoLocal` passes that straight into `toLocaleString`. Guard:

  ```js
  const fmtOpts = { hour12: false, hour: '2-digit', minute: '2-digit' };
  if (tz) fmtOpts.timeZone = tz;
  local.sunrise = rise.date.toLocaleString('en-US', fmtOpts);
  ```

  (Same pattern is already used a few lines below for `localTimeOpts` / `localDateOpts`.) The same crash path exists in `PropagationPanel`'s `currentLocalToMin` if `timeZone` ever comes in as `''` rather than `undefined`.

- **Unrelated visual change: duplicate-seconds removal.** Commit `39da4c7` drops `<span style={{ fontSize: '26px', color: '#00cc00' }}>:{seconds}</span>` (`src/layouts/ClassicLayout.jsx:838`). Factually correct — `utcTime` already carries `HH:MM:SS` — but it also throws away the deliberate treatment of making seconds smaller at 26px vs 44px. Unrelated to #911; should probably be its own PR so the tradeoff is discussed.

- **Self-reference in polar case is a little surprising.** `convertTimeUTCtoLocal` returns the original `sunTimes`, and the caller assigns that back as `sunTimes.local = sunTimes` — so `deSunTimes.local === deSunTimes`. It works, but `deSunTimes.local.sunset === ''` only because of the self-loop. A small comment at the top of the function (or returning `{ sunrise: sunTimes.sunrise, sunset: '' }`) would prevent someone from accidentally mutating `local` and reaching through to the parent.

- **`currentLocalToMin` reallocates every render.** Minor perf/code-tidy thing; could be wrapped with `useMemo` keyed on `[currentTime, timeZone]`, or inlined.

### Verdict

**CHANGES_REQUESTED** — the original two blockers are fixed and the core logic is right, but the missing `timeZone` prop on three of the four `DockableApp` call sites reintroduces the exact bug this PR is meant to fix for the dockable Band Conditions panel, and the mis-matched `useMemo` dep arrays plus the empty-timezone crash are small but real correctness issues that should land before merge.

alanhargreaves:
## Point Size Change

I did call out the point size change in the PR description.
image
I'll add some more words to make it clearer.

## useMemo dependencies

Ooops. I wonder how that even worked while I was testing. Fixed.

## Pass `timeZone` to PropagationPanel in Dockable for all instances

Fixed.

## Invalid `config.timezone` will crash the `deSunTimes` memo

Fixed.

## Removal of duplicate seconds leaves us with "large" seconds in time display

Used utcTime.substring(0,5) to extract the hh:mm and put the smaller display of seconds back.
image

## Self Reference

That was an error on my part. I **know** it would copy by reference, and I **meant** to do by value. Changed to pass the elements.

## currentLocalToMin() issues

Re-did it as an inline function. Trying as a `useMemo` was giving me react issues that I could not work out.

accius (approved):
Round-2 items all addressed across commits 255186f, ac1f0cd, 63aa488, a1c4573 + a341c33 (console.debug removed) + 164dba0 (Staging merged in to pick up the new \`**/__fixtures__/\` prettier ignore so format CI passes).

Verified on the merged tree:
- All 4 \`PropagationPanel\` instances in DockableApp pass \`timeZone\`
- \`useMemo\` deps correct on both \`deSunTimes\` and \`dxSunTimes\`
- Empty-timezone guard in \`convertTimeUTCtoLocal\` (\`if (tz) fmtOps.timeZone = tz;\`)
- Polar branch returns a fresh \`local\` object (no self-reference)
- Classic seconds restored at smaller font via \`utcTime.substring(0,5)\` + \`:{seconds}\` span

Format / test (20.x) / test (22.x) all green. Approving — thanks for the careful iteration.

accius:
## Production regression report — `RangeError: invalid time zone specified` on first load

After v26.3.1 hit production today (PR #956 merge), users started hitting **`RangeError: invalid time zone specified`** on initial page load. Workaround: the user clicks "Clear cache and reload", and it goes away.

### Root cause

`src/components/PropagationPanel.jsx:147-158` calls:

```js
currentTime
  .toLocaleString('en-US', {
    timeZone: timeZone,   // ← receives raw `config.timezone` from all 7 call sites
    hour12: false,
    hour: 'numeric',
    minute: 'numeric',
  })
```

…and `src/utils/config.js:37` defaults `config.timezone` to **empty string**:

```js
timezone: '', // IANA timezone (e.g. 'America/Regina') — empty = browser default
```

The comment is wrong: `Intl.DateTimeFormat({ timeZone: '' })` does **not** fall back to the browser default — it throws `RangeError: Invalid time zone specified`. Same goes for any other invalid value such as `Etc/Unknown`.

So any user whose `config.timezone` is empty/invalid (which is the **default** for fresh installs, and what's stored in many users' localStorage from prior versions that didn't surface a TZ picker) crashes the panel on every render.

### Why "clear cache + reload" fixes it

Clearing localStorage drops the persisted blank `timezone`, and the next config load can pick up `serverConfig.timezone` (set from `.env`/`/api/config`) which on most deployments is a real IANA name. Hard reload also drops any stale bundle.

### Why this didn't bite during the 11 days on Staging

Staging users mostly already had a populated `config.timezone` in localStorage from earlier configuration; the empty-default path wasn't exercised on first visit. Production absorbed 295 commits at once, including the localStorage-shape changes from the v26.3.0 layouts work, so any user with a stale-but-not-purged config that lacked `timezone` (or had it as `''`) hit the unguarded path.

### This was flagged at review time

From the original review of this PR: *"empty-TZ crash path"* was called out as a blocker but the PR was squash-merged. The validation pattern actually exists in `useTimeState.js:107-115` already:

```js
const safeTimezone = useMemo(() => {
  if (!timezone) return '';
  try {
    new Intl.DateTimeFormat(undefined, { timeZone: timezone });
    return timezone;
  } catch {
    return '';
  }
}, [timezone]);
```

…and the same hook also gates the option correctly:

```js
const localTimeOpts = { hour12: use12Hour };
if (safeTimezone) {
  localTimeOpts.timeZone = safeTimezone;   // only set if non-empty
}
```

PropagationPanel does **not** apply either guard.

### Suggested fix (minimal)

In `PropagationPanel.jsx`, mirror the `useTimeState` pattern — only set `timeZone` when it's truthy and validates:

```js
const safeTimeZone = useMemo(() => {
  if (!timeZone) return undefined;
  try {
    new Intl.DateTimeFormat(undefined, { timeZone });
    return timeZone;
  } catch {
    return undefined;
  }
}, [timeZone]);

const currentLocalMin = function () {
  const opts = { hour12: false, hour: 'numeric', minute: 'numeric' };
  if (safeTimeZone) opts.timeZone = safeTimeZone;
  let [hr, mn] = currentTime.toLocaleString('en-US', opts).split(':').map(Number);
  return (hr % 24) * 60 + mn;
};
```

Or — simpler — export `safeTimezone` from `useTimeState` and thread that through to all 7 PropagationPanel call sites instead of raw `config.timezone`. That keeps the validation logic in one place and prevents this class of bug from coming back the next time someone adds another `toLocaleString` call.

Happy to send the fix as a follow-up PR if useful.

accius:
Hotfix landed on main as `68ebe67` via PR #957. PropagationPanel now applies the same `safeTimeZone` validation pattern used in `useTimeState.js`, so the empty-default `config.timezone` no longer crashes the panel. Misleading "empty = browser default" comment on `DEFAULT_CONFIG.timezone` corrected at the same time. Users hitting the error should be unstuck on next page load (no manual cache clear required).

PR #924, opened 2026-04-24
[Feature/sat] Add above horizon but below minimum elevation indication

MichaelWheeley:
## What does this PR do?

(satellite) prior to PR the two possible indication states were 'Below Horizon' or 'VISIBLE'. Visibility only occurred when the satellite exceeded the minimum elevation so there was an ambiguous range where the satellite could be just above the horizon but still labeled as 'Below Horizon' which could lead to confusion.

A new state is introduced 'Below Minimum Elevation' where the satellite is above the horizon but below minimum elevation.

As shown in the graphics below the states are,
* Below Horizon
* Below Minimum Elevation
* Visible

Claude was used for translation of key word for all languages.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [X] Translation
- [ ] Map layer plugin

## Checklist

image
image
image


MichaelWheeley:
@accius this PR is failing checks but it appears to be a fault of the workflow, see https://github.com/accius/openhamclock/pull/925

accius:
Hi Michael — I dug into the CI failure. You were right that the workflow file was causing it, but the fix needed to happen on this branch too. Here's what was going on and what I just pushed:

**Root cause:** `.github/workflows/publish-p533-data.yml` on Staging has `contents: write  # ...` with two spaces before the `#` comment. Prettier wants one space. When CI runs on a PR, it tests the *merge* of your branch into Staging, so your PR inherits that formatting violation even though you never touched the file.

**What I pushed to your branch** (with your maintainer_can_modify permission):
1. `996a044` — merge `origin/Staging` into your branch (brings you up to date with 40-ish commits of WASM/P.533 work)
2. `52e011d` — the one-character prettier fix (`  #` → ` #`) on the workflow file

CI is re-running now — the `format` check should pass this time. Your actual feature changes (the 16 language files + `useSatelliteLayer.js`) were never the problem and are untouched.

Side note: PR #925 (from your `temp` branch) was going to fix the same workflow file. Once this PR merges, #925 becomes redundant and we'll close it.

PR #864, opened 2026-04-02
[improvement-854] Added variable debug message functionality for

denete:
## What does this PR do?

 Added variable debug message functionality for messages sent to the console. This overrides current console.* and also provides for future scoped logger.* functionality.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

(Notes on the usage of querystring values are included in src/debug/README.md)

1. Open the JS console
2. Note that no console messages show once this has loaded in the app
3. Add query string to URL and verify the different levels of console output

Examples:
```
?log=none
?log=error
?log=warn
?log=info
?log=debug
?log=all
```

## Checklist

- [X] App loads without console errors
- [ ] Tested in **Dark**, **Light**, and **Retro** themes
- [ ] Responsive at different screen sizes (desktop + mobile)
- [ ] If touching `server.js`: caches have TTLs and size caps (we serve 2,000+ concurrent users)
- [ ] If adding an API route: includes caching and error handling
- [ ] If adding a panel: wired into Modern, Classic, and Dockable layouts
- [ ] No hardcoded colors — uses CSS variables (`var(--accent-cyan)`, etc.)
- [ ] No `.bak`, `.old`, `console.log` debug lines, or test scripts included

PR #931, opened 2026-04-26
[Feature/satellite] adds next pass and ending

MichaelWheeley:
## What does this PR do?

Feature adds satellite next pass information, and when visible the time remaining until the 'end' time when it goes below the minimum elevation.

Next passes are calculated every hour in the useSatellites.js

Bug fix on orbit.js where the number of passes calculated was always one more that the maximum requested.

possible improvement - remove logging (which is enabled = false)

Should the 'Ending' time be positive and count down, or be negative and count towards zero ?

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

Satellite below minimum elevation, will be visible in about a minute ('Next Pass:')
Next Pass Screenshot 2026-04-25 194240

Satellite visible for the next 3 minutes ('Ending:')
Next Pass Visible ending in 3 minutes Screenshot 2026-04-25 194924

Satellite now below minimum elevation, next visible in about 90minutes ('Next Pass:')
Next Pass no longer visible below minimum elevation next pass about 90 minutes Screenshot 2026-04-25 195224

Satellite now below horizon
Next Pass below horizon next pass about 90 minutes Screenshot 2026-04-25 195353

### DIFF
--- a/src/hooks/useSatellites.js
+++ b/src/hooks/useSatellites.js
@@ -19,8 +19,6 @@ export const useSatellites = (observerLocation, satelliteConfig) => {
   const [loadingNextPass, setLoadingNextPass] = useState(true);
   const [tleData, setTleData] = useState({});
 
-  const enableLogging = false; // Set to true to enable detailed logging for debugging
-
   // Fetch TLE data
   useEffect(() => {
     const fetchTLE = async () => {
@@ -199,14 +197,15 @@ export const useSatellites = (observerLocation, satelliteConfig) => {
     const minElevation = satelliteConfig?.minElev || 5.0;
     const maxPasses = 2;
 
-    if (enableLogging) {
+    // debug logging
+    {
       const formatDate = (date) => date.toISOString().slice(0, 19).replace('T', ' ');
       let logStr = `[Satellite] calculating next passes,`;
       logStr += `\n observer lat=${groundStation.latitude}, lon=${groundStation.longitude}, alt=${groundStation.height}m,`;
       logStr += `\n time range=${formatDate(startDate)} to ${formatDate(endDate)},`;
       logStr += `\n minElevation=${minElevation}°`;
       logStr += `\n maxPasses=${maxPasses}`;
-      console.log(logStr);
+      console.debug(logStr);
     }
 
     const nextPasses = [];
@@ -242,9 +241,9 @@ export const useSatellites = (observerLocation, satelliteConfig) => {
     // Sort alphabetically by name for a consistent, static list
     nextPasses.sort((a, b) => a.name.localeCompare(b.name));
 
-    // log passes for debugging
-    if (enableLogging) {
-      const formatDate = (date) => date.toISOString().slice(0, 19).replace('T', ' ');
+    // debug logging
+    {
+      const formatDate = (date) => new Date(date).toISOString().slice(0, 19).replace('T', ' ');
       nextPasses.forEach(({ name, startTimes, endTimes }) => {
         let logStr = `[Satellite] Next passes for ${name}: `;
         if (startTimes.length === 0) {
@@ -256,7 +255,7 @@ export const useSatellites = (observerLocation, satelliteConfig) => {
           });
         }
 
-        console.log(logStr);
+        console.debug(logStr);
       });
     }
 
--- a/src/hooks/useSatellites.js
+++ b/src/hooks/useSatellites.js
@@ -215,8 +215,8 @@ export const useSatellites = (observerLocation, satelliteConfig) => {
     }
 
     const nextPasses = [];
-    try {
-      Object.entries(tleData).forEach(([name, tle]) => {
+    Object.entries(tleData).forEach(([name, tle]) => {
+      try {
         // Handle both line1/line2 and tle1/tle2 formats
         const line1 = tle.line1 || tle.tle1;
         const line2 = tle.line2 || tle.tle2;
@@ -237,10 +237,10 @@ export const useSatellites = (observerLocation, satelliteConfig) => {
           startTimes,
           endTimes,
         });
-      });
-    } catch (e) {
-      // Skip satellites with invalid TLE
-    }
+      } catch (e) {
+        // Skip satellite with invalid TLE, continue processing others
+      }
+    });
 
     // Sort alphabetically by name for a consistent, static list
     nextPasses.sort((a, b) => a.name.localeCompare(b.name));
--- a/src/hooks/useSatellites.js
+++ b/src/hooks/useSatellites.js
@@ -5,6 +5,7 @@
  */
 import { useState, useEffect, useCallback } from 'react';
 import * as satellite from 'satellite.js';
+import Orbit from '../utils/orbit.js';
 
 function round(value, decimals) {
   const factor = Math.pow(10, decimals);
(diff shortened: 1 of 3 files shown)


accius (changes requested):
Thanks for this — the orbit.js off-by-one fix alone is worth landing, and the next-pass / ending countdown is a really nice UX touch. CI is green and I tested the visual states locally against your screenshots and they match.

A few things to address before merge — one real bug, the rest are nits the author can take or leave.

### 🔴 Blocker — silent skip after first bad TLE

`useSatellites.js` wraps the entire `Object.entries(tleData).forEach(...)` in a single \`try/catch\`:

\`\`\`js
try {
  Object.entries(tleData).forEach(([name, tle]) => {
    ...
    const orbit = new Orbit(name, \`${name}\\n${line1}\\n${line2}\`);
    const passes = orbit.computePassesElevation(...);
    ...
    nextPasses.push({ ... });
  });
} catch (e) {
  // Skip satellites with invalid TLE
}
\`\`\`

If any one satellite throws (malformed TLE, propagation edge case in \`computePassesElevation\`, etc.), the catch fires and the forEach aborts — every subsequent satellite loses its next-pass data for that hour. The comment says "Skip satellites with invalid TLE" but the implementation actually skips ALL remaining sats after the first bad one. Move the \`try/catch\` *inside* the forEach:

\`\`\`js
Object.entries(tleData).forEach(([name, tle]) => {
  try {
    ...
  } catch (e) {
    // skip this sat, keep processing others
  }
});
\`\`\`

### 🟡 Style nits

- **Debug logging block** — \`const enableLogging = false;\` plus the \`console.log\` blocks should come out (PR template literally says "no console.log debug lines"). If you want to keep a runtime opt-in, OHC has the existing \`?log=info|debug|all|none\` URL-param pattern (#864) you could hook into.
- **\`if (typeof enableLogging !== 'undefined' && enableLogging)\`** — the variable is always defined at component scope, so the \`typeof\` guard is dead. Just \`if (enableLogging)\`.
- **\`pass.start && pass.end && (startTimes.push(pass.start), endTimes.push(pass.end))\`** — comma-operator side-effect. A plain \`if (pass.start && pass.end) { startTimes.push(...); endTimes.push(...); }\` reads cleaner.
- **\`/******…*****/\` banner-comment lines** — there are ~25 of them sprinkled around every existing useEffect / helper in both files. There is precedent in \`useSatelliteLayer.js\` from earlier work, but the new ones balloon the diff with no information value. Consider stripping them so the substantive change reads more clearly.
- **Hook return shape** — \`useSatellites\` now returns \`{ data, loading, loadingNextPass, nextPassData }\` but \`loadingNextPass\` and \`nextPassData\` aren't read by any consumer (they're only used internally via the merge into \`data\`). Safe to drop from the return object until something needs them.

### 🟢 Design call you asked about

> "Should the 'Ending' time be positive and count down, or be negative and count towards zero?"

The \`+HH:MM:SS\` prefix reads ambiguously — looks like elapsed time, not time-remaining. Two options that read better:

1. **No sign** — \`Ending: 00:03:12\`. The label tells you it's a countdown.
2. **Minus sign** — \`Ending: -00:03:12\` (time until zero).

I'd vote (1). Same applies to "Next Pass:".

### 🟢 What's good

- The \`orbit.js\` off-by-one fix (\`> maxPasses\` → \`>= maxPasses\`) is correct.
- The visible / below-min-elev / below-horizon state transitions in the info window look clean — your screenshots show the three states behaving exactly as you'd want.
- The hourly-only re-computation cadence is the right call for a 7-day forward simulation across ~46 sats.
- React deps are correct (\`observerLocation\`, \`tleData\`, \`satelliteConfig\`) — config changes will re-trigger immediately.

Once the silent-skip is fixed and the debug log block is gone, I'm happy to merge. The rest you can take or leave. Thanks again — happy to push the bug fix directly to your branch if you'd rather (\`maintainerCanModify: true\` on the PR).

MichaelWheeley:
## @accius on April 26 you said,

> ...
> ### 🔴 Blocker — silent skip after first bad TLE
> 
> 
> 
> `useSatellites.js` wraps the entire `Object.entries(tleData).forEach(...)` in a single \`try/catch\`:
>
> ...
> 
*moved try/catch inside satellite processing loop, if one fails it continues processing the remainder*
> 
> ### 🟡 Style nits
> 
> - **Debug logging block** — `const enableLogging = false;` plus the \`console.log\` blocks should come out (PR template literally says "no console.log debug lines"). If you want to keep a runtime opt-in, OHC has the existing \`?log=info|debug|all|none\` URL-param pattern (#864) you could hook into.
> 

*modified to write to `console.debug()` which is suppressed without `/?log=debug` URL, removed `const enableLogging = false;` statement*

> - **\`if (typeof enableLogging !== 'undefined' && enableLogging)\`** — the variable is always defined at component scope, so the \`typeof\` guard is dead. Just \`if (enableLogging)\`.
> 
*removed `typeof enableLogging !== 'undefined'`*
> - **\`pass.start && pass.end && (startTimes.push(pass.start), endTimes.push(pass.end))\`** — comma-operator side-effect. A plain \`if (pass.start && pass.end) { startTimes.push(...); endTimes.push(...); }\` reads cleaner.
> 

*changed `x && func()` to `if(x)func()`*

> - **\`/******…*****/\` banner-comment lines** — there are ~25 of them sprinkled around every existing useEffect / helper in both files. There is precedent in \`useSatelliteLayer.js\` from earlier work, but the new ones balloon the diff with no information value. Consider stripping them so the substantive change reads more clearly.
> 

*removed*

> - **Hook return shape** — \`useSatellites\` now returns \`{ data, loading, loadingNextPass, nextPassData }\` but \`loadingNextPass\` and \`nextPassData\` aren't read by any consumer (they're only used internally via the merge into \`data\`). Safe to drop from the return object until something needs them.
> 

*`loadingNextPass, nextPassData` dropped from return of `useSatellites`*

> 
> ### 🟢 Design call you asked about
> 
> > "Should the 'Ending' time be positive and count down, or be negative and count towards zero?"
> ...
>

*removed + from secondsFromNow indication, and also in 'From Now' indication of modal*

> 
> ...
> 
> Once the silent-skip is fixed and the debug log block is gone, I'm happy to merge. The rest you can take or leave. Thanks again — happy to push the bug fix directly to your branch if you'd rather (\`maintainerCanModify: true\` on the PR).

*all points fixed and ready for review / merge*

accius (approved):
Re-reviewed against the 2026-04-26 comments. Author pushed 8 follow-ups — the blocker is gone and the headline nits are fixed.

### ✅ Resolved
- 🔴 Silent-skip — `try/catch` is now inside `forEach`, so one bad TLE no longer aborts the rest of the batch.
- `enableLogging` boolean + `console.log` block — replaced with `console.debug`, hidden by default.
- Dead `typeof enableLogging !== 'undefined'` guard — gone.
- `/*****/` banner-comment lines — swept.
- Hook return shape — kept at `{ data, loading }`; the unused fields didn't ship.
- `+HH:MM:SS` countdown prefix — dropped from both modal and side panel.
- `orbit.js` off-by-one — `> maxPasses` → `>= maxPasses`. ✓

### ⚠️ Still outstanding — nits, not blockers
None of these need to land before merge; flagging for next pass.

1. **Comma-operator wasn't fully unwound.** `useSatellites.js:225-229`:
   ```js
   passes.forEach((pass) => {
     if (pass.start && pass.end) {
       (startTimes.push(pass.start), endTimes.push(pass.end));
     }
   });
   ```
   The outer `&&`-as-statement was replaced with `if`, but the parenthesised comma expression inside is still there. Two plain statements would read cleaner.
2. **`orbit.error && console.warn(...)`** at `useSatellites.js:220` — same short-circuit-as-statement pattern. New code, same nit.
3. **Always-on debug blocks.** The two `console.debug` blocks (~L200, ~L248) are wrapped in bare `{ ... }` scopes that always execute — they build the log strings even though `console.debug` is hidden. Cheap at ~46 sats hourly, but not free.

Approving and merging — thanks for the quick turnaround on the blocker. Happy to take the three nits in a follow-up if you want, or leave as-is.

PR #932, opened 2026-04-27
Client side logging cleanup

alanhargreaves:
## What does this PR do?

Tidies up some client side logging. Moving a lot of messaging out of `console.log()` into levels such as `console.warn()`, `console.info()` and `console.debug()`.

Removed an unnecessary debug from `usePOTA`.

Made the messaging in `useEarthquake` and `usePropagation`  use similar formatting to everything else.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Load the web site with varying degrees of log level.

## Checklist

It was kind of the point that we see console messages from this change

### DIFF
--- a/src/components/WorldMap.jsx
+++ b/src/components/WorldMap.jsx
@@ -1704,7 +1704,7 @@ export const WorldMap = ({
 
       // Initialize state ONLY on first mount (when empty)
       if (Object.keys(pluginLayerStates).length === 0) {
-        console.log('Loading saved layer states:', initialStates);
+        console.debug('Loading saved layer states:', initialStates);
         setPluginLayerStates(initialStates);
       }
 
--- a/src/hooks/app/useAppConfig.js
+++ b/src/hooks/app/useAppConfig.js
@@ -65,9 +65,9 @@ export default function useAppConfig() {
     setConfig(newConfig);
     saveConfig(newConfig);
     applyTheme(newConfig.theme || 'dark');
-    console.log('[Config] Saved to localStorage:', newConfig.callsign);
+    console.debug('[Config] Saved to localStorage:', newConfig.callsign);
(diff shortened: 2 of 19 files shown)


MichaelWheeley:
created as console_log.code-search which is directly openable, but renaming as console_log.code-search.txt for github upload

[console_log.code-search.txt](https://github.com/user-attachments/files/27183754/console_log.code-search.txt)

from /src these are the remaining console.log() calls, should they also be converted?

alanhargreaves:
I think some of the console.log calls are ok (eg initialisation stuff). I'm happy to go through the rest of them, though. Moving this back to draft.

alanhargreaves:
I've taken @MichaelWheeley 's input into account in here and it's now ready for review.

accius (approved):
Approving — broad but mechanical refactor: noisy `console.log` calls move to the appropriate level (`debug` for per-event chatter, `info` for one-shot lifecycle events, `warn`/`error` retained where appropriate).

- The signal/noise separation makes browser DevTools much more usable — debug calls hide behind the verbosity toggle, info-level lifecycle messages stay visible.
- Touched files all match the same pattern (downgrade `console.log` → `console.debug` or `console.info`, keep formatting consistent), no logic changes spotted.
- One small structural improvement that's already in this PR: prefixes like `[Earthquakes]` and `[usePropagation]` are now consistent with the rest of the codebase.
- Minor pre-existing issue you flagged in the WASM/useState commit reversion — agree that handling an "expected error" by catching it is kludgy, but separating that into its own bug is the right call rather than coupling it to a logging cleanup.
- CI green; no behavior change.

PR #935, opened 2026-04-30
[util] geo.js extends maidenhead grid support

MichaelWheeley:
## What does this PR do?

- Extends Maidenhead grid locator functionality in `src/utils/geo.js`
- Adds size 2, 4, (6), 8 Maidenhead grid locator support, e.g. DM / DM12 / DM12KV / DM12KV99
- Code is based on the library already imported into OHC, `@hamset/maidenhead-locator` [link](https://github.com/HoshinoSuzumi/lib-maidenhead-locator), however this library only supports size 6. 
- In making this change it was noted that duplicate versions exist and that they should probably all use the common `geo.js` code, see below.
- Adds `geo.test.js` with unit test for Maidenhead grid locator functionality.

Why? I was trying to create a map layer that revealed size two or size three grid squares and needed the bounding-box coordinates. No other code appears to **fully** support the Maidenhead standard.

### noted that,

- ### The following are functionally the same,
  - `parseGridSquare` from `geo.js`
  - `maidenheadToLatLon` from `@hamset/maidenhead-locator` (`package.json`)
  - coded function `maidenheadToLatLon` (exported) from `server\routes\callsign.js`

- components using `parseGridSquare` from `geo.js`
  - `src\components\DXFavorites.jsx` imported but not used
  - `src\components\DXGridInput.jsx`
  - `src\components\SettingsPanel.jsx`
  - `src\layouts\EmcommLayout.jsx`
  - `src\plugins\layers\useWinlinkGateways.js`

- components that define function `maidenheadToLatLon` and export it 
  - `server\routes\callsign.js`

- components that import `maidenheadToLatLon` but uncertain if it is coming from `server\routes\callsign.js` or `@hamset/maidenhead-locator`
  - `server\routes\dxcluster.js`
  - `server\routes\n1mm.js`
  - `server\routes\rbn.js`
  - `server\routes\wsjtx.js`

- components that import `maidenheadToLatLon` but don't use it
  - `server\routes\propagation.js`
  - `server\routes\pskreporter.js`

- components that use `maidenheadToLatLon` but it appears not imported? possibly code not functional?
  - `src\plugins\layers\useN3FJPLoggedQSOs.js`

- ### The following are functionally the same,
  - `calculateGridSquare` or `latLonToMaidenhead` from `./geo.js`
  - `WGS84ToMaidenhead` from `@hamset/maidenhead-locator`

- components using `WGS84ToMaidenhead` from `@hamset/maidenhead-locator`;
  - `src\hooks\usePOTASpots.js`
  - `src\hooks\useSOTASpots.js`
  - `src\hooks\useWWBOTASpots.js`
  - `src\hooks\useWWFFSpots.js`

- components using `calculateGridSquare` from `geo.js`
  - `src\components\AzimuthalMap.jsx`
  - `src\components\DXFavorites.jsx`
  - `src\components\LocationPanel.jsx`
  - `src\components\SettingsPanel.jsx`
  - `src\components\WorldMap.jsx`
  - `src\hooks\app\useTimeState.js`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/src/utils/geo.js
+++ b/src/utils/geo.js
@@ -99,7 +99,7 @@ export const maidenheadToBoundingBox = (grid) => {
  * @param precision Precision (character length) of the grid locator returned, can be 2, 4, 6, or 8. (default if not specified is 6)
  * @returns Maidenhead grid locator
  */
-export const LatLonToMaidenhead = ({ lat, lon }, precision = 6) => {
+export const latLonToMaidenhead = ({ lat, lon }, precision = 6) => {
   if (lat < -90 || lat > 90) throw new Error('invalid latitude, it should be between -90 and 90');
   if (lon < -180 || lon > 180) throw new Error('invalid longitude, it should be between -180 and 180');
 
@@ -149,10 +149,10 @@ export const parseGridSquare = (grid) => {
 
 /**
  * Calculate Maidenhead grid square (size 6) from coordinates.
- * Deprecated, use LatLonToMaidenhead({lat, lon}) which defaults to size 6.
+ * Deprecated, use latLonToMaidenhead({lat, lon}) which defaults to size 6.
  */
 export const calculateGridSquare = (lat, lon) => {
-  return LatLonToMaidenhead({ lat, lon }, 6);
+  return latLonToMaidenhead({ lat, lon }, 6);
 };
 
 /**
@@ -553,7 +553,7 @@ export const classifyTwilight = (solarElevationDeg) => {
 export default {
   validateGridLocator,
   maidenheadToLatLon,
-  LatLonToMaidenhead,
+  latLonToMaidenhead,
   maidenheadToBoundingBox,
   parseGridSquare,
   calculateGridSquare,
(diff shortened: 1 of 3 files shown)


accius (approved):
Approving — useful extension of grid-locator support, with tests, and the legacy entry points still work.

- `validateGridLocator` regex `^[A-R]{2}([0-9]{2}([A-Xa-x]{2}([0-9]{2})?)?)?$` correctly enforces the size-2/4/6/8 progression with the right alphabet ranges (A–R for fields, A–X for sub-squares).
- `parseGridSquare` and `calculateGridSquare` are kept as deprecated thin wrappers — verified that the legacy 4-char and 6-char paths produce the same lat/lon as before (field+square+subsquare arithmetic matches).
- `maidenheadToLatLon` short-circuits on `!bbox` before reading `bbox.length`, so invalid input returns `null` rather than throwing — same contract as old `parseGridSquare`.
- Tests in `geo.test.js` cover both hemispheres (San Diego + Sydney), all four sizes, default-precision behaviour, and the legacy wrappers. CI green.
- Side note from the PR description (not blocking, but worth eventually closing): there are several files that import `maidenheadToLatLon` from server-side `callsign.js` vs the `@hamset/maidenhead-locator` package. Standardising those onto the new shared `geo.js` exports would be a nice follow-up — could file as a separate cleanup PR.

PR #938, opened 2026-05-01
[fix] Validate that `imageUrl` parses as `https://*.nasa.gov` before fetching.

MichaelWheeley:

## 🟡 Medium — slower bleed, still worth fixing

  - **`fetchDialAMoon` follows upstream-supplied URL** — `server/routes/space-weather.js:476-482`. Validate that `imageUrl` parses as `https://*.nasa.gov` before fetching.

### DIFF
--- a/server/routes/space-weather.js
+++ b/server/routes/space-weather.js
@@ -473,8 +473,26 @@ module.exports = function (app, ctx) {
       timestamp: Date.now(),
     };
 
+    const isValidNasaImageUrl = (url) => {
+      if (!url) return false;
+
+      let parsed;
+      try {
+        parsed = new URL(url);
+      } catch {
+        return false; // not a valid URL at all
+      }
+
+      if (parsed.protocol !== 'https:') return false;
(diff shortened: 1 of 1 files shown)


accius (approved):
Approving — small, surgical SSRF fix: validates the upstream-supplied `imageUrl` against `https://*.nasa.gov` before fetching.

- `URL` parse + protocol check + `hostname.endsWith('.nasa.gov')` is the right shape. Note: `endsWith('.nasa.gov')` (with the leading dot) blocks the bare apex `nasa.gov` and any malicious `evil-nasa.gov`-style host — good.
- Minor cleanup that's not blocking: `throw new Error('Rejected non‑NASA URL:', imageUrl)` — the `Error` constructor only reads its first argument, so the URL never makes it into the message. Suggest `throw new Error(\`Rejected non-NASA URL: ${imageUrl}\`)` (and ASCII hyphen instead of the U+2011 in 'non‑NASA') so the log line is actually useful when this fires.
- Server-side only, single file, CI green.

PR #941, opened 2026-05-02
[fix] MUF_MAP_CACHE periodic purge

MichaelWheeley:

  **6. `MUF_MAP_CACHE` has no eviction** — `server/routes/propagation.js:737`
  Its sibling `PROP_HEATMAP_CACHE` got a cleanup interval at line 605 — this one was missed. **Fix:** add the same `setInterval` purge + `MAX_ENTRIES` cap.

## details
- refactored `PROP_HEATMAP_CACHE` periodic purge to use `const maintainCache = (cache, ttlMs, maxEntries, label) => {}`
- added `MUF_MAP_CACHE` purge with same pattern
- changed `console.log` to `logDebug`

### DIFF
--- a/server/routes/propagation.js
+++ b/server/routes/propagation.js
@@ -622,7 +622,7 @@ module.exports = function (app, ctx) {
     }
 
     if (purged > 0) {
-      console.log(`[${label}] purged ${purged} stale entries, ${Object.keys(cache).length} remaining`);
+      logDebug(`[${label}] purged ${purged} stale entries, ${Object.keys(cache).length} remaining`);
     }
   };
 
--- a/server/routes/propagation.js
+++ b/server/routes/propagation.js
@@ -597,37 +597,43 @@ module.exports = function (app, ctx) {
     return { sfi, ssn, kIndex };
   }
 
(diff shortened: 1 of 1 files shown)


accius (approved):
Approving — clean refactor that adds the missing `MUF_MAP_CACHE` purge and folds `PROP_HEATMAP_CACHE` onto the same helper.

- Closing the audit gap for `MUF_MAP_CACHE` is the headline; sharing `maintainCache(...)` between the two propagation caches keeps the eviction logic in one place going forward.
- `MUF_MAP_TTL` is 5 min, purge interval is 10 min, eviction cutoff is `2 × TTL` = 10 min — entries can survive up to ~20 min worst case before sweep, which matches the existing PROP_HEATMAP behaviour. Fine.
- Side note: this defines a local `maintainCache` here, and #942 adds a near-identical one in `emcomm.js` (only the timestamp field name differs, `ts` vs `timestamp`). Worth lifting to `server/utils/cache.js` as a small follow-up so the two don't drift.
- Bonus cleanup: switching the leftover `console.log` to `logDebug` matches the project standard.

PR #942, opened 2026-05-02
[fix] add periodic purges to emcomm.js caches

MichaelWheeley:

  **5. Unbounded EmComm caches keyed on user lat/lon** — `server/routes/emcomm.js:11, 67, 137`                                                                                                                                                                               
  `alertsCache`, `sheltersCache`, `disastersCache` are plain objects with no eviction. Keys are `lat,lon` rounded to 1–2 decimals → up to ~64k unique keys per endpoint, each holding multi-KB GeoJSON. **Fix:** convert to `Map` + LRU + TTL purge — copy the pattern from  `propagation.js:605` or `callsign.js:30`. 

## details,
- added periodic purge of `alertsCache`, `sheltersCache`, and `disastersCache` caches
- subtle difference between `propagation.js` and `emcomm.js` is use of `ts' vs. `timestamp` for the .timestamp key in the `.map`.
- not determined if the maximum sizes of the various caches is sufficient
- noted that the disasters uses a state identifier, the alerts and shelters use lat/lon
- basic testing of each of them was successful

### DIFF
--- a/server/routes/emcomm.js
+++ b/server/routes/emcomm.js
@@ -6,10 +6,48 @@
 module.exports = function (app, ctx) {
   const { fetch, logDebug, logErrorOnce } = ctx;
 
+  const maintainCache = (cache, ttlMs, maxEntries, label = 'Cache') => {
+    const now = Date.now();
+    let purged = 0;
+
+    // Remove stale entries
+    for (const key of Object.keys(cache)) {
+      if (now - cache[key].timestamp > ttlMs * 2) {
+        delete cache[key];
+        purged++;
+      }
+    }
+
+    // Enforce max size by evicting oldest
+    const remaining = Object.keys(cache);
+    if (remaining.length > maxEntries) {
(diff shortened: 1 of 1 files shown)


accius (approved):
Approving — clean factoring of the per-cache cleanup pattern, all three caches (`alertsCache`, `sheltersCache`, `disastersCache`) now bounded.

- Good call extracting `maintainCache(cache, ttlMs, maxEntries, label)` — same shape as the version going into `propagation.js` in #941, just with `.timestamp` instead of `.ts`. Long term it might be worth lifting one shared helper into `server/utils/`, but that's a follow-up and not blocking.
- 200-entry hard cap × ~few-KB GeoJSON values puts the worst-case footprint in the low MB per cache — fine for a 2,000-user instance.
- Eviction sort touches every key but only on the 10-min tick, no hot-path cost.
- CI green, single file, only server-side state.

PR #943, opened 2026-05-02
fix(security): harden rig-bridge relay and status endpoints

ceotjoe:
### Security fixes

- **SSRF in \`/api/rig-bridge/status\`:** endpoint now validates the caller-supplied \`host\` via \`validateCustomHost\`, connects to the resolved IP (preventing DNS-rebinding TOCTOU), and is gated behind \`requireWriteAuth\`
- **Relay key leak:** \`RIG_BRIDGE_RELAY_KEY\` no longer leaves the server — \`relay/credentials\` and \`relay/configure\` now return a random 256-bit per-session token; \`requireRelayAuth\` validates against a server-side Map, so a leaked token is useless without a matching server-side entry
- **Long-poll DoS:** \`relay/commands\` now caps concurrent long-poll waiters at 10 per IP (returns 429 on excess)
- **URL injection in download scripts:** \`new URL()\` validation added after the regex sanitisation in the installer script builder
- **SSRF utility extracted:** \`isPrivateIP\` / \`validateCustomHost\` moved from \`dxcluster.js\` into new shared \`server/utils/ssrf.js\`, removing the duplicate

### Auth evolution: HMAC → server-side token store

Initial implementation used HMAC-SHA256(masterKey, sessionId) as a stateless token. In practice this broke whenever \`RIG_BRIDGE_RELAY_KEY\` differed between the issuing and validating process (restarts, env changes). Replaced with a server-side Map keyed by sessionId storing a random token issued at credential-setup time.

### Persistent relay credentials (survive server restarts)

Relay tokens are now written to \`data/relay-tokens.json\` (same path waterfall as \`settings.json\`: \`/data\` → \`ROOT_DIR/data\` → \`/tmp\`) on every issuance and loaded back at startup. rig-bridge authenticates once; credentials remain valid indefinitely across server restarts and deploys without any user action.

Tokens are **not** deleted on relay-session expiry — only when explicitly revoked or pruned (see below).

New endpoint: \`DELETE /api/rig-bridge/relay/revoke/:sessionId\` (behind \`requireWriteAuth\`) for manual credential invalidation.

### Stale token cleanup

Token entries now store \`{ token, lastUsed }\` instead of a bare string. \`lastUsed\` is updated in memory on every successful auth and flushed to \`relay-tokens.json\` hourly so the timestamp survives restarts.

On each server startup a background job fires 15 seconds after boot, removes any token whose \`lastUsed\` is older than 30 days, and persists the trimmed file — startup latency is unaffected. Old plain-string entries in existing \`relay-tokens.json\` files are migrated automatically (\`lastUsed\` set to now) so nothing is pruned unexpectedly on first upgrade.

### cloud-relay plugin fixes (rig-bridge 2.1.3)

- **Crash fix:** the three fire-and-forget \`http.request\` calls in \`executeCommand\` (setFreq / setMode / setPTT) had no \`.on('error', ...)\` handler. When rig-bridge TLS is enabled the loopback connection returns a TLS handshake instead of plain HTTP, triggering an unhandled \`'error'\` event that crashed the process. Error handlers added; protocol now selected from \`config.tls?.enabled\` with \`rejectUnauthorized: false\` for the self-signed loopback cert.
- **Fixed silent auth-error messages:** \`pushState\`'s \`makeRequest\` callback was declared as \`(err, status)\` but referenced a third \`data\` parameter that was never in scope. In strict mode this threw a ReferenceError caught by the catch block, so 401/403 responses always showed the hardcoded fallback instead of the server's error message. Fixed by adding \`data\` as the third callback parameter.
- **Session in push payload:** \`session\` field now included in the state-push POST body so the server can identify the session from the body when the header is absent.
- **connect() validation:** \`session\` is now checked alongside \`url\` and \`apiKey\` — missing session ID is caught at startup rather than failing silently on first push.

## ⚠️ Breaking change for existing Cloud Relay users

Any rig-bridge that was configured before this change has the raw \`RIG_BRIDGE_RELAY_KEY\` stored as its \`apiKey\`. After deploy the server will reject it with a 401 and the error message reads:

> *Invalid relay credentials — re-run Connect Cloud Relay in OHC Settings → Rig Bridge to generate fresh credentials*

**Recovery:** open OHC Settings → Rig Bridge → click **Connect Cloud Relay**, copy the new config payload into rig-bridge's Cloud Relay plugin, restart rig-bridge. After that, credentials persist across all future server restarts.

## Test plan

- [ ] Local rig-bridge status check (`/api/rig-bridge/status?host=localhost&port=5555`) still returns reachable state
- [ ] `relay/configure` returns a session-scoped random token (64-char hex), not the raw `RIG_BRIDGE_RELAY_KEY` value
- [ ] rig-bridge cloud-relay plugin authenticates successfully after re-running setup
- [ ] Stale credentials return HTTP 401 with the recovery hint message
- [ ] After server restart, rig-bridge reconnects without re-running setup (token loaded from `relay-tokens.json`)
- [ ] `relay/commands` long-poll returns 429 after 10 concurrent connections from the same IP
- [ ] Download installer scripts return 400 for a malformed `Host` header
- [ ] Clicking a DX cluster entry to tune no longer crashes rig-bridge when TLS is enabled
- [ ] `DELETE /api/rig-bridge/relay/revoke/:sessionId` invalidates the token and subsequent pushes receive 401
- [ ] Tokens unused for 30+ days are removed from `relay-tokens.json` after server restart (check log: `Pruned N stale relay token(s)`)
- [ ] Active tokens are not pruned; `lastUsed` is updated on each successful auth and written to disk within the hour
- [ ] Existing `relay-tokens.json` with old plain-string format is migrated without data loss

### DIFF
--- a/rig-bridge/package.json
+++ b/rig-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openhamclock-rig-bridge",
-  "version": "2.1.1",
+  "version": "2.1.2",
   "description": "Universal rig control bridge for OpenHamClock — plugin architecture supporting USB, flrig, rigctld, and more",
   "main": "rig-bridge.js",
   "bin": "rig-bridge.js",
--- a/rig-bridge/package.json
+++ b/rig-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openhamclock-rig-bridge",
-  "version": "2.1.2",
+  "version": "2.1.3",
   "description": "Universal rig control bridge for OpenHamClock — plugin architecture supporting USB, flrig, rigctld, and more",
   "main": "rig-bridge.js",
   "bin": "rig-bridge.js",
--- a/rig-bridge/plugins/cloud-relay.js
+++ b/rig-bridge/plugins/cloud-relay.js
@@ -144,7 +144,7 @@ const descriptor = {
         payload.aprsPackets = pendingAprs.splice(0, 50);
       }
 
-      makeRequest(`${serverUrl}/api/rig-bridge/relay/state`, 'POST', payload, (err, status) => {
+      makeRequest(`${serverUrl}/api/rig-bridge/relay/state`, 'POST', payload, (err, status, data) => {
         if (err) {
           if (serverReachable) console.error(`[CloudRelay] Push error: ${err.message}`);
           serverReachable = false;
--- a/rig-bridge/plugins/cloud-relay.js
+++ b/rig-bridge/plugins/cloud-relay.js
@@ -135,7 +135,7 @@ const descriptor = {
       lastState = { ...currentState };
 
       // Include batched data in the push
-      const payload = { ...currentState };
+      const payload = { ...currentState, session };
       if (hasDecodes) {
         payload.decodes = pendingDecodes.splice(0, 50);
         totalDecodes += payload.decodes.length;
@@ -162,7 +162,12 @@ const descriptor = {
             console.log(`[CloudRelay] Pushed state (${currentState.freq} Hz ${currentState.mode}${decodeInfo})`);
           }
         } else if (status === 401 || status === 403) {
-          console.error(`[CloudRelay] Authentication failed (${status}) — check relay API key`);
+          try {
+            const msg = JSON.parse(data)?.error || data;
+            console.error(`[CloudRelay] Authentication failed (${status}): ${msg}`);
+          } catch {
+            console.error(`[CloudRelay] Authentication failed (${status}) — check relay API key and session`);
+          }
         }
       });
     }
@@ -264,8 +269,8 @@ const descriptor = {
     }
 
     function connect() {
-      if (!serverUrl || !apiKey) {
-        console.error('[CloudRelay] Cannot start: url and apiKey are required');
+      if (!serverUrl || !apiKey || !session) {
+        console.error('[CloudRelay] Cannot start: url, apiKey, and session are required');
         return;
       }
 
--- a/rig-bridge/plugins/cloud-relay.js
+++ b/rig-bridge/plugins/cloud-relay.js
@@ -209,56 +209,50 @@ const descriptor = {
     }
 
     // Execute a command received from the cloud
+    // Uses https when rig-bridge TLS is enabled, http otherwise.
+    // All requests have an error handler to prevent unhandled-error crashes.
     function executeCommand(cmd) {
       totalCommands++;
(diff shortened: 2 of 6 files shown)


accius (approved):
Approving — substantial security hardening across the rig-bridge surface. The main concerns the audit raised (raw relay key in responses, SSRF on `/status`, long-poll DoS, URL injection in installer scripts) are all addressed, and the iteration history shows good debugging discipline (HMAC scheme → server-side token store after the cross-process key-mismatch was reproduced via the diagnostic logging).

A few things worth calling out — none blocking:

1. **Breaking change for existing Cloud Relay users.** Documented clearly in the PR body, and the 401 response now includes the recovery hint. Worth a WhatsNew entry next release noting that anyone currently using Connect Cloud Relay needs to reconnect once.

2. **`/api/rig-bridge/status` now requires write auth.** `RigContext.jsx:71` calls this from the frontend without a token. Locally that's fine because `requireWriteAuth` no-ops when `API_WRITE_KEY` isn't configured, but on the public site (where the key is set) the server-side health-check proxy will silently fail in the catch block. The end result is that public users get no `reachable`/`auth=enabled` diagnostic from the server — they just see whatever the direct CORS-bounded request returns. This is the tradeoff for fixing the SSRF, but worth verifying the user-visible degradation is acceptable, or threading the API token through the frontend call.

3. **Token store is unbounded.** `relayIssuedTokens` is a `Map` persisted to disk; the only prune is the 30-day-unused sweep at startup, plus explicit revoke. Realistically that's fine — each entry is small, churn is low, and `lastUsed` flushes hourly so the prune is meaningful — but if someone scripts repeated configure calls it would grow until the next restart. A periodic in-process sweep (not just startup) would close that.

4. **`saveRelayTokens` is synchronous** (`fs.writeFileSync` on every set/delete). Same pattern as `settings.json` so it's consistent with the rest of the codebase, just noting that under heavy churn this would block the event loop briefly. Not a real concern at current usage.

5. **`isPrivateIP` is exported but no longer called by either consumer** — `rig-bridge.js` only imports `validateCustomHost`, and `dxcluster.js` is reduced to the same. Keeping it exported for future reuse is fine.

6. Cloud-relay plugin fixes (`makeRequest` data param, TLS-aware `executeCommand` with error handlers, session in payload, connect() validation) — all clean and address real bugs flagged in the iteration. Version bump to 2.1.3 in both `package.json` and `rig-bridge.js`.

CI green; SSRF helper extracted to `server/utils/ssrf.js`; no functional regression spotted on the rig-bridge HTTP surface.

PR #945, opened 2026-05-03
[fix] lockout frequent POST to /api/presence

MichaelWheeley:

  **2. `/api/presence` — anyone can spoof any callsign** — `server/routes/presence.js:24-54`
  No auth, no per-IP limit. A POST with `{callsign:"K1JT", lat, lon}` pins that callsign on the Active Users layer for everyone. **Fix:** at minimum bind callsign↔IP, rate-limit to 1 update/60s/IP. 

## details,
- POST operation IP remote address `req.socket.remoteAddress` stored
- lookup prevents same remote address being used within 60s
- if detected that remote address is already associated with another callsign then that other callsign is deleted
- remote address purge occurs every ten mins, addresses purged after sixty minutes

image

### DIFF
--- a/server/routes/presence.js
+++ b/server/routes/presence.js
@@ -22,7 +22,7 @@ module.exports = function (app, ctx) {
 
   // map of POST operation IP remote addresses, with periodic cleanup
   const remoteAddresses = new Map();
-  const remoteAddress_TTL = 1 * 60 * 1000; // 1 minute
+  const remoteAddress_TTL = 60 * 60 * 1000; // 60 minutes
   setInterval(() => {
     const cutoff = Date.now() - remoteAddress_TTL;
     for (const [key, remoteAddress] of remoteAddresses)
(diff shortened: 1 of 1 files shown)


accius (approved):
Approving — closes the spoof vector by binding callsign↔IP and rate-limiting per-IP, which is what the audit asked for.

A few things worth tightening as follow-ups (none blocking):

- **Cleanup interval is 10 s, not 10 min.** The comment says `every ten mins` but `setInterval(..., 10000)` is 10 seconds. Functionally fine (light work, small Map), but please either bump to `10 * 60 * 1000` to match the comment or fix the comment. As written it'll run ~360× more often than intended.
- **`req.socket.remoteAddress` vs proxy.** OHC sets `app.set('trust proxy', TRUST_PROXY)` in `server/middleware/index.js`, so the canonical "client IP" everywhere else in the codebase is `req.ip`. Behind Railway/Cloudflare, `req.socket.remoteAddress` is the upstream proxy — meaning every spoofer sharing the same egress IP gets rate-limited as a single client (good for safety, bad if a legit shared NAT is ever in play). Switching to `req.ip` lines up with the rest of the routes.
- **Lockout response code.** A 429 (Too Many Requests) is a slightly more accurate signal to the client than 400 for the rate-limit case; not strictly required.

Logic itself looks right: same-IP-with-different-callsign purges the prior callsign from `activeUsers`, the entry is refreshed each time it's set, and the 60-min entry TTL is well clear of the 1-min lockout window. CI is green, single file.

PR #946, opened 2026-05-04
[fix] errorLogState periodic purge

MichaelWheeley:

  ## 🟡 Medium — slower bleed, still worth fixing                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                             
  - **`logErrorOnce` map never pruned** — `server/utils/logging.js:20-34`. Keys include `err.message`, which varies per host/port/URL → slow leak over weeks. Add a periodic prune older than `2 × ERROR_LOG_INTERVAL`.

## details,
- purges errorLogState with time cutoff `const cutoff = 2 * ERROR_LOG_INTERVAL;`
- repeats at `ERROR_LOG_INTERVAL`

### DIFF
--- a/server/utils/logging.js
+++ b/server/utils/logging.js
@@ -20,6 +20,19 @@ function logWarn(...args) {
 const errorLogState = {};
 const ERROR_LOG_INTERVAL = 5 * 60 * 1000;
 
+// Periodic cleanup: purge expired cache entries
+setInterval(() => {
+  const now = Date.now();
+  const cutoff = 2 * ERROR_LOG_INTERVAL;
+
+  for (const key of Object.keys(errorLogState)) {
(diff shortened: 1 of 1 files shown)


accius (approved):
Approving — straightforward, targeted fix for the audit finding.

- Cleanup interval matches the keep-window (`ERROR_LOG_INTERVAL`), and the cutoff is `2 × ERROR_LOG_INTERVAL` so an entry is purged ~5–10 min after its last refresh. That's well clear of the dedup window itself, so we won't accidentally re-emit messages we just suppressed.
- `delete` while iterating `Object.keys(errorLogState)` is safe because keys is a snapshot array at iteration start.
- Single file, server-side only, all CI green.

Issue #950, opened 2026-05-04
[fix / refactor] minor dialamoon NASA throw


MichaelWheeley:
...
- Minor cleanup that's not blocking: `throw new Error('Rejected non‑NASA URL:', imageUrl)` — the `Error` constructor only reads its first argument, so the URL never makes it into the message. Suggest `throw new Error(\`Rejected non-NASA URL: ${imageUrl}\`)` (and ASCII hyphen instead of the U+2011 in 'non‑NASA') so the log line is actually useful when this fires.
...

_Originally posted by @accius in https://github.com/accius/openhamclock/pull/938#pullrequestreview-4221591378_
            
please assign to @MichaelWheeley I'm not able to self-assign

PR #952, opened 2026-05-04
[fix] fixes issue#950 non-NASA imageUrl report in throw

MichaelWheeley:
## What does this PR do?

- fixes https://github.com/accius/openhamclock/issues/950
- single line change reports imageUrl as part of throw, previously missing

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/server/routes/space-weather.js
+++ b/server/routes/space-weather.js
@@ -491,7 +491,7 @@ module.exports = function (app, ctx) {
     const imageUrl = meta?.image?.url;
     if (!imageUrl) throw new Error('No image URL in Dial-A-Moon response');
     if (!isValidNasaImageUrl(imageUrl)) {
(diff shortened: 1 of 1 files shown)


accius (approved):
LGTM. The original `throw new Error('...:', imageUrl)` silently dropped `imageUrl` because `Error()` ignores arguments after the message. Template literal fixes this and the rejected URL will now actually appear in logs. Also corrects the non-breaking hyphen in the original string.

Minor nit (non-blocking): the `\'...\'` escaping inside the template literal isn't needed — `'${imageUrl}'` reads the same — but it's harmless.

Issue #949, opened 2026-05-04
[fix / refactor] lockout frequent POST to /api/presence#945


MichaelWheeley:
...
A few things worth tightening as follow-ups (none blocking):

- **Cleanup interval is 10 s, not 10 min.** The comment says `every ten mins` but `setInterval(..., 10000)` is 10 seconds. Functionally fine (light work, small Map), but please either bump to `10 * 60 * 1000` to match the comment or fix the comment. As written it'll run ~360× more often than intended.
- **`req.socket.remoteAddress` vs proxy.** OHC sets `app.set('trust proxy', TRUST_PROXY)` in `server/middleware/index.js`, so the canonical "client IP" everywhere else in the codebase is `req.ip`. Behind Railway/Cloudflare, `req.socket.remoteAddress` is the upstream proxy — meaning every spoofer sharing the same egress IP gets rate-limited as a single client (good for safety, bad if a legit shared NAT is ever in play). Switching to `req.ip` lines up with the rest of the routes.
- **Lockout response code.** A 429 (Too Many Requests) is a slightly more accurate signal to the client than 400 for the rate-limit case; not strictly required.

_Originally posted by @accius in https://github.com/accius/openhamclock/pull/945#pullrequestreview-4221612602_
            
please assign to @MichaelWheeley I'm not able to self-assign

PR #954, opened 2026-05-05
[fix/refactor] fixes issue#949, cleanup lockout frequent POST to /api/presence

MichaelWheeley:
## What does this PR do?

- fixes https://github.com/accius/openhamclock/issues/949
  - remoteAddresses cache cleanup interval set to 10min, was 10s.
  - remote client IP address change req.socket.remoteAddress to req.ip
  - change HTTP status code returned from 400 to 429 i.e. 'Too Many Requests' (but actual message returned 'IP Address lockout until timeout' not modified).

- tested using
  - `http://localhost:3001/api/presence`
  - `curl -X POST http://localhost:3001/api/presence -H "Content-Type: application/json" -d '{ "callsign": "K6MRW", "lat": 32.0, "lon": -117.0 }'`
  - `curl -X POST http://localhost:3001/api/presence -H "Content-Type: application/json" -d '{ "callsign": "zK6MRW", "lat": 32.0, "lon": -117.0 }'`

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

### DIFF
--- a/server/routes/presence.js
+++ b/server/routes/presence.js
@@ -42,7 +42,7 @@ module.exports = function (app, ctx) {
     let remoteAddressLockout = remoteAddresses.has(remoteAddress)
       ? Date.now() < remoteAddresses.get(remoteAddress).createTime + remoteAddress_lockout_period
       : false;
-    if (remoteAddressLockout) return res.status(400).json({ error: 'IP Address lockout until timeout' });
+    if (remoteAddressLockout) return res.status(429).json({ error: 'IP Address lockout until timeout' });
 
     if (!callsign || typeof callsign !== 'string' || callsign.length < 3 || callsign.length > 12) {
       return res.status(400).json({ error: 'Valid callsign required' });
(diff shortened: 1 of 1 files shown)


accius (approved):
LGTM. Three real fixes in a small diff:

1. **Cleanup interval bug** — was firing every 10s despite the comment claiming 10 minutes. Now correctly `10 * 60 * 1000`.
2. **429 vs 400** — rate-limit lockout is exactly what 429 Too Many Requests is for; 400 was misleading. Standard-aligned and clients/proxies can act on it correctly.
3. **`req.ip` over `req.socket.remoteAddress`** — `req.ip` honors the existing `app.set('trust proxy', TRUST_PROXY)` configuration in `server/middleware/index.js`, which defaults to `1` on Railway. So this gives the real client IP behind the proxy instead of the proxy's address — a strict improvement for the lockout map.

Note: `const cutoff = Date.now() - remoteAddress_TTL` inside the cleanup interval is unused (the loop checks `createTime + TTL` directly) — pre-existing, not introduced here, but easy to drop in a follow-up.

Issue #947, opened 2026-05-04
[refactor] unify use of .timestamp vs .ts so common cache maintenance util function can be used


MichaelWheeley:
`- Good call extracting `maintainCache(cache, ttlMs, maxEntries, label)` — same shape as the version going into `propagation.js` in #941, just with `.timestamp` instead of `.ts`. Long term it might be worth lifting one shared helper into `server/utils/`, but that's a follow-up and not blocking.`

_Originally posted by @accius in https://github.com/accius/openhamclock/pull/942#pullrequestreview-4221570351_

this is also mentioned here,
https://github.com/accius/openhamclock/pull/941#pullrequestreview-4221586640

PR #955, opened 2026-05-05
[refactor] Issue#947, create server/utils/cache.js with maintainCache

MichaelWheeley:
## What does this PR do?

closes https://github.com/accius/openhamclock/issues/947

- Created `server/utils/cache.js` with `maintainCache(...)` functionality. Removed existing definitions of same function.
- Load `maintainCache` in `server.js` and pass in `ctx` object.
- Validated that `emcomm.js` and `propagation.js` are able to call `maintainCache(...)` in `cache.js`
- So that all components follow a standard pattern, identified in `/server/**` any use of `.ts` as a field name in a cache and rename from `.ts` to `.timestamp`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

### DIFF
--- a/server.js
+++ b/server.js
@@ -98,6 +98,9 @@ const { endpointStats, writeLimiter, requireWriteAuth } = applyMiddleware(app, {
   CORS_ORIGINS,
 });
 
+// ── Utilities ──
+const { maintainCache } = require('./server/utils/cache');
+
 // ── Build shared context object ──
 const ctx = {
   // Core
@@ -144,6 +147,7 @@ const ctx = {
   requireWriteAuth,
   writeLimiter,
   endpointStats,
+  maintainCache,
 };
 
 // ── Visitor stats service ──
--- a/server/routes/dxpeditions.js
+++ b/server/routes/dxpeditions.js
@@ -19,11 +19,11 @@ module.exports = function (app, ctx) {
   const sourceCaches = new Map();
 
   async function cachedFetch(name, fetcher) {
-    const entry = sourceCaches.get(name) || { data: null, ts: 0 };
-    if (entry.data && Date.now() - entry.ts < SOURCE_TTL) return entry.data;
+    const entry = sourceCaches.get(name) || { data: null, timestamp: 0 };
+    if (entry.data && Date.now() - entry.timestamp < SOURCE_TTL) return entry.data;
     try {
       const fresh = await fetcher();
-      sourceCaches.set(name, { data: fresh, ts: Date.now() });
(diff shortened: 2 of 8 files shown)


accius (changes requested):
Refactor direction is good — pulling `maintainCache` into a single `server/utils/cache.js` and standardizing `.timestamp` is the right cleanup. One blocker before merge:

### Blocker: `logDebug` is undefined in `server/utils/cache.js`

```js
// server/utils/cache.js
const maintainCache = (cache, ttlMs, maxEntries, label = 'Cache') => {
  ...
  if (purged > 0) {
    logDebug(`[${label}] purged ${purged} stale entries, ...`);  // ReferenceError
  }
};
```

In the original copies inside `emcomm.js` and `propagation.js`, `logDebug` was captured from the enclosing `ctx` closure. The extracted module has no such closure, and there's no `require`/import of `logDebug` here. As soon as a cache eviction triggers (`purged > 0`) this will throw `ReferenceError: logDebug is not defined` and crash the cache maintenance call.

Suggested fix — accept a logger as an argument so the util stays decoupled:

```js
const maintainCache = (cache, ttlMs, maxEntries, label = 'Cache', logDebug = () => {}) => {
  ...
};
```

…and update callers to pass `ctx.logDebug` (or just drop the log line if nobody cares about purge counts).

### Non-blocking notes

- **`wsjtx.js` line 819**: changing `q.ts_utc || q.ts` to `q.ts_utc || q.timestamp` is a semantic fallback change for QSO records, not just a cache field rename. In this codebase QSOs only seem to set `ts_utc` (line 858), so the fallback is practically dead either way — but worth being explicit about: there's no other place in `wsjtx.js` writing `.ts` *or* `.timestamp` on a QSO, so this branch will only ever match if a QSO comes in from outside with that field. Probably fine, but flagging.
- **`dxNewsMerge.test.js`**: renaming the local `ts` map parameter to `timestamp` is purely cosmetic. Slightly out of scope for "rename the cache field", but harmless.

Once the `logDebug` issue is fixed I'm happy to re-review.

MichaelWheeley:
@accius removed the logDebug from the utils/cache.js so should be good to go...

accius (approved):
Blocker resolved — `logDebug` is no longer called from the extracted util, so no more `ReferenceError` risk. Dropping the purge-count log line is fine; it was debug-level noise that was rarely useful in practice anyway. The unused `label` parameter in the signature is harmless and keeps callers stable.

Approving and merging.

PR #959, opened 2026-05-07
fix(update): use npm ci to prevent stale node_modules after update

ceotjoe:
## Summary

- Replaces `npm install --include=dev` with `npm ci` in `scripts/update.sh`
- `npm ci` wipes `node_modules` before installing, so newly added production dependencies (like `rss-parser` in v26.3.x) are guaranteed to be present after an update
- Fixes a user-reported issue where updating via `update.sh` left `rss-parser` absent, causing `[FATAL] Cannot find module 'rss-parser'` on service restart; re-running `setup-pi.sh` (which writes into a clean tree) resolved it

## Test plan

- [ ] Run `update.sh` on a Pi with an existing install from before `rss-parser` was added and confirm the service starts cleanly
- [ ] Confirm `npm ci` fails fast if `package.json` and `package-lock.json` are out of sync (expected safety net behaviour)


ceotjoe:
Will be newly created.

PR #970, opened 2026-05-10
[Translation] [feature/satellite] Language translations for all satellite features

MichaelWheeley:
## What does this PR do?

- new keys for satellite feature so that everything is translated
- all languages updated, thanks AI

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [X] Translation
- [ ] Map layer plugin

## Checklist

- [X] App loads without console errors
- [X] Tested in **Dark**, **Light**, and **Retro** themes
- [ ] Responsive at different screen sizes (desktop + mobile)
- [ ] If touching `server.js`: caches have TTLs and size caps (we serve 2,000+ concurrent users)
- [ ] If adding an API route: includes caching and error handling
- [ ] If adding a panel: wired into Modern, Classic, and Dockable layouts
- [X] No hardcoded colors — uses CSS variables (`var(--accent-cyan)`, etc.)
- [X] No `.bak`, `.old`, `console.log` debug lines, or test scripts included

## Screenshots (if visual change)
image
image
image
image

PR #971, opened 2026-05-10
Workflow change to check src/lang/*.json key order, additional tools provided

MichaelWheeley:
## What does this PR do?

- Enforce `src/lang/*.json` key order in workflow.
- Adds tools
  - `npm run lang:check`
  - `npm run lang:sort`
- (The sort order in the js and mjs are the same order that VSCode natively uses with its built-in JSON sorter).
  - i.e. vscode.json-language-features
image

## Type of change
workflow
- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. edit a language JSON file such as `src/lang/en.json` and switch a few lines around
2. PR will fail merge to Staging or Main
3. use `npm run lang:sort`, and push changes

when fail occurs guidance is provided,
image


MichaelWheeley:
by design this PR fails checks as originally written, necessary to sort language JSON keys to get it to pass

image

accius (approved):
LGTM. Nice tooling addition — the sort scripts are deterministic, the CI gate makes the standard self-enforcing, and the matching VSCode sort behavior means contributors can use the IDE's built-in sorter. Will need PR #970 to rebase + re-sort after this lands.

PR #981, opened 2026-05-12
[BUG 976] Make useLocalInstall() more complete

### LINKED ISSUE #976: [BUG] useLocalInstall() does not recognize common homelab/internal DNS domains such as .home
### Summary

The `useLocalInstall()` hook currently only treats the following hostnames as “local installs”:

```
* localhost
* 127.0.0.1
* ::1
* *.local
* RFC1918 private IP ranges (192.168.x.x, 10.x.x.x, 172.16-31.x.x)
```

This prevents self-hosted OpenHamClock instances accessed through common internal DNS/reverse proxy setups from being recognized as local installs.

As a result, features gated behind isLocalInstall (such as UDP Spots) are incorrectly disabled even though the application is running entirely on the local LAN.

Current code:
```
if (host === 'localhost' || host === '127.0.0.1' || host === '::1') return true;
if (host.endsWith('.local')) return true;
if (host.startsWith('10.') || host.startsWith('192.168.')) return true;
```

### Problem

Many modern homelab/self-hosted environments use internal DNS domains such as:

* .home
* .lan
* .internal
* custom split-horizon DNS domains

For example:
`https://hamclock.home`
is a completely local installation behind:

* local DNS
* local reverse proxy
* local TLS certificates
* LAN-only routing

However, OpenHamClock reports "Hosted mode" and disables local-only functionality (like UDP Spots as the DX Cluster source)

### Additional Issue with .local

On macOS/iOS, .local is reserved for Bonjour/mDNS and is often unreliable or inappropriate for manually-defined DNS records.

Many homelab users intentionally avoid .local for this reason and instead use:

* .home
* .lan
* custom internal domains

So limiting local detection to .local creates avoidable interoperability issues.

### Suggested Fix

At minimum, support additional common internal domains:
```
if (
  host.endsWith('.local') ||
  host.endsWith('.home') ||
  host.endsWith('.lan')
) return true;
```

### Better Long-Term Approach

Instead of hardcoding domain suffixes, consider determining “local install” based on:

* whether the backend is self-hosted
* whether the frontend/backend origins match
* whether the server is reachable on RFC1918/private addressing
* an explicit environment variable/config option

### Environment

* OpenHamClock latest release (26.3.2)
* Ubuntu 24.04
* Caddy reverse proxy
* AdGuard Home local DNS
* macOS clients

alanhargreaves:
## What does this PR do?

- Makes `useLocalInstall()` check a few more domain names
- Tidies up `useLocalInstall()`
- Removes extra instances of isLocalInstall (passing the one copy around) and reduces the calls to useLocalInstall to the single call in `App`
- Adds a `.env` variable called localServer, which defaults to false if not set. If this is set to `true` it will override all other checks in `useLocalInstall() in the client.
- Logs an informational message if we are overriding the local testing

## Suggested Release Notes Text

The checks as to whether or not a local server is being used, have been made more thorough. In addition to comparing the client url  against ".local", openhamclock now compares against ".home", ".lan" and ".internal" as well as the existing checking for IP addresses in the RFC-1918 private IP address ranges. A new override in .env has also been provided. If SERVERLOCAL is set to true, then the check for using a local server will always be true without the overhead of the extra checks. If set to false (or not set), the extra checking will take place.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Set `SERVERLOCAL=true` in `.env` and restart server
2. Watch informational messages in the client and verify that we see
   ```
   [useLocalInstall] Bypassing extra testing as serverLocal is set true
   ```
   in the informational messages.
3. Remove the SERVERLOCAL entry (or comment it or set it to false) and restart server
4. Verify that we do NOT see the above informational message

The only other way to verify the extra checks that have been added would be to be running locally in such a way that we previously were not.

## Checklist


accius (commented):
Good direction overall. Adding the `.home` / `.lan` / `.internal` host patterns expands coverage cleanly, and hoisting the `useLocalInstall()` call up to `App` and passing the boolean down via props eliminates the duplicate hook calls in `SettingsPanel`/`WorldMap`. The `serverLocal` env-driven override is a useful escape hatch for unusual deployments (Tailscale, reverse-proxied LAN hosts on public-looking domains, etc.).

Two things before this is ready to come out of draft:

1. The "How to test" section in the PR description is still TODO — please fill that in so reviewers (and future you) can validate the expanded host pattern matrix and the `serverLocal` override path.
2. The frontend reads `serverCfg.serverLocal` in `useAppConfig.js`, but I don't see the corresponding server-side change that surfaces that field through `/api/config` (or wherever `getServerConfig()` resolves). If that lives in another commit on the branch, great — otherwise it'll need to be added so the override actually flows through end-to-end.

Holding the merge until it's out of draft.

alanhargreaves:
OK, it seems to be working correctly on my local instance.

MichaelWheeley:
may need to rebase branch against 425662f to pass CI/format check

accius (dismissed):
A couple things to sort out before this lands.

1. CI is red because of `iturhfprop-service/README.md`, but that's not from your changes. Staging picked up a Prettier fix in commit 1dfbe0d after you branched, so your tree still has the old version of that file. Rebase or merge Staging into the branch and the format check should clear.

2. Actual code issue on the server-side default. In `server/config.js`:

```js
serverLocal: process.env.SERVERLOCAL !== false && jsonConfig.serverLocal !== false,
```

`process.env.SERVERLOCAL` is always either a string (`'true'` or `'false'`) or `undefined`. It can never equal the boolean `false`, so `process.env.SERVERLOCAL !== false` is always true. End result is `serverLocal` is `true` whenever `jsonConfig.serverLocal` isn't explicitly the boolean false, which is the opposite of what the PR description and the `.env.example` comment say (should default to false if not set).

I think the intent is something like:

```js
serverLocal: process.env.SERVERLOCAL === 'true' || jsonConfig.serverLocal === true,
```

Or normalize the env var to a boolean somewhere upstream and then compare cleanly. Either way the current expression isn't going to behave the way the docs claim.

Rest of the changes look clean. The `.home` / `.lan` / `.internal` additions are good, and consolidating to a single `useLocalInstall()` call passed down via props instead of multiple component-local instances is the right cleanup.

accius:
Still one issue with this. The `!== 'false'` check passes when the env var is just plain unset, not only when it's set to "false". So if you don't put a SERVERLOCAL line in your .env at all (which is what most fresh installs look like), it ends up `true` and the local-install override kicks in for everyone.

Easier to read it as "only true if you explicitly say so":

```js
serverLocal: process.env.SERVERLOCAL === 'true' || jsonConfig.serverLocal === true,
```

That way unset, "false", or anything weird all stay off.

alanhargreaves:
Doh! yes I want the default to be false. Fixed.

PR #982, opened 2026-05-12
[Fix] Maidenhead grid special case on anti-meridian. Add meridian & anti-meridian unit test cases.

MichaelWheeley:
## What does this PR do?

- This is NOT an urgent PR, it is a fix for a special case identified.
- Strictly `{lat:0, lon:180}` is not valid since it should correctly be written `{lat:0, lon:-180}`, these are both points on the anti-meridian with locator `AJ00aa`.
- Code accepts `{lat:0, lon:180}` since it might sometimes be expected to be functional. Problem was identified that `{lat:0, lon:180}` was calculating `SJ00aa` which is an invalid locator.
- Unit test cases added on meridian and anti-meridian.
- Remove box-coordinate test from Sydney, one test San Diego is sufficient for coverage since the box-coordinate function is also used internally.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/src/utils/geo.js
+++ b/src/utils/geo.js
@@ -104,7 +104,10 @@ export const latLonToMaidenhead = ({ lat, lon }, precision = 6) => {
   if (lon < -180 || lon > 180) throw new Error('invalid longitude, it should be between -180 and 180');
 
   const latNorm = lat + 90;
-  const lonNorm = lon + 180;
+
+  // Handle case where longitude is given as +180, which should be treated as -180,
+  // by normalizing to 0-360 range first then applying modulus again to get back to -180..180 range.
+  const lonNorm = (((lon + 180) % 360) + 360) % 360;
 
   // Field (2 chars): 20° lon x 10° lat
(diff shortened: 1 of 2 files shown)


accius (commented):
Nice fix. The `(((lon + 180) % 360) + 360) % 360` normalization handles both `+180` (often accepted as input) and `-180` (canonical) → field char `'A'`, eliminating the invalid `'S'` that the old `lon + 180` math produced at the antimeridian.

Test coverage is solid: prime meridian, just-west-of-antimeridian, and both `±180` cases. Making the bounding-box test conditional via `if (latLonSWCornerGrid6 && latLonNECornerGrid6)` is a clean way to let the antimeridian cases skip that assertion (where the box wraps and the simple `toBeCloseTo` check wouldn't work cleanly anyway).

Merging.

PR #983, opened 2026-05-13
[server utils] Add utils normalize, mutex, statemachine

MichaelWheeley:
## What does this PR do?

- add server utils
  - normalize = string to numeric conversion, JSON string to numeric conversion
  - mutex = locking/unlocking, safe value and counters for asynchronous operation
  - statemachine = basic state machine with defined named states and handlers
- includes unit tests for each component

### utils
- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/server/utils/mutex.js
+++ b/server/utils/mutex.js
@@ -1,0 +1,90 @@
+/**
+ * A simple mutex implementation for Node.js to manage concurrent access to shared resources.
+ * The Mutex class provides basic locking and unlocking functionality, while MutexValue and MutexCounter
+ * extend this functionality to manage a value and a counter, respectively.
+ *
+ * @class Mutex
+ * @typedef {Mutex}
+ */
+class Mutex {
+  constructor() {
+    this.locked = false;
+    this.queue = [];
+  }
+
+  async lock() {
+    return new Promise((resolve) => {
+      if (!this.locked) {
+        this.locked = true;
+        resolve();
+      } else {
+        this.queue.push(resolve);
+      }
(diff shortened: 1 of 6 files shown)


accius (commented):
Three small, well-scoped utilities with tests. CI green. A few non-blocker notes for follow-up:

1. `normalizeJsonTree` mutates its input in place. Probably fine for server-side use, but consider returning a new tree if you ever want to share it with code that holds onto the original.
2. In `statemachine.test.js` there's a bug in the "should throw if invalid state index is provided" test: `new StateMachine(['A'], { A: () => 'B' }); sm.run();` — `sm` is never assigned (the `new StateMachine(...)` result is thrown away). The assertion passes only because accessing `.run()` on the undefined `sm` throws a ReferenceError, which satisfies `.toThrow()` but not for the intended reason. Worth a follow-up to fix.
3. `statemachine.js` does `require('../utils/mutex')` from inside `server/utils/` — works but the relative path is awkward; `./mutex` would be clearer.
4. PR body notes "not used yet, coming soon" — please make sure the integration PR lands close behind, otherwise this is dead code on main.

Merging.

MichaelWheeley:
> 4. PR body notes "not used yet, coming soon" — please make sure the integration PR lands close behind, otherwise this is dead code on main.

- coming in satellite.js changes soon

PR #985, opened 2026-05-14
Move the QRZ.COM credentials settings to the Integrations Page

### LINKED ISSUE #984: [BUG] Move the QRZ Credentials setting to the Integrations page
## Describe the Bug

In spite of a clear notice that this is a setting only for folks running local servers people are still asking about how to set it.

I propose moving it to the Integrations settings panel and have it part of the Local-only settings and have it **only** settable if `isLocalInstall` is true.

## Expected Behaviour

Should only be able to set the QRZ credentials is `isLocalInstall` is true.

## Environment

All

---

73!

alanhargreaves:
## What does this PR do?

In spite of a clear notice that this is a setting only for folks running local servers people are still asking about how to set it.

This PR moves it to the Integrations settings panel and have it part of the Local-only settings and have it only settable if isLocalInstall is true.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Run up with a local server
   a. Verify the QRZ.COM Credentials settings do not appear on the Profiles Settings panel
   b. Verify the QRZ.COM Credentials settings appear on the integrations page, the fields can be filled in and on completion of both text fields, the button becomes enabled.
2. Fudge isLocalServer to be false in SettingsPanel and restart server
   a. Verify the QRZ.COM Credentials settings do not appear on the Profiles Settings panel
   b. Verify that the QRZ.COm credentials settings cannot be edited

## Checklist

### DIFF
--- a/src/components/SettingsPanel.jsx
+++ b/src/components/SettingsPanel.jsx
@@ -2738,6 +2738,79 @@ export const SettingsPanel = ({
                     />
                   </div>
                 </div>
+
+                <details style={{ marginTop: 10 }}>
+                  <summary
+                    style={{
+                      cursor: 'pointer',
+                      color: 'var(--accent-amber)',
+                      fontSize: 12,
+                      userSelect: 'none',
+                    }}
+                  >
+                    Learn how
+                  </summary>
+
+                  <div
+                    style={{
+                      marginTop: 8,
+                      color: 'var(--text-secondary)',
+                      fontSize: 12,
+                      lineHeight: 1.55,
+                    }}
+                  >
+                    <div style={{ marginBottom: 6 }}>
+                      <b>Quick start (Local Only):</b>
+                    </div>
+
+                    <ol style={{ margin: 0, paddingLeft: 18 }}>
+                      <li>
+                        Run OpenHamClock locally:
+                        <div style={{ fontFamily: 'var(--font-mono)', marginTop: 4 }}>npm start</div>
+                        Open the local URL shown in your terminal (example: http://127.0.0.1:3001).
+                      </li>
+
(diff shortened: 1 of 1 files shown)


MichaelWheeley:
@alanhargreaves 
> 1. Run up with a local server
>  a. Verify the QRZ.COM Credentials settings do not appear on the Profiles Settings panel
>   b. Verify the QRZ.COM Credentials settings appear on the integrations page, the fields can be filled in and on completion > of both text fields, the button becomes enabled.

verified

> 2. Fudge isLocalServer to be false in SettingsPanel and restart server

- think you mean **`isLocalInstallation`** rather than `isLocalServer`,
- was able to fudge by returning `true` or `false` at line 7 of `src\hooks\app\useLocalInstall.js`

>   a. Verify the QRZ.COM Credentials settings do not appear on the Profiles Settings panel
>   b. Verify that the QRZ.COm credentials settings cannot be edited

verified

accius (commented):
Good cleanup — addresses real user confusion about a setting that doesn't apply on the hosted site. The Save & Test button and both input fields now gate on `isLocalInstall`, which is exactly right, and the styling matches the surrounding Integrations sections cleanly.

One small nit: the Remove button (rendered when `qrzStatus?.configured && qrzStatus.source !== 'env'`) is not gated on `isLocalInstall`, so on a non-local install where credentials were somehow already configured the Remove button is still active. Edge case (you can't reach the configure flow on hosted in the first place) but worth tightening if you spin up another commit.

Otherwise looks good, merging.

PR #1006, opened 2026-05-19
Newutils cleanup

MichaelWheeley:
## What does this PR do?

- per https://github.com/accius/openhamclock/pull/983#pullrequestreview-4315166311
- fix require path
- state machine unit test, state should gracefully reset if invalid state provided, it doesn't throw in this condition (throws occur only in the sm constructor, during the run() errors are reported but handled)
- (did not modify json normalization mutation)

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

### DIFF
--- a/server/utils/statemachine.js
+++ b/server/utils/statemachine.js
@@ -1,4 +1,4 @@
-const { Mutex, MutexValue, MutexCounter } = require('../utils/mutex');
+const { Mutex, MutexValue, MutexCounter } = require('./mutex');
 
 /**
  * A simple state machine implementation with safety checks to prevent invalid states and handlers.
--- a/server/utils/statemachine.test.js
+++ b/server/utils/statemachine.test.js
@@ -44,11 +44,10 @@ describe('StateMachine with invalid inputs', () => {
     expect(() => new StateMachine(['A'], {})).toThrow();
   });
 
-  it('should throw if invalid state index is provided', () => {
-    expect(() => {
-      new StateMachine(['A'], { A: () => 'B' });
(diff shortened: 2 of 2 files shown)


accius (commented):
Reviewed the changes. Small, well-targeted cleanup and it looks good.

**statemachine.js** — the require path fix from `../utils/mutex` to `./mutex` is correct; both files live in `server/utils/` so the redundant directory hop wasn't wrong, just confusing. Cleaner this way.

**statemachine.test.js** — good catch on the old `should throw if invalid state index is provided` test. It referenced a `sm` variable that was never assigned, so it was only "passing" because the undefined reference threw a ReferenceError — not because the state machine actually threw. Since SAFETY 3 in `run()` only logs and resets on an invalid next state (it never throws outside the constructor), rewriting it as `should gracefully reset if invalid state index is provided` and asserting the reset-to-'A' behavior matches what the code actually does.

Dropping `should reset if next state is invalid` is fine since the rewritten test already exercises the invalid-next-state path; no coverage lost.

One minor note, not blocking: the rewritten test is marked `async` and awaits `sm.currentState()`, but `sm.run()` itself is not awaited. It works here because the handler is synchronous, but awaiting `run()` would be more consistent with the other tests in the file and more robust if the handler ever becomes async.

CI is green on both Node 20.x and 22.x, and tests pass locally. Approving in spirit — clean fix.

— K0CJH

Issue #987, opened 2026-05-15
[FEATURE]  Include the Satellite status box in the Hide UI option


navyitc:
##   Include the Satellite status box in the Hide UI option

The Satellite Status box can be minimized but still floats on the Map. You can move it around but not hide it.

## Describe Alternatives You've Considered - sighing in frustration - I dislike seeing the floating box

A clear and concise description of any alternative solutions or features you've considered. Turning off Satellite views - not preferred.

## Use Case

How would this feature benefit amateur radio operators? Be specific about the use case:

- [ ] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [ ] Contesting
- [ ] POTA/SOTA
- [ ] DXing
- [ ] Emergency communications
- [ X ] Other: Clean, uncluttered map display

## Would you be willing to help implement this?

- [ ] Yes, I can submit a PR
- [ X ] Yes, I can help test
- [ ] No, but I'd use it!

Thanks for your hard work and an amazing tool for the shack - visitors are amazed and astounded by its awesomeness - as am I! I use it all day every day to extract useful data and operating conditions status.
---

73, WZ4K!

accius:
Done — on Staging in 6858893. The Hide UI toggle now also hides the satellite status box, the same way it hides the other floating map panels. Thanks for the report and the kind words, WZ4K. 73 — K0CJH

PR #1008, opened 2026-05-20
(fix issue 987) Shrink satellites dialog to minimal size

MichaelWheeley:
## What does this PR do?

- fixes https://github.com/accius/openhamclock/issues/987
- shrink satellites dialog to minimal size
- eliminate text when minimized so only satellite icon and right-arrow appear
- doesn't eliminate satellites button altogether like other boxes in the 'Show UI' pattern, but makes satellite less obtrusive

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist

before
image

after
image
image

### DIFF
--- a/server/routes/satellites-tracked.js
+++ b/server/routes/satellites-tracked.js
@@ -43,11 +43,17 @@
 // removed TEVEL satellites, added TEVEL2 satellites
 // remove (deactivated) GOES-13
 // remove (decayed) SO-124, CAS-4A, CAS-4B, EO-88, XW-2A, XW-2B, XW-2C, XW-2F
+//
+// May 18, 2026
+// added #41866 GOES-16 celestrak_weather
+// added #55506 ELEKTRO-L4 celestrak_active
+// added #67756 ELEKTRO-L5 celestrak_active
+// removed #53106(payload) = #53109(rocket body) = IO-117(Greencube), satellite decommissioned
 
 const HAM_SATELLITES = {
   // ── High Priority — Popular FM Satellites ──────────────────────
   ISS: {
-    // CelesTrak group: amateur
+    data_source: 'celestrak_amateur',
     norad: 25544,
     name: 'ISS (ZARYA)',
     color: '#00ffff',
@@ -58,7 +64,7 @@ const HAM_SATELLITES = {
     tone: '67.0 Hz',
   },
   'SO-50': {
-    // CelesTrak group: amateur
+    data_source: 'celestrak_amateur',
     norad: 27607,
     name: 'SO-50',
     color: '#00ff00',
@@ -70,7 +76,7 @@ const HAM_SATELLITES = {
     armTone: '74.4 Hz',
   },
   'AO-91': {
-    // CelesTrak group: amateur
+    data_source: 'celestrak_amateur',
     norad: 43017,
     name: 'AO-91 (Fox-1B)',
     color: '#ff6600',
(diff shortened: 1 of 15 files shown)


accius (commented):
Thanks for tackling #987, Michael. The minimized state looks much cleaner in the screenshots and it's nice to keep the satellite control discoverable rather than hiding it entirely.

One correctness issue I'd like resolved before merge:

**Inline sizing styles are not reset on expand.** In the minimized branch you now set several styles directly on the persistent `win` element:

```js
win.style.width = 'fit-content';
win.style.minWidth = 'unset';
win.style.maxWidth = 'fit-content';
win.style.height = 'fit-content';
win.style.minHeight = 'unset';
```

The window element is created once and reused (`if (!win) { ... }`). The expanded path only resets `maxHeight` and `overflowY` (lines 251-252) — it never restores `width`/`minWidth`/`maxWidth`/`height`/`minHeight`. So after a minimize → expand cycle the window stays stuck at `fit-content` width instead of returning to the original `260px`. Please add the matching resets in the expanded branch, e.g.:

```js
win.style.width = '260px';
win.style.minWidth = '';
win.style.maxWidth = '';
win.style.height = '';
win.style.minHeight = '';
```

Minor notes:
- The `titleBar` template now mixes 2-space and original indentation, and the closing backtick is unindented. Worth running the formatter so the diff stays tidy.
- The conditional `padding`/`width`/`flex` block inside the title bar's inline `style` works, but the same minimized sizing is also being applied imperatively to `win` below — consider keeping all the minimized sizing in one place to avoid drift.
- Checklist shows Dark/Light/Retro theme testing and responsive testing unchecked. Since this is purely a visual change, please confirm the minimized pill renders correctly in all three themes.

Logic is otherwise straightforward and CI is green. Happy to re-review once the expand-reset is added.

— K0CJH

MichaelWheeley:
@accius 
- updated per your review comments
- rebased to latest staging
- respectfully, reverted your recent change 6858893 which hides the satellite dialog completely. The intent of this PR is to have the ability to minimize the satellite window but NOT to have satellites disappear as clutter. Suggest the real way to achieve HideUI maybe be to have a pinnable window model?

accius (commented):
Re-reviewed after the latest commits — the blocking item from my last review is resolved.

**Resolved:** the expanded branch now restores `width` / `minWidth` / `maxWidth` / `height` / `minHeight` (the "reset to default size constraints" block), so a minimize → expand cycle returns the window to `260px` instead of staying stuck at `fit-content`. Verified against the diff.

The real change against current Staging is small and clean — two files:

- `useSatelliteLayer.js` — minimize-to-icon, plus the sizing reset.
- `WorldMap.jsx` — drops `.sat-data-window` from the `mapUiHidden` selector list, reverting commit `6858893` so the satellite window *minimizes* rather than disappearing entirely with the Hide-UI toggle.

That second change is a deliberate reinterpretation of #987 (minimize to an icon, vs. hide completely). I agree minimize is the better reading of the issue — just calling it out so it's a conscious decision rather than a silent revert.

Two non-blocking notes:

1. **The GitHub diff view is misleading here.** It shows 14 files / ~1,200 lines because the branch merged Staging into itself (commit `1afe6cc`) and carries a duplicated "shrink satellites dialog" commit. The actual delta applied to Staging is only the two files above — `git diff Staging...head` is 39 lines. Please **squash-merge** this so the self-merge and the duplicate commit don't land on Staging.
2. Minimized sizing is still set in two places — conditionally in the `titleBar` inline style and imperatively on `win`. It works; it's the drift risk I flagged last time. Worth consolidating eventually, not now.

CI is green and it merges cleanly. Engineering side is good — squash on merge.

— K0CJH

MichaelWheeley:
see https://github.com/accius/openhamclock/pull/1010

PR #1014, opened 2026-05-24
docs(docker): improve container documentation

lbatalha:
## What does this PR do?

- Add documentation to clarify that there are container images available in the github container registry, as it is not currently obvious
- Improve docker compose to use that image
- Document ITURHFPROP microservice as part of a compose file

I think it would be nice if the ITURHFPROP service also had a github action to build a container image similar to OHC, to simplify new deployments.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [x] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

To test the compose file change simply `docker compose pull` and verify that the image is pulled from ghcr.io


accius (commented):
Thanks for picking this up — the GHCR images aren't discoverable today and the iturhfprop microservice compose example is a nice add. The `docker-compose` → `docker compose` (V2) update is overdue. Verified the GHCR image is real: `docker-image.yml` publishes to `ghcr.io/accius/openhamclock` on every tagged release.

**One blocker, one substantive note, a couple of small things.**

**Blocker — wrong base branch.** This PR targets `main`, but all in-flight work for this repo merges to `Staging` first (`main` only takes a release at a time). External docs PRs are no exception. Could you retarget the PR to `Staging`? (Easiest from the GitHub UI under "Edit" on the PR title; no need to recreate.)

**docker-compose.yml has both `build: .` and `image: ...`.** With both present, `docker compose up` still *builds* by default and tags the build as the image name — it doesn't pull the prebuilt image automatically. So the README change implies "use the prebuilt image" but the compose file's default behavior is unchanged. Two cleaner options:

- For the "pull the prebuilt image" UX: remove `build: .` from the default compose file and document `docker compose build` for users who want to build from source.
- Or split into two files — `docker-compose.yml` (pulls `image:`) and `docker-compose.build.yml` (overrides with `build:`).

As written, anyone following the new docs will still do a local build on first `up`, which defeats the point of pointing them at GHCR.

**Smaller notes:**

- README reorders Quick Start to put "Container Deployment" before "Local Install". Fine editorially, just confirming it's deliberate.
- iturhfprop compose snippet — the `iturhfprop:` service has no `image:` tag, no `restart:` policy, no healthcheck, and no port mapping. Users following the snippet will get a working container but it's less robust than the main service block. Minor — could mirror the main service's structure for consistency, or leave it minimal since it's an internal-only service on the compose network.
- The dropped sentence "(in the `iturhfprop-service/` directory)" → "(documentation and deployment examples in `iturhfprop-service/README.md`)" is a slight regression in discoverability — the directory pointer is useful for people who don't think to open the subdir README. Could mention both.

CI itself: the checks list is empty here, which usually means GitHub didn't trigger them for an `main`-targeted PR with this profile — that'll sort itself once the base is changed to `Staging`.

— K0CJH

accius:
I have repointed the PR to Staging from my side.

accius:
After pointing to staging it looks like you will need to run a prettier on the code and update the PR.

lbatalha:
The lint issue was related to me leaving the correct indentation for the compose file in the code snippet, this is important so new people copy pasting dont accidentally break the yaml, so I added a quick skeleton of a compose file to provide more context, what do you think?

As for the compose build issue, not sure how claude got that review wrong as its well [documented behavior](https://docs.docker.com/reference/compose-file/build/#using-build-and-image) 😅.

As for the small comments:

- README change is deliberate, but depends on the project preferences
- I've added the restart policy, the rest do not make sense for this service
- I think linking to the doc right away achieves the same as before, but is a bit more specific

Sorry about targeting this to main, completely forgot its supposed to go to staging 🤦‍♀️

accius (commented):
Re-reviewed against `a6ac4dd2`. Two things picked up from the last round, but a couple of items still open.

**Resolved:**

- ✅ Retargeted to `Staging`.
- ✅ iturhfprop compose snippet now includes parent service context (`services:` / `openhamclock:` / `...`) so the indentation reads cleanly.
- ✅ Restart policy added to the snippet.

**Still open:**

1. **The restart policy key is wrong.** The snippet has `restart_policy: unless-stopped`, but Compose's top-level key is `restart` — `restart_policy` is a Swarm-mode `deploy.restart_policy` thing and Compose will reject or ignore it. The main service in `docker-compose.yml` uses `restart: unless-stopped` (correct). Please change the snippet to match:

   ```yaml
     iturhfprop:
       build:
         context: ./iturhfprop-service
       restart: unless-stopped
       container_name: iturhfprop
   ```

2. **`build:` + `image:` together in `docker-compose.yml` — still the same behavior.** With both keys present, `docker compose up` still *builds* by default and tags the build as `ghcr.io/accius/openhamclock:latest`; it does not automatically pull. So the README change implies "use the prebuilt GHCR image" but the compose default still triggers a local build. Either:

   - Remove `build: .` from the default compose file and document `docker compose build` for the (smaller) audience who wants to build from source, **or**
   - Add a note to the README telling users to `docker compose pull` (and optionally `--no-build`) if they want the registry image.

   The current state will leave most users building from source while believing they're pulling from GHCR.

**Heads-up on CI** — there are no check runs against your latest commit (`a6ac4dd2`). This is a fork-PR thing; the workflows need to be re-approved after the base-branch change. @accius will sort that — nothing for you to do.

Once #1 is fixed and the compose-default question is settled either way, this is mergeable.

— K0CJH

lbatalha:
Fixed the policy key typo.

As for the #2, your understanding is incorrect as per the docs I have link previously. You can also test it out with docker or podman compose, both by default will always attempt to pull the image first, and only if it doesnt exist will trigger a build. The user can always trigger a build for testing using `docker compose build`.

Here is an example:
```bash
❯ docker rmi ghcr.io/accius/openhamclock:latest
Untagged: ghcr.io/accius/openhamclock:latest
[...]
```

```bash
❯ docker compose up -d --wait --remove-orphans
 ✔ Image ghcr.io/accius/openhamclock:latest Pulled                                                        6.9s
 ✔ Network ohc                              Created                                                       0.1s
 ✔ Container openhamclock                   Healthy                                                       7.1s
 ✔ Container iturhfprop                     Healthy                                                       2.1s
```

accius:
Looks good! (sorry for the faulty review, Hadn't gotten chance to check against the automated triage I run at the beginning of today's work) Merged! Thanks for your first contribution to OHC!!!

PR #1021, opened 2026-05-26
feature(docker): iturhfprop container image workflow

lbatalha:
## What does this PR do?

This PR adds a github workflow to automatically build and publish a container image for the iturhfprop microservice.

It also changes the documentation and associated snippets to reference this image. (It also fixes the snippet linting issue I missed earlier 🥲 )

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Github workflow should auto-trigger on similar conditions as the existing `docker-build` one.

## Extra thoughts

I wonder if this microservice should have its own repo, would that make sense?

### DIFF
--- a/.github/workflows/docker-image-iturhfprop.yml
+++ b/.github/workflows/docker-image-iturhfprop.yml
@@ -1,0 +1,81 @@
+name: Build and Publish ITURHFProp Docker Image
+
+on:
+  push:
+    branches: ['main']
+    paths:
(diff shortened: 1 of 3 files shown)


lbatalha:
Closing, after talking about it, It would be better to also do a review of the existing docker workflows

PR #1028, opened 2026-05-28
fix(actions): update github actions to node24-based versions

lbatalha:
## What does this PR do?
This updates all non-docker workflow actions to versions that are Node 24 based, as Node 20 will stop being bundled in github workers starting [June 2nd 2026](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)

For each action I had a look at the release history and found no breaking changes that apply to us. https://github.com/actions/github-script/releases/tag/v9.0.0 this appears to be the one with most of the breaking changes, but I do not think they apply to our current use.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

- All actions should complete correctly


accius (dismissed):
Two things on this.

1. This targets `main`, but we run everything through `Staging` first and only cut `main` on release. Could you retarget the PR to `Staging`? Same diff, just different base branch.

2. The version bumps look reasonable but worth calling out that the routine CI only actually exercises a subset of these workflows. The PR run here covered `ci.yml` and `build-wasm.yml`, which both pass with the new versions, so `actions/checkout@v6`, `actions/setup-node@v6`, `actions/upload-artifact@v7`, and `mymindstorm/setup-emsdk@v16` are confirmed good in our setup.

The other ones are not exercised by a normal PR:

- `auto-in-progress.yml` and `close-issue.yml` use `actions/github-script@v9`, which is the bump with the biggest breaking-change surface. We won't actually find out if those still work until someone opens or comments on an issue.
- `publish-p533-data.yml` only fires on a manual or scheduled trigger.
- `build-rig-listener.yml` and `rig-listener-build.yml` only fire on rig-listener releases, so the `download-artifact@v8` / `softprops/action-gh-release@v3` jumps are unverified until the next rig-listener release.

Once this is on Staging I'd want to manually fire a workflow_dispatch on `publish-p533-data` and post a test comment somewhere that triggers `close-issue` / `auto-in-progress` before letting it move to main, just so we don't find out something broke on the day we actually need one of those workflows. Happy to do that test pass on my side after merge if it's easier.

accius:
I moved this to point to staging

PR #1041, opened 2026-05-31
fix(rig-bridge): stop auth retry storm and rate-limit server log warnings

ceotjoe:
## Summary

- **rig-bridge client** (`plugins/cloud-relay.js`): on a 401/403 response, a new `handleCredentialsInvalid()` helper permanently stops both the push timer and the long-poll loop, logging once with a clear action for the user. Previously, `pushState` kept retrying every 2 s and `longPollCommands` retried every 1 s on any non-200 — flooding the server log at ~1 warn/s indefinitely.
- **Server** (`server/routes/rig-bridge.js`): `requireRelayAuth` now rate-limits `logWarn` to once per 5 minutes per sessionId. MeshCom ingest failure warnings are similarly throttled to once per minute per subtype.

## Test plan

- [ ] Configure rig-bridge with a stale/wrong session token and confirm the server log shows only one auth warning, then silence for 5 minutes
- [ ] Confirm rig-bridge logs a single clear error message and stops retrying after receiving a 401
- [ ] Re-run Connect Cloud Relay in OHC Settings → Rig Bridge, confirm normal relay operation resumes with fresh credentials
- [ ] Verify no regression in normal relay auth (valid credentials still pass through)

### DIFF
--- a/rig-bridge/package.json
+++ b/rig-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openhamclock-rig-bridge",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "description": "Universal rig control bridge for OpenHamClock — plugin architecture supporting USB, flrig, rigctld, and more",
   "main": "rig-bridge.js",
   "bin": "rig-bridge.js",
--- a/rig-bridge/plugins/cloud-relay.js
+++ b/rig-bridge/plugins/cloud-relay.js
@@ -62,6 +62,7 @@ const descriptor = {
     let stateChangeHandler = null;
     let lastPttState = null; // Track PTT separately to detect PTT-specific changes
     let serverReachable = false;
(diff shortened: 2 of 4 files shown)


accius:
Couple of small things noticed during review, neither blocking, both are follow-up-able whenever you feel like it:

The `authWarnLastLogged` map keys on sessionId prefix but never clears stale entries. If a misbehaving client (or a deliberate one) cycles through a lot of bad sessionIds, that map slowly grows. The `relayPollCountByIP` cleanup pass that runs every 5 minutes a few lines up could probably absorb this one with a small tweak.

The other one is just messaging. The error tells the user to "re-run Connect Cloud Relay in OHC Settings" but in practice they also need to restart rig-bridge for the new credentials to actually take effect, since the old token is still in memory and `credentialsInvalid` stays set for the life of the process. Maybe worth adding "and restart rig-bridge" to that string so people don't bounce back wondering why it's still broken after clicking the button.

PR #1042, opened 2026-05-31
fix(rig-bridge): address PR #1041 review feedback

ceotjoe:
Follow-up to #1041.

## Summary

- **authWarnLastLogged unbounded growth**: the existing 5-minute session cleanup interval now also prunes `authWarnLastLogged` entries older than 2× `AUTH_WARN_INTERVAL` (10 min), so a client cycling through many bad sessionIds cannot grow the map indefinitely.
- **Misleading error message**: the credentials-invalid log line now reads *"…get fresh credentials, then restart rig-bridge"* — without the restart hint, users clicking Connect Cloud Relay would find the relay still broken because `credentialsInvalid` stays set for the life of the process.

## Test plan

- [ ] Confirm `authWarnLastLogged` entries are removed on the 5-minute sweep after their TTL expires
- [ ] Confirm the updated error message appears in rig-bridge console on a 401 response

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -118,6 +118,10 @@ THEME=dark
 # Layout: 'modern' or 'classic'
 LAYOUT=modern
 
+# SERVERLOCAL - if set to true overrides all useLocalInstall() tests in the client; otherwise
+# falls back on the local tests.
+# SERVERLOCAL=false
+
 # Timezone: IANA timezone identifier
 # Set this if your local time shows incorrectly (e.g. same as UTC).
 # Common with privacy browsers that spoof timezone.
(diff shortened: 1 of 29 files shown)


accius (approved):
Both followups from #1041 look right.

The cleanup-loop addition in `server/routes/rig-bridge.js` reuses the existing 5-minute interval and prunes `authWarnLastLogged` entries older than `AUTH_WARN_INTERVAL * 2` — that's the correct cutoff because anything newer is still inside the rate-limit window. The map can't grow unbounded anymore even if a client cycles through fresh sessionIds.

The error-message tweak in `cloud-relay.js` is the right call too — `credentialsInvalid` stays sticky for the life of the process, so "get fresh credentials" without "then restart rig-bridge" would have left users stuck.

Good to merge.

K0CJH

Issue #1001, opened 2026-05-19
[Accessibility] Screen-reader spot list announcements + semantic table markup


accius:
Sub-issue of #997 (accessibility umbrella).

## Problem

The DX Cluster, POTA, SOTA, WWFF, and PSK Reporter spot lists are built from `<div>` grids today. Screen readers can read them, but row/column relationships aren't communicated — a user can't ask the reader to "next column" or "this row's frequency" the way they could with a real `<table>`.

Additionally, frequencies are rendered as plain numbers (e.g. "14.074") which most screen readers will pronounce as "fourteen point zero seven four" — usable, but not the natural ham announcement "fourteen point zero seven four megahertz" that gives context immediately.

## What we want

1. Convert the spot grids to semantic tables (`<table>`, `<thead>`, `<tbody>`, `<tr>`, `<th>`, `<td>`) with `scope` attributes on headers. Keep the same CSS grid look — just under semantic markup.
2. Add a hidden `<span class="visually-hidden">` after each frequency that says "megahertz" so screen readers get the unit, while sighted users see just the number.
3. Same for age ("3 m ago" → screen reader hears "3 minutes ago" via aria-label or visually-hidden span).

## Scope

Affects DXClusterPanel, ActivatePanel (covers POTA/SOTA/WWFF), and PSKReporterPanel. Be careful not to regress the visual density work done in #995.

accius:
Done in 33ed861 on Staging.

All three items in the issue body addressed:

1. **Semantic table markup**: applied as ARIA roles on the existing div grids (`role="table"` / `role="row"` / `role="columnheader"` / `role="cell"`) rather than converting to real `<table>` elements. Real `<table>` plus `display: grid` would have lost the table role in the accessibility tree, so the ARIA-on-divs path preserves the #995 density work pixel-for-pixel while still giving screen readers a proper table to navigate. Hidden header rows added on all three panels so column navigation has labels even though the visual UI omits them.

2. **Frequency "megahertz" announcement**: new `.visually-hidden` utility class in `main.css` (standard clip-path pattern), and a hidden `<span> megahertz</span>` after each numeric frequency in DXClusterPanel, ActivatePanel, and PSKReporterPanel.

[...]

Issue #997, opened 2026-05-19
[Accessibility] Audit and improve experience for blind / screen-reader users


accius:
## Is your feature request related to a problem?

A blind ham operator reached out noting that very few ham radio applications are accessible to blind users — and asked whether OHC has been audited for screen-reader compatibility. The honest answer is: not systematically.

This isn't a single bug. It's a tracking/umbrella issue for a structured accessibility pass across the whole app, with the explicit goal of making sure blind and low-vision hams get a usable, dignified experience — not just a stripped-down "lite" version.

## Describe the Solution You'd Like

A structured audit and follow-up improvements covering, at minimum:

**Screen reader support**

- Verify every interactive element has an accessible name (`aria-label`, associated `<label>`, or visible text)
- Add `aria-live` regions for status changes — band changes, new spots, WSJT-X decodes, satellite pass alerts, lightning proximity alerts, etc. — so screen-reader users get notified of important events without polling the UI
- Ensure landmark roles (`<nav>`, `<main>`, `<aside>`, etc.) are correct
- Verify every panel is reachable and operable via keyboard alone

**Keyboard navigation**

- All controls reachable via Tab in a logical order
- All interactive elements operable via Enter / Space
- Visible focus indicators that are not removed by `outline: none`
- Standard shortcuts considered (e.g. `/` to focus a search, `?` for help)

**Visual / non-screen-reader accessibility**

- Color contrast meets WCAG AA in all themes (Dark, Light, Retro, custom)
- Information not conveyed by color alone — band-color heatmaps, propagation maps, status indicators all need text/icon alternatives
- Charts and data visualizations have textual summaries available to screen readers

**Data tables / spot lists**

- Use semantic table markup where appropriate so screen readers can navigate by column / row
- Frequency and time announced in human-readable forms (e.g. "fourteen point zero seven four megahertz, three minutes ago")

**Map**

- A non-map view of the same data for screen-reader users (e.g. a list of spots with bearing + distance from DE)? Or at least an `aria-described` summary of what's currently on the map?

## Describe Alternatives You've Considered

- Linking to a separate screen-reader-friendly site would let other dashboards remain "the accessible one" while OHC stayed "the rich/visual one". Not preferred — we want OHC to be the inclusive option, not delegate.

## Use Case

[...]

PR #1043, opened 2026-05-31
feat(dx): callsign lookup in DX target window (#1036)

### LINKED ISSUE #1036: [FEATURE]  Allow for callsign lookup in DX target window
## Is your feature request related to a problem? No

I sometimes have a callsign already and wish to enter it into OHC to find out the beam heading and location on the map. Currently the only method I have to look it up on QRZ and check the details tab which is kinda slow!

In the DX Target tab there would be some kind of option to type in a callsign. When this happens, it pulls the grid square and beam heading (and other information).

## Use Case

- [ x ] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [ x ] Contesting
- [ x ] POTA/SOTA
- [ x ] DXing

- [ ] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ x ] No, but I'd use it!

## Additional Context

Add any other context, screenshots, or mockups about the feature request here.

Are there similar features in other amateur radio software? If so, what do you like/dislike about their implementation?

pretty much the only software i have seen that sorta does this is N1MM. Its probably found in other logging software as well!
---

73!

ceotjoe:
## Summary

- Adds an interactive `DXCallsignInput` field to the DX target section in all three layouts (DockableApp, Modern, Classic)
- User types a callsign and presses Enter — resolves via the existing `/api/callsign/:call` endpoint (QRZ → HamQTH → prefix fallback) and sets the map pin + grid square automatically
- Replaces the previous static callsign badge; the field stays in sync when spots are clicked in the DX cluster
- No backend changes required

## Test plan

- [ ] Type a known callsign (e.g. `DL1ABC`) in the DX target callsign field and press Enter → map pin moves, grid square updates
- [ ] Test with a callsign near lat=0 or lon=0 (equator/prime meridian) — position must not be dropped
- [ ] Escape key reverts the input without triggering a lookup
- [ ] Lock the DX position → field becomes read-only
- [ ] Enter an unknown/invalid callsign → field flashes red and reverts
- [ ] Click a spot in the DX cluster → callsign field updates to match
- [ ] Verify behaviour in all three layouts: Modern, Classic, Dockable

Closes #1036

Made callsign editable in DX panel:

Bildschirmfoto 2026-05-31 um 16 47 08


ceotjoe:
Waits on https://github.com/accius/openhamclock/pull/1038, which also has changes in the same area.

lbatalha:
#1038 has been merged, I think you can go ahead with this. I will create a PR later to remove the timezone toggle feature entirely from the panel, but that shouldn't overlap your changes.
I am also going to submit one for the callsign lookup popup thing which adds a new authenticated api fetch from HamQTH, for the API endpoint you are calling, but that should just have similar fallback/fallthrough logic (QRZ -> HamQTH -> DXCC HamQTH -> builtin country lookup) and be seamless.

chrisdebian (commented):
Nice implementation — the `escapingRef` pattern to prevent blur-triggering a lookup on Escape is a clean solution, and the AbortController usage is correct.

A few things worth considering:

**1. Missing `aria-label` on the input**
The `title` attribute isn't reliably announced by screen readers. Given the open accessibility tracking issue (#997), it's worth adding an explicit `aria-label`:
```jsx
aria-label={t('app.dxLocation.callsignInputAriaLabel', 'DX target callsign')}
```

**2. No cleanup on unmount**
The `abortRef.current` controller isn't cancelled if the component unmounts during a fetch. A cleanup effect would prevent the "Can't perform a React state update on an unmounted component" warning:
```javascript
useEffect(() => {
  return () => abortRef.current?.abort();
}, []);
```

**3. Error state not announced to screen readers**
The red border flash is visual-only. Adding `aria-invalid={error}` and an `aria-live` region for the error message would make the failure state accessible — directly relevant to #997.

Otherwise this looks solid and well thought-through — the lock behaviour, sync from external prop, and loading state are all handled cleanly.

ceotjoe:
@chrisdebian 
Hi Chris, thanks for the thorough review — really appreciated!

All three points have been addressed in the latest commit (`a7c25fa`):

- **`aria-label`** — added an explicit `aria-label` on the input (backed by an i18n key across all 16 language files), alongside the existing `title`.
- **Unmount cleanup** — added a cleanup `useEffect` that calls `abortRef.current?.abort()` on unmount, so no state updates fire on dead components.
- **Error announced to screen readers** — added `aria-invalid={error}` and `aria-describedby` on the input, plus a visually-hidden `role="alert" aria-live="assertive"` region that reads out the localised "Callsign not found" string when a lookup fails. Should tick the box for #997.

73 de Jörg

chrisdebian (approved):
All three points addressed cleanly:
- `aria-label` on the input ✅
- Unmount cleanup via `useEffect(() => () => abortRef.current?.abort(), [])` ✅
- `aria-invalid` + `aria-describedby` + visually-hidden `aria-live="assertive"` region ✅
- All 16 language files ✅

The `escapingRef` pattern is clean, abort controller handling is correct, and the live region pattern is the right approach for #997.

accius (approved):
Read through this end-to-end, like the design a lot. AbortController on the in-flight fetch + abort on unmount is exactly the pattern I want to see, and using a ref flag for "this blur came from Escape, skip the commit" is a clean way to dodge the double-fetch trap. The `data.lat != null && data.lon != null` check correctly survives lat=0 / lon=0 callsigns, which is the kind of thing that's easy to miss. Translation coverage across all 16 languages is also nice to see in one PR rather than as follow-ups.

Six nits, none blocking, all small enough to take or leave:

**1. The visually-hidden span at the bottom of the component uses inline styles** that duplicate the `.visually-hidden` class added in #1001 (`src/styles/main.css:24`). Could swap the inline `position/width/height/clip/overflow` block for `className="visually-hidden"` and lose 7 lines.

**2. `role="alert"` and `aria-live="assertive"` are doing the same job.** `role="alert"` implicitly creates an assertive live region. Either alone is enough. Pick one.

**3. Direct DOM border manipulation in `handleBlur` / `handleFocus`:**
```js
inputRef.current.style.borderBottomColor = 'transparent';
```
Works, but it's the kind of thing that'll have a future maintainer reaching for `useState` instead. A `focused` boolean state + driving `borderBottomColor` from the style prop would be more React-idiomatic.

**4. The `setTimeout(() => setError(false), 2000)` in the catch block has no cleanup.** If the component unmounts during the 2s error window, setError fires on a dead component. Stash the timer id in a ref and clear it in the unmount effect, same way `abortRef` is handled.

**5. Subtle race**: if `dxCallsign` prop changes while a fetch is in flight (user clicks a different DX cluster spot mid-lookup), the sync useEffect updates inputValue to the new prop, but the in-flight fetch still resolves and calls `onDXChange` with the previous lookup result. Aborting on dxCallsign change too would close that gap. Edge case but real.

**6. Empty-state UX**: the input replaces the conditional `{dxCallsign && <span>}`, so when no DX is set the input renders as blank space with no placeholder. Users won't know they can click there. Adding `placeholder={t('app.dxLocation.callsignInputPlaceholder', 'Callsign?')}` (or similar) would help discoverability.

Otherwise this is good to go. Approving.

PR #1044, opened 2026-05-31
fix(smartsdr): subscribe to transmit status so PTT/MOX state is detected

### LINKED ISSUE #1035: [BUG]  SmartSDR plugin: transmit/PTT state never detected (On Air stays RX during TX)
### Summary
With a FlexRadio connected via the SmartSDR plugin, frequency and mode track
correctly, but transmit state is never detected. The On Air indicator stays on
RX while the radio is transmitting. With `--debug` enabled, keying the radio
produces **no console output at all** — no transmit/tx/MOX traffic is logged.
This suggests the SmartSDR plugin subscribes to slice status (frequency/mode)
but not to the radio's transmit status, so the Flex never sends transmit-state
updates.

### Environment
- Radio: FlexRadio FLEX-8600
- SmartSDR version: v4.2.18
- rig-bridge version: v2.2.0
- OS: Windows 11
- Connection: SmartSDR plugin (direct to radio, port 4992), Slice Index 0
- OpenHamClock: hosted version at openhamclock.com via Cloud Relay

### Steps to reproduce
1. Configure rig-bridge with Radio Type = SmartSDR, pointed at the Flex.
   Confirm it connects (frequency and mode display and track correctly).
2. Stop the running instance and start rig-bridge from the command line with
   debug logging: `node rig-bridge.js --debug`
3. Confirm OpenHamClock shows RX.
4. Key the radio (physical PTT button).

### Expected behavior
rig-bridge detects the transmit state and OpenHamClock's On Air indicator
switches from RX to TX while transmitting.

### Actual behavior
- SmartSDR's own TX indicator correctly shows the radio is transmitting.
- OpenHamClock's On Air indicator stays on RX.
- The `--debug` console shows **nothing** when keying up — no transmit, tx, or
  MOX messages are received or logged.
- Frequency and mode continue to track correctly throughout, so the radio
  connection itself is healthy.

### Additional notes
- Tested with a single rig-bridge instance running (no duplicates).
- Slice Index is 0 / Slice A, which is the operating slice.
- Because the console shows zero transmit traffic, this looks like a missing
  transmit-status subscription in the SmartSDR plugin rather than a parsing or
  display issue. On the SmartSDR API, transmit state comes from a separate
  subscription than slice/frequency status.

ceotjoe:
## Summary

- Adds `sub radio all` subscription on connect so the FlexRadio delivers MOX/transmit-state events
- Handles `interlock state=` status lines as the primary PTT source
- Bumps rig-bridge to v2.2.2

## Test plan

- [ ] Connect rig-bridge to a FlexRadio via SmartSDR plugin
- [ ] Confirm frequency and mode still track correctly
- [ ] Key the radio — On Air indicator should switch from RX to TX
- [ ] With `--debug`, keying up should log `[SmartSDR] PTT → TX (mox)`
- [ ] Unkey — indicator should return to RX

Fixes #1035

### DIFF
--- a/rig-bridge/package.json
+++ b/rig-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openhamclock-rig-bridge",
-  "version": "2.2.1",
+  "version": "2.2.2",
   "description": "Universal rig control bridge for OpenHamClock — plugin architecture supporting USB, flrig, rigctld, and more",
   "main": "rig-bridge.js",
   "bin": "rig-bridge.js",
--- a/rig-bridge/plugins/smartsdr.js
+++ b/rig-bridge/plugins/smartsdr.js
@@ -80,7 +80,7 @@ module.exports = {
         console.log(`[SmartSDR] Session handle: ${handle}`);
         sendCmd('sub slice all');
         // Subscribe to transmit status so MOX/PTT changes are delivered
-        sendCmd('sub tx');
+        sendCmd('sub tx all');
         return;
       }
 
--- a/rig-bridge/plugins/smartsdr.js
+++ b/rig-bridge/plugins/smartsdr.js
@@ -78,8 +78,9 @@ module.exports = {
       if (firstChar === 'H') {
         handle = line.slice(1).trim();
         console.log(`[SmartSDR] Session handle: ${handle}`);
-        // Subscribe to slice status updates
         sendCmd('sub slice all');
+        // Subscribe to transmit status so MOX/PTT changes are delivered
+        sendCmd('sub tx');
         return;
       }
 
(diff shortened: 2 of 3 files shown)


ceotjoe:
I can not test this on my end, no FlexRadio. :-)

accius (approved):
Clean fix for the missing PTT/MOX detection.

Subscribing to `sub radio all` is the right channel — the SmartSDR API does push interlock status through the radio object rather than slice or transmit, so without it we'd never see TRANSMITTING. Treating `TRANSMITTING` and `PTT_REQUESTED` as on-air (and everything else as off) matches the SmartSDR state machine — `PTT_REQUESTED` is the brief gate before the relay actually closes, so calling that TX too means the On Air indicator doesn't lag a few hundred ms behind the key.

Dropping the old `key === 'tx'` branch in the slice parser is correct — that key never actually arrived in slice updates, which is why this bug existed in the first place.

Good to merge once you've smoke-tested with a real Flex.

K0CJH

PR #1045, opened 2026-05-31
feat(a11y): implement ARIA tablist pattern across all tab strips (#1031)

### LINKED ISSUE #1031: [Accessibility] SettingsPanel + MeshComPanel tab strips should use the ARIA tablist pattern
Sub-issue of #997 (accessibility umbrella). Surfaced while closing #1003 — out of scope there because this is a real semantic refactor, not a mechanical aria-label add.

## Problem

The top tab strip in `src/components/SettingsPanel.jsx` (Station / Integrations / Display / Map Layers / Satellites / Profiles / Community / Alerts / Rig Bridge) is built as a row of nine `<button>` elements that toggle `activeTab`:

```jsx
<button
  onClick={() => setActiveTab('station')}
  style={{ ... active styling driven by activeTab === 'station' ... }}
>
  📻 {t('station.settings.tab.title.station')}
</button>
```

Currently nothing tells assistive tech this is a tab list. A screen reader user just hears nine separate buttons with no relationship to the panel below.

This was deliberately left alone in the #1003 mechanical pass because the right fix isn't `aria-pressed` — it's the full ARIA tablist pattern.

## What we want

Wrap the tab strip in a `role="tablist"` container, change each tab button to `role="tab"`, and use `aria-selected={activeTab === '<id>'}` instead of (or in addition to) the styling-driven active state. Wire each tab to its corresponding panel content with `aria-controls` / `id` pairs, and put `role="tabpanel"` on the panel container so the relationship is announced.

Also wants the standard arrow-key navigation between tabs (left/right arrows move focus along the strip, Home/End jump to first/last), and `tabindex` roving so only the active tab is in the tab order. The W3C ARIA Authoring Practices "Tabs" pattern is the reference: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

## Scope note

The MeshComPanel tab strip (just landed in 62a2a3a with `aria-pressed`) has the same underlying issue and should get the same treatment in this PR. Same pattern, same conversion. Worth doing in one shot rather than spreading the refactor across multiple PRs.

Anything else in the codebase using a tab-strip pattern should also be caught in the same pass: search for `setActiveTab` and `activeTab ===` to find them.

ceotjoe:
## Summary

- Adds shared `src/utils/ariaTabKeyDown.js` utility for W3C APG arrow-key / Home / End navigation
- Converts all 7 tab-strip components to the full ARIA tablist pattern: `role="tablist"`, `role="tab"` + `aria-selected` + `aria-controls`, roving `tabIndex`, `role="tabpanel"` + `aria-labelledby`
- Removes `aria-pressed` from MeshComPanel (wrong pattern for tabs)

**Components updated:** SettingsPanel (9 tabs), MeshComPanel, PotaSotaPanel, DXFilterManager, ActivateFilterManager, PSKFilterManager, PSKReporterPanel

## Test plan

- [ ] Open Settings panel — screen reader announces "Settings tabs" tablist, each tab announces as selected/not selected, arrow keys move between tabs
- [ ] MeshCom panel — same pattern, no more aria-pressed
- [ ] POTA/SOTA panel — arrow keys navigate between POTA/WWFF/SOTA/WWBOTA
- [ ] DX Filter, Activate Filter, PSK Filter modals — tablist navigation works
- [ ] PSKReporter tx/rx sub-tabs — tablist with arrow keys
- [ ] Visual styling unchanged across all components

Closes #1031

### DIFF
--- a/src/components/ActivateFilterManager.jsx
+++ b/src/components/ActivateFilterManager.jsx
@@ -2,7 +2,8 @@
  * ActivateFilterManager Component
  * Filter modal for ActivatePanel type spots - Bands, Grids, Modes
  */
-import React, { useState } from 'react';
+import React, { useState, useRef } from 'react';
+import { ariaTabKeyDown } from '../utils/ariaTabKeyDown.js';
 
 const BANDS = ['160m', '80m', '60m', '40m', '30m', '20m', '17m', '15m', '12m', '10m', '6m', '2m', '70cm'];
 const MODES = [
@@ -84,6 +85,8 @@ const GRID_REGIONS = [
 ];
 
 export const ActivateFilterManager = ({ filters, onFilterChange, isOpen, onClose, name }) => {
+  const ACTIVATE_TABS = ['bands', 'grids', 'modes'];
+  const activateTabRefs = useRef({});
   const [activeTab, setActiveTab] = useState('bands');
   const [customGrid, setCustomGrid] = useState('');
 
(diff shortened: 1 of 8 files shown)


accius (approved):
Nice cleanup of the tablist pattern across all seven tab strips. The shared `ariaTabKeyDown` helper is exactly the kind of utility that should exist for this — arrow / Home / End wrapped behavior in one place, components just pass their tab-id array and the active state setter.

A few things I liked:

The roving `tabIndex` (0 on active, -1 on others) is the W3C APG pattern and means tab key skips the inactive tabs entirely — keyboard users don't have to step through every tab to leave the strip. The `ref` callback that populates `tabRefs.current` and the `setActiveTab + .focus()` move on arrow press completes the loop.

Dropping `aria-pressed` from MeshComPanel is correct — `aria-pressed` is the toggle-button pattern and was lying about what those tabs were. Replacing it with `aria-selected` + `role="tab"` is the right semantic.

MeshComPanel switching from conditional mount to `hidden`/`display:none` panels is the only non-trivial behavior change. It means all three tab subtrees stay mounted, which has a small memory cost but matches APG and gives screen readers stable `aria-controls` targets. I think that's the right tradeoff here — those panels are cheap.

PSKReporter only wires the tablist for the PSK panel mode (not for the wsjt/Decodes/QSOs mode) — that's a deliberate scope decision and reasonable. Worth a followup later but not blocking.

Good to merge.

K0CJH

PR #1049, opened 2026-06-01
fix(setup-pi): resolve kiosk session type at install time (#1026)

### LINKED ISSUE #1026: [BUG] Raspberry Pi setup, --kiosk mode causes white-screen at boot
## Describe the Bug

- noted in https://github.com/accius/openhamclock/pull/1025
- after running `scripts/setup-pi.sh --kiosk` the installation was successful however at device reboot at first the desktop appeared followed sometime later by a complete white screen.
- internet search implied that problem was caused by detection or setting of Wayland screen type.
- I manually modified './kiosk.sh' (which is a file created by the setup) as follows,

was
```
SESSION_TYPE="${XDG_SESSION_TYPE:-}"
if [ -z "$SESSION_TYPE" ] && [ -n "$WAYLAND_DISPLAY" ]; then
    SESSION_TYPE="wayland"
fi
if [ -z "$SESSION_TYPE" ]; then
    # Last resort: default to x11 so the script always does something useful
    SESSION_TYPE="x11"
fi
```

changed to
```
#SESSION_TYPE="${XDG_SESSION_TYPE:-}"
#if [ -z "$SESSION_TYPE" ] && [ -n "$WAYLAND_DISPLAY" ]; then
#    SESSION_TYPE="wayland"
#fi
#if [ -z "$SESSION_TYPE" ]; then
    # Last resort: default to x11 so the script always does something useful
#    SESSION_TYPE="x11"
#fi
SESSION_TYPE="x11"
```

- subsequent reboot of device and OHC loaded

## To Reproduce
As described

## Expected Behavior
Expected OHC to load on device reboot

## Screenshots
None

## Environment

- **OS**: [e.g., Windows 11, macOS Sonoma, Ubuntu 22.04, Raspberry Pi OS]
- **Browser**: [e.g., Chrome 120, Firefox 121, Safari 17]
- **Node.js Version**: [e.g., 20.10.0]
- **OpenHamClock Version**: [e.g., 3.0.0]
- **Running As**: [Web browser / Electron app / Docker]

## For Raspberry Pi Users

- **Pi Model**: Raspberry Pi 3B+
- **Pi OS Version**: Bookworm 64-bit
- **Display**: HDMI to Sony TV screen
- **Running in Kiosk Mode**: YES

## Console Errors
None

## Additional Context
None

accius:
Fixes #1026 (white screen at boot on Pi 3B+ / Bookworm with `--kiosk`).

## Root cause

The generated `kiosk.sh` was detecting the display server at boot via `XDG_SESSION_TYPE` plus a `WAYLAND_DISPLAY` fallback. On Michael's Bookworm 64-bit Pi (X11) that fallback picked the wayland branch and Chromium came up with `--ozone-platform=wayland` against an X server — hence the white screen.

## Fix

Move the X11-vs-Wayland choice from boot time to install time:

1. New CLI flag: `--session-type=auto|x11|wayland` on `setup-pi.sh`. Default `auto`.
2. New `resolve_session_type` function picks via:
   - explicit override
   - `/etc/os-release` codename (bookworm/bullseye/buster → x11, trixie → wayland)
   - `XDG_SESSION_TYPE` from the installer's shell
   - `x11` as the last-resort default (never silently picks wayland)
3. The generated `kiosk.sh` now bakes the resolved value in as a single constant, with a runtime override via `OPENHAMCLOCK_SESSION_TYPE` env var if anyone needs to flip it without re-running setup.

This means Michael's Bookworm case would have resolved to `x11` at install time and never hit the boot-time wayland path that was producing the white screen.

## Test plan

Smoke-tested locally:
- `bash -n` passes
- `resolve_session_type` returns the right value for fake `/etc/os-release` codenames (bookworm/bullseye → x11, trixie → wayland), explicit overrides win, garbage values rejected by the validator
- Generated `kiosk.sh` runs end-to-end with both baked-in defaults; `OPENHAMCLOCK_SESSION_TYPE` override works

Needs a real-Pi test to fully validate — Michael, would you mind re-running `scripts/setup-pi.sh --kiosk` from this branch on the same Pi 3B+ Bookworm box and confirming it comes up clean at next boot? The installer should print `Kiosk display server: x11` and `~/openhamclock/kiosk.sh` should contain `SESSION_TYPE="${OPENHAMCLOCK_SESSION_TYPE:-x11}"`.

If the codename detection still misses your setup, `--session-type=x11` on the CLI is the explicit escape hatch.

K0CJH

### DIFF
--- a/scripts/setup-pi.sh
+++ b/scripts/setup-pi.sh
@@ -199,7 +199,11 @@ esac
 # Resolving here, at install time, lets the user see and override the
 # choice up front, and bakes a single constant into kiosk.sh.
 resolve_session_type() {
+    # Diagnostic output goes to stderr so it shows in install logs without
+    # polluting the function's stdout (which is the resolved value). Useful
+    # while we're still ironing out why auto-detect picks wrong on some Pis.
     if [ "$SESSION_TYPE_OVERRIDE" = "x11" ] || [ "$SESSION_TYPE_OVERRIDE" = "wayland" ]; then
+        echo "  [session-detect] CLI override: $SESSION_TYPE_OVERRIDE" >&2
         echo "$SESSION_TYPE_OVERRIDE"
         return
     fi
@@ -208,16 +212,28 @@ resolve_session_type() {
         # shellcheck disable=SC1091
         local codename
         codename=$(. /etc/os-release && echo "${VERSION_CODENAME:-}")
+        echo "  [session-detect] /etc/os-release VERSION_CODENAME='$codename'" >&2
         case "$codename" in
-            bookworm|bullseye|buster) echo "x11"; return ;;
-            trixie) echo "wayland"; return ;;
+            bookworm|bullseye|buster)
+                echo "  [session-detect] codename matched debian X11 line, picking x11" >&2
+                echo "x11"; return ;;
+            trixie)
+                echo "  [session-detect] codename matched trixie, picking wayland" >&2
+                echo "wayland"; return ;;
         esac
+        echo "  [session-detect] codename '$codename' did not match any known release, falling through" >&2
+    else
+        echo "  [session-detect] /etc/os-release not readable, falling through" >&2
     fi
 
+    echo "  [session-detect] XDG_SESSION_TYPE='${XDG_SESSION_TYPE:-}'" >&2
     case "${XDG_SESSION_TYPE:-}" in
-        x11|wayland) echo "$XDG_SESSION_TYPE"; return ;;
+        x11|wayland)
+            echo "  [session-detect] falling back to XDG_SESSION_TYPE: $XDG_SESSION_TYPE" >&2
+            echo "$XDG_SESSION_TYPE"; return ;;
     esac
 
+    echo "  [session-detect] no signal matched, defaulting to x11" >&2
     echo "x11"
 }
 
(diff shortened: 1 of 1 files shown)


MichaelWheeley (commented):
see https://github.com/accius/openhamclock/issues/1026
script runs but Wayland is set and not X11, required use of command line override to force X11.

accius:
Looking at your install log more carefully I noticed every package version carries \`deb13u1+rpt1\` — your Pi has been upgraded to Trixie. The codename heuristic correctly picked \`trixie → wayland\` but you're actually still running X11 (a lot of Pi 3 owners do, because labwc is rough on VideoCore IV).

Pushed [16fead3](https://github.com/accius/openhamclock/commit/16fead3) that swaps the detection order: check \`XDG_SESSION_TYPE\` first (what's *actually* running), fall back to codename only when XDG is \`tty\` or empty (which is how SSH installs look, so headless first-installs still get the right OS-default behaviour). I verified the 5 relevant cases by inlining the function and faking env state — your case (XDG=x11 on Trixie) now picks x11.

Would you mind one more spin?

\`\`\`
cd ~/openhamclock
git pull
scripts/setup-pi.sh --kiosk
\`\`\`

The \`[session-detect]\` lines should now show \`using XDG_SESSION_TYPE: x11\` and \`kiosk.sh\` should bake \`SESSION_TYPE="\${OPENHAMCLOCK_SESSION_TYPE:-x11}"\` without needing the \`--session-type=x11\` override.

73 de K0CJH

MichaelWheeley:
see https://github.com/accius/openhamclock/issues/1026#issuecomment-4643212863 
still setting Wayland

Issue #1003, opened 2026-05-19
[Accessibility] Complete the aria-label audit across remaining panels


accius:
Sub-issue of #997 (accessibility umbrella).

## Problem

The first pass under #997 (commit f2d292c) added `aria-label` to icon-only buttons in DXClusterPanel, ActivatePanel, and PSKReporterPanel. The rest of the codebase hasn't been audited yet.

## What we want

Walk through every component and find icon-only buttons (and other interactive elements with no visible accessible name) and add `aria-label`. For toggleable controls also add `aria-pressed`. Components to audit:

- WeatherPanel
- SatelliteWindow / SatelliteInfo
- BandConditionsPanel (N0NBH)
- BandActivityHeatmap (new in 26.3.3)
- Header.jsx (top nav buttons, theme picker, sidebar toggles)
- SettingsPanel.jsx (lots of inline buttons)
- SpaceWeatherPanel
- DigitalModesPanel
- MeshComPanel
- APRSPanel
- ContestPanel
- DXpeditionPanel
- DXNewsTicker
- DXFavorites
- WhatsNew (close button)
- Any panel-header X / minimize buttons in addMinimizeToggle.js

For each: if the button has visible text (e.g. "On"/"Off"), `aria-label` is optional but `aria-pressed` should still be set when it's a toggle. If the button is purely an icon, `aria-label` is required.

## Why "good first issue"

It's mostly mechanical — find icon-only buttons, look at the existing `title=` attribute, copy that to `aria-label`. Following the pattern from f2d292c keeps it consistent. Small, low-risk commits per component are welcome.

accius:
Done in 62a2a3a on Staging. Walked the full list of components from the issue body and applied the f2d292c pattern (aria-label mirrors title, aria-pressed for toggle state).

Touched: BandActivityHeatmap, MeshComPanel, APRSPanel, DXFavorites, SettingsPanel (profile-manager icons, preview-tone, prop-mode/prop-power toggle grids), Header (mobile settings + fullscreen), addMinimizeToggle helper (covers every panel that uses it), and the satellite info window template inside useSatelliteLayer.js.

Checked-and-no-changes-needed: WeatherPanel, BandConditionsPanel, SpaceWeatherPanel, DigitalModesPanel, DXpeditionPanel (no icon-only buttons), DXNewsTicker (already labelled), WhatsNew (close button has visible text).

Two deliberate punts that need their own issue if we want to do them properly:

[...]

Issue #1000, opened 2026-05-19
[Accessibility] Add aria-live regions for status changes


accius:
Sub-issue of #997 (accessibility umbrella).

## Problem

Screen-reader users miss status changes that sighted users notice visually — band changes in WSJT-X, new DX spots arriving, satellite pass beginning/ending, lightning proximity alerts, WSJT-X relay connect/disconnect, Cloud Relay state changes, etc. Without aria-live regions, these updates pass silently.

## What we want

Identify the meaningful status-change events and wrap their announcements in appropriately-prioritised `aria-live` regions:

- **`aria-live="polite"`** — non-urgent: new spots arrived, satellite pass in 10 min, weather refresh
- **`aria-live="assertive"`** — urgent: lightning within proximity radius, severe weather alert, relay disconnected mid-session
- **`role="status"` / `role="alert"`** — semantic roles where they fit better

Be careful with frequency: a live region that announces every single new DX spot would be unbearable. Better to debounce ("3 new spots in the last minute, ZL2ABC the newest") or only announce notable transitions.

## Scope

A short audit of status indicators in OHC, paired with a small set of debounced live-region wrappers in the appropriate components. Probably touches: WSJT-X integration, DX Cluster panel, Satellite predictions, Lightning layer, Weather alerts.

## Tester wanted

If you use a screen reader, your feedback on cadence (too chatty vs. too quiet) would be invaluable.

accius:
First pass shipped in 79fc1a4 on Staging. Covers three of the targets in the issue body:

- **Weather alerts (NWS)** in WeatherPanel — each alert div now has `role="alert"`. New severe-weather alerts announce assertively as they appear.
- **Lightning proximity** — the "X strikes detected" wrapper inside the proximity panel template now has `role="alert" aria-live="assertive"`. Announces on the 0 → N transition. May re-announce when the strike count or closest-distance changes within an active warning, which is acceptable for a life-safety indicator. Smarter transition-only announcement is a follow-up if it proves chatty in practice.
- **WSJT-X relay status** — the "Waiting for WSJT-X" empty state in PSKReporterPanel now has `role="status" aria-live="polite"` on its container. Transitions between disabled / disconnected / connected announce politely. Container disappears once decodes start flowing, so it doesn't keep announcing.

**Still open** (the part where cadence input from a screen-reader user is the bottleneck):

- Debounced DX-spot arrival summary ("3 new spots in last 60s, ZL2ABC the newest")
[...]

Issue #995, opened 2026-05-19
[FEATURE] Toggle visibility of spotter info in DX Cluster panel


accius:
## Is your feature request related to a problem?

A user reports that the DX Cluster panel currently shows both who was spotted (the DX) and who reported the spot (the spotter — the "de XXXX" line), and the spotter info takes up real estate in a panel where vertical space is at a premium. They argue that the important information is *who was spotted*, not who spotted them, and that being able to hide the spotter would tighten the panel up significantly.

## Describe the Solution You'd Like

Add a toggle (panel header gear menu, settings checkbox, or a small icon button on the panel header) that lets the user hide the "de SPOTTER" sub-row on each spot. Choice persisted to localStorage so it survives reloads.

When hidden, each spot row collapses from two visual rows (DX + spotter) to one (DX + age), roughly doubling the spots visible in the same vertical space.

## Describe Alternatives You've Considered

- Smaller font for spotter info — doesn't really solve the density problem.
- Spotter on hover/tooltip — adds an interaction cost vs. just glancing at a denser list.

## Use Case

- [X] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [X] Contesting
- [ ] POTA/SOTA
- [X] DXing
- [ ] Emergency communications
- [ ] Other

Especially valuable for contesters and DXers who care about which DX is being heard rather than who's reporting it — and for users on smaller screens or split layouts where panel real estate is tight.

## Would you be willing to help implement this?

- [ ] Yes, I can submit a PR
- [ ] Yes, I can help test
- [X] No, but I'd use it!

## Additional Context

Filed on behalf of a user. Implementation is mostly cosmetic — see `src/components/DXClusterPanel.jsx` around line 269 where the "de \<spotter\>" row is rendered.

accius:
Implemented on Staging in acbfb87.

The DX Cluster panel header now has a small **de** pill toggle. When off, the spotter sub-column is hidden and the grid template collapses from 4 columns to 3, giving roughly double the spot density in the same vertical space. Choice persists to \`ohc_dx_show_spotter\` in localStorage.

The toggle button uses \`aria-pressed\` so screen readers announce its state correctly (groundwork for #997).

Will ship with the next production release.

73 de K0CJH

PR #1051, opened 2026-06-02
fix: WCAG 2.1 AA accessibility pass — 40+ violations fixed across 14 files

w9fyi:
## Summary

This PR addresses **40+ WCAG 2.1 Level A and AA violations** found across the application, covering all the most critical failure categories. Every change is purely additive ARIA/semantic markup — **no visual appearance is changed**. The build continues to pass (`vite build ✓`).

### Critical (blocks keyboard or screen reader use)

- **Viewport zoom locked** — removed `user-scalable=no` / `maximum-scale=1.0` from viewport meta (WCAG 1.4.4)
- **No skip link** — added `<a href="#main-content">Skip to main content</a>` as first body child with off-screen CSS (WCAG 2.4.1)
- **Modal dialogs have no dialog semantics** — added `role="dialog" aria-modal="true" aria-labelledby` to SettingsPanel, KeybindingsPanel, and DonateButton modals (WCAG 4.1.2)
- **Interactive `<span>` / `<div>` elements not keyboard-operable** — converted to proper `<button>` or `<a>` elements:
  - Callsign, version number, and local clock in Header (`<span onClick>` → `<button>`)
  - QRZToggle (`<span onClick>` → `<button aria-pressed>`)
  - CallsignLink (`<span onClick>` opening QRZ.com → `<a href target="_blank">`)
  - WeatherPanel expand/collapse row (`<div onClick>` → `<button aria-expanded aria-controls>`)
  - DX cluster spot rows (`<div onClick>` → `role="button" tabIndex={0} onKeyDown`)
- **Form inputs with no label** — added `<label htmlFor>` to DXCluster quick-search and RigControl frequency input (WCAG 1.3.1)

### High (major screen reader degradation)

- **Icon-only buttons have no accessible name** — added `aria-label` to all mobile header icon buttons (settings, fullscreen), all SidebarMenu action buttons (mode cycle, layout lock, update, fullscreen, settings), and modal close buttons in KeybindingsPanel and DonateButton (WCAG 4.1.2)
- **Settings tabs have no ARIA tab pattern** — applied full `role="tablist"` / `role="tab"` + `aria-selected` / `role="tabpanel"` + `aria-labelledby` to all nine settings tabs (WCAG 4.1.2)
- **SVGs read by screen readers** — added `aria-hidden="true" focusable="false"` to all 17 icon components in Icons.jsx (WCAG 1.1.1)
- **Analog clock has no text equivalent** — added `role="img" aria-label` with live time string to the SVG clock face (WCAG 1.1.1)
- **Theme buttons have no selected-state announcement** — added `aria-pressed` + `aria-label` to ThemeSelector buttons (WCAG 4.1.2)
- **Status LED conveys state by color only** — added `role="status" aria-label` to RigControl LED (WCAG 1.4.1)
- **`<Header>` uses `<div>` not `<header>`** — changed outermost element to `<header>` landmark (WCAG 1.3.1)
- **DX news ticker infinite scroll with no reduced-motion guard** — added `@media (prefers-reduced-motion: reduce)` suppressing all 14 CSS animations site-wide (WCAG 2.3.3)

### Medium

- **No visible focus indicator** — added `:focus-visible` global rule with 2px amber outline (WCAG 2.4.7 / 2.4.11)
- **New-tab links no warning** — added `aria-label` with "(opens in new tab)" to all three donation links
- **DX cluster spot live region** — added `role="status" aria-live="polite"` offscreen element announcing spot count updates (WCAG 4.1.3)
- **Map-toggle button state** — added `aria-pressed` to DXCluster map-toggle button

### CSS additions (`main.css`)

```css
.skip-link       — off-screen skip navigation link, visible on focus
.sr-only         — utility class for visually-hidden accessible text
:focus-visible   — global focus ring (2px amber, outline-offset 2)
@media (prefers-reduced-motion: reduce)  — halts all 14 CSS animations
```

## Test plan

- [ ] Tab through the page from the skip link — verify focus lands on main content
- [ ] Tab to the callsign button in the header — verify VoiceOver/NVDA announces "Settings — {callsign}, button"
- [ ] Tab to the local clock — verify it is announced as a button with the format-toggle label
- [ ] Open Settings (`?` key or callsign click) — verify VoiceOver announces "Settings dialog"
- [ ] Tab through Settings tabs — verify each is announced as "tab, 1 of 9, selected/not selected"
- [ ] Tab to DXCluster search box — verify it is labeled "Search DX cluster spots by callsign"
- [ ] Tab to a DX spot row and press Enter — verify spot click fires
- [ ] Expand the weather panel with keyboard — verify `aria-expanded` toggles
- [ ] Enable system-level `prefers-reduced-motion: reduce` — verify the DX news ticker stops and all animations freeze
- [ ] Tab through the sidebar in icon-only mode — verify each button has a spoken label
- [ ] Check that all icon-only SVGs are skipped by screen readers (no "graphic" announcements)
- [ ] Open DevTools Accessibility panel — verify no `<div>`/`<span>` interactive elements remain in the tree

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,10 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta
-      name="viewport"
-      content="width=device-width, initial-scale=1.0, viewport-fit=cover"
-    />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <title>OpenHamClock - Amateur Radio Dashboard | Live DX Cluster, Propagation, Band Conditions</title>
 
     <!-- SEO Meta -->
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover"
+      content="width=device-width, initial-scale=1.0, viewport-fit=cover"
     />
     <title>OpenHamClock - Amateur Radio Dashboard | Live DX Cluster, Propagation, Band Conditions</title>
 
@@ -107,6 +107,7 @@
     <script src="/vendor/leaflet/leaflet.js"></script>
   </head>
   <body>
+    <a href="#main-content" class="skip-link">Skip to main content</a>
     <div id="root"></div>
     <script>
       // Detect stale JS chunks after an update and force-reload once.
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -668,7 +668,8 @@ const App = () => {
   }, []);
 
   return (
-    <div
+    <main
+      id="main-content"
       style={{
         width: '100vw',
         height: '100vh',
@@ -803,7 +804,7 @@ const App = () => {
         onClose={() => setShowWwbotaFilters(false)}
       />
       <WhatsNew showWhatsNew={config.showWhatsNew} />
-    </div>
+    </main>
   );
 };
 
--- a/src/components/AnalogClockPanel.jsx
+++ b/src/components/AnalogClockPanel.jsx
@@ -148,7 +148,13 @@ export const AnalogClockPanel = ({ currentTime, sunTimes }) => {
       </div>
 
       {/* Clock face */}
-      <svg width={size} height={size} style={{ flexShrink: 0 }}>
+      <svg
+        width={size}
+        height={size}
+        style={{ flexShrink: 0 }}
+        role="img"
+        aria-label={`Analog clock showing ${String(hours).padStart(2, '0')}:${String(minutes).padStart(2, '0')}:${String(seconds).padStart(2, '0')} local time`}
+      >
         {/* Outer ring */}
         <circle
           cx={center}
@@ -204,13 +210,17 @@ export const AnalogClockPanel = ({ currentTime, sunTimes }) => {
             fontSize: size > 150 ? '12px' : '10px',
           }}
         >
-          <span style={{ color: 'var(--accent-amber)' }}>
-            <span style={{ marginRight: '2px' }}>&#9788;</span>
+          <span style={{ color: 'var(--accent-amber)' }} aria-label={`Sunrise ${localSunrise}`}>
+            <span aria-hidden="true" style={{ marginRight: '2px' }}>
+              &#9788;
+            </span>
             {localSunrise}
           </span>
-          <span style={{ color: 'var(--accent-purple)' }}>
+          <span style={{ color: 'var(--accent-purple)' }} aria-label={`Sunset ${localSunset}`}>
             {localSunset}
-            <span style={{ marginLeft: '2px' }}>&#9790;</span>
+            <span aria-hidden="true" style={{ marginLeft: '2px' }}>
+              &#9790;
+            </span>
           </span>
         </div>
       )}
--- a/src/components/CallsignLink.jsx
+++ b/src/components/CallsignLink.jsx
@@ -65,23 +65,32 @@ export function useQRZ() {
 export function QRZToggle({ style }) {
   const { enabled, toggle } = useQRZ();
   return (
-    <span
+    <button
+      type="button"
       onClick={(e) => {
         e.stopPropagation();
         toggle();
       }}
       title={enabled ? 'Click callsigns to open the callbook lookup (ON)' : 'Callsign links disabled (OFF)'}
+      aria-label={
+        enabled ? 'QRZ callsign links enabled — click to disable' : 'QRZ callsign links disabled — click to enable'
+      }
+      aria-pressed={enabled}
       style={{
         cursor: 'pointer',
         fontSize: '11px',
         opacity: enabled ? 1 : 0.4,
         userSelect: 'none',
         transition: 'opacity 0.2s',
+        background: 'none',
+        border: 'none',
+        padding: 0,
+        lineHeight: 1,
         ...style,
       }}
     >
-      🔍
-    </span>
+      <span aria-hidden="true">🔍</span>
+    </button>
   );
 }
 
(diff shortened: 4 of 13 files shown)


accius (changes requested):
Thanks for the WCAG pass — there are real wins in here (viewport unlock, semantic `<header>`, real `<a target=\"_blank\">` for QRZ links, dialog/aria-modal on modals, WeatherPanel aria-expanded/controls). A few things to clean up before this can land:

### Blockers

**1. `.github/workflows/watch-upstream-pr.yml` doesn't belong in this PR.** That workflow targets `--repo w9fyi/openhamclock` and watches PR #1051 from your fork — it's personal automation for tracking your own PR, not something we want running in the upstream repo every 6 hours. Please drop the file (and the chore commit that added it) before merge.

**2. DXClusterPanel — `role=\"button\"` on spot rows clobbers `role=\"row\"`.** PR #1001 landed on 2026-05-27 and added full ARIA table semantics (`role=\"table\"` / `role=\"row\"` / `role=\"columnheader\"` / `role=\"cell\"`) to these exact `<div>`s so screen-reader users get column navigation across callsign/freq/mode/age. Your change overwrites `role=\"row\"` with `role=\"button\"`, which collapses every spot into a single labeled control and removes the column navigation. Pick one model: either keep the table semantics from #1001 (and drop `role=\"button\"`, keeping just the `onKeyDown` for Enter/Space — which is fine on a `role=\"row\"`), or replace the table model deliberately and explain why. I'd lean toward keeping the table — that was the considered choice and the spot density (#995) is real value.

**3. DXClusterPanel new `role=\"status\" aria-live=\"polite\"` region.** This announces \"X spots loaded, latest: CALL on FREQ\" on every spot arrival. On a busy band that's easily 60–100+ announcements per minute and will be unbearable for a screen-reader user. Either remove the live region entirely or debounce announcements to ≥10s and only say \"N new spots\" rather than the full latest-spot detail.

**4. SettingsPanel tablist work duplicates PR #1045** (squash-merged to Staging on 2026-06-01). #1045 implemented the full W3C APG tablist pattern across all 7 tab strips including SettingsPanel, with a shared `ariaTabKeyDown` helper for arrow/Home/End roving-tabIndex navigation. Your version of the SettingsPanel tablist adds the roles but is missing the keyboard navigation — please rebase onto current Staging so the merge conflict surfaces and we can keep the #1045 keyboard behavior. (Heads-up: this branch is based on May 7 commits and is missing #997 / #1000-1003 / #1037 / #1045 — a rebase will also surface the table-row conflict in DXClusterPanel above.)

**5. Skip link target `#main-content` doesn't exist.** I grepped App.jsx / DockableApp.jsx / SidebarMenu.jsx and nothing has `id=\"main-content\"`, so the skip link will focus an empty anchor. Wire it to a real target — probably the main map container or the root `<main>` (which would also let you swap the wrapper `<div>` for `<main>` and pick up another landmark for free).

### Nits

- **`.sr-only` duplicates `.visually-hidden`** which already exists in main.css (added in #1001 / 33ed861 with the standard clip-rect pattern). Use the existing class in JSX instead of inline visually-hidden styles (e.g., the new `<label htmlFor=\"dx-cluster-search\">` block).
- **Duplicate `:focus-visible` rule.** main.css already has a global `:focus-visible` block at line 13 from PR #997 (f2d292c). Your new block at the end of the file overrides it with a different outline width. Either drop the new one or replace the original — but having two is confusing.
- **Removed `title=` attributes** on QRZToggle, Header callsign/version/clock, and Header fullscreen button. Sighted mouse users lose the hover tooltip. `aria-label` is for AT only; keep both — `title` AND `aria-label` — so both audiences get the affordance.

### Real wins to call out

These are clean — keep them as-is:

- `index.html` viewport meta — dropping `user-scalable=no` / `maximum-scale=1.0` is a clear WCAG 1.4.4 win
- CallsignLink rewrite — going from `<span onClick>` + `window.open` to a real `<a href target=\"_blank\" rel=\"noopener noreferrer\">` means middle-click / cmd-click / right-click \"open in new tab\" now all work the way users expect
- Header `<div>` → `<header>` landmark
- Modal `role=\"dialog\"` + `aria-modal=\"true\"` + `aria-labelledby` on Settings/Keybindings/Donate
- WeatherPanel summary-row `<div onClick>` → `<button>` with `aria-expanded`/`aria-controls`
- Form `<label htmlFor>` on DXCluster search + RigControl frequency

Once the five blockers are addressed, this is a great improvement. Happy to re-review after the rebase.

K0CJH

w9fyi:
Thanks for the thorough review, K0CJH. All five blockers addressed in the latest push — here's a summary:

**Blocker 1 — Workflow file**: Dropped the `watch-upstream-pr.yml` file and its commit entirely. The branch is now a clean single-commit rebase onto current Staging.

**Blocker 2 — DXClusterPanel `role="button"`**: Resolved conflict in favour of #1001's table semantics (`role="row"` / `role="cell"` / etc.). Kept the `onKeyDown` handler (Enter/Space activates `onSpotClick`) — this is valid on `role="row"`. Also preserved the `type="button"` fix on the map-toggle button and kept both `title=` and `aria-label` + `aria-pressed` on the spotter-toggle from #1003.

**Blocker 3 — DXCluster `role="status"` live region**: Removed entirely. #1000 already handles connection-status announcements; spot-arrival churn on a busy band was exactly the "unbearable" case you described.

**Blocker 4 — SettingsPanel rebase**: Rebased onto current Staging. Took #1045's full W3C APG tablist pattern (roving `tabIndex`, `ref`-driven focus management, `ariaTabKeyDown`) in every conflict. Our version's roles-without-keyboard-nav is gone.

**Blocker 5 — Skip link `#main-content`**: Wired. Changed the root `<div>` in `App.jsx` to `<main id="main-content">` — skip link now focuses a real target and the page picks up a `<main>` landmark for free.

**Nits**:
- Replaced both inline visually-hidden style blocks (DXClusterPanel `<label>` and RigControlPanel `<label>`) with `className="visually-hidden"`, using the existing #1001 class.
- Removed the duplicate `.sr-only` definition (never referenced in JSX) and the duplicate `:focus-visible` block that was overriding #997's cyan ring with amber.
- `title=` attributes are preserved alongside `aria-label` on QRZToggle, Header settings button, and Header fullscreen button — HEAD already had all three from #997/#1003; our original commit had incorrectly removed them, fixed in this rebase.

`vite build ✓` on the rebased branch.

accius:
CI's flagging one tiny prettier diff on `index.html`. Your `<meta name="viewport">` got split across multiple lines, project prettier config wants it on one. `npx prettier --write index.html` and re-push should green it.

Walked through the code already and it's looking really good (all the June 2 blockers are addressed, big bonus wins on reduced-motion + skip-link), so once that one line is sorted I'll approve and we can land it.

accius (approved):
Huge progress on this since last review. Walking through the five blockers from June 2:

1. **Workflow file**: gone from the branch.
2. **DXCluster `role="button"` clobbering `role="row"`**: fixed. Dropped `role="button"` and kept just `onKeyDown` for Enter/Space, which is exactly the right call. The `role="row"` table semantics from #1001 are preserved.
3. **Chatty per-spot aria-live region**: gone.
4. **Rebase onto current Staging**: done (base is now b20d5b7). SettingsPanel changes dropped entirely, which neatly sidesteps the duplicate-with-#1045 problem. Smart.
5. **Skip link target `#main-content`**: wired up by swapping the App.jsx wrapper `<div>` for `<main id="main-content">`, which also gives us a real landmark for free. Exactly what I'd hoped.

Bonus wins I want to call out:
- Reduced-motion media query is a great WCAG 2.3.3 add. Halting the WSPR animations, DX news ticker, and loading spinner under `prefers-reduced-motion` is exactly right.
- `type="button"` on buttons that weren't explicitly typed. Prevents accidental form submits, good defensive practice.
- `<a href>` replacing `<span onClick>` for the donate flow keeps middle-click and cmd-click working.
- Skip-link CSS uses the off-screen-then-position:fixed-on-focus pattern, which is the actual usable skip-link pattern (not just visually-hidden the whole time).

One small thing didn't get addressed and I still think it's worth doing, but it's a nit not a blocker:

The `title=` attribute got removed from a lot of icon-only buttons (status LED, modal close, sidebar menu items, theme selector, layout lock, fullscreen, settings, donate, CallsignLink). The `aria-label` is enough for AT users, but sighted mouse users lose the hover tooltip. For icon-only buttons especially, the tooltip is often the only way a new user discovers what the icon does. The map toggle in DXClusterPanel got it right, keeping both `title=` and `aria-label=`. Worth restoring `title=` alongside `aria-label=` on the rest, either here or as a quick follow-up.

Otherwise this is a great improvement and good to go. Approving so it can land.

PR #1058, opened 2026-06-02
fix(windows): add update.ps1, improve setup UX, and fix PS 5.1 parse errors

ceotjoe:
## Summary

- Adds `scripts/update.ps1` — a Windows equivalent of `update.sh` (git pull + npm ci + build + .env restore) so Windows users have a proper update path
- Adds explicit **Windows install section** to README using a `iwr` one-liner, mirroring the Linux/macOS curl pattern
- Clarifies README update headings: "Linux / macOS / Pi" vs "Windows" as clear peers
- Fixes `setup-windows.ps1` update path to run `npm ci` + `npm run build` (previously skipped the build step)
- Fixes `start.bat` template to check for `node_modules` on launch and self-heal after a bare `git pull`
- Fixes PS 5.1 parse errors in both `.ps1` scripts caused by non-ASCII characters (box-drawing chars, checkmarks, emoji) being misread as CP1252 string terminators on Western European Windows systems

## Background

Issue report in Facebook regarding Windows setup

Triggered by a v26.4.0 crash report on Windows where `convert-csv-to-json` (added in the satellite PR) was missing because the user had no documented update path and did a bare `git pull` without `npm install`.

## Test plan

- [ ] Run `iex (iwr https://raw.githubusercontent.com/accius/openhamclock/main/scripts/setup-windows.ps1).Content` on a clean Windows 11 machine
- [ ] Run `.\scripts\update.ps1` from an existing install after a `git pull`
- [ ] Verify `start.bat` auto-runs `npm install` if `node_modules` is missing
- [ ] Verify both `.ps1` scripts parse without errors on PowerShell 5.1 (Windows 11 default)


accius (approved):
Nice to finally have a proper Windows update path. update.ps1 follows the same backup → pull → install → build → restore dance as update.sh, defensive checks at the top, version diff at the bottom. The "73 de OpenHamClock" sign-off is a nice touch.

Couple of small observations (not blocking):

- `Setup-Repository` in setup-windows.ps1 now duplicates the update logic that update.ps1 owns. The "directory exists" branch in the setup script becomes kind of vestigial once people have both scripts. Could be a future cleanup to make setup-windows.ps1 hand off to update.ps1 in that case.
- The `.env.backup` and `config.json.backup` files get left in the install dir after a successful update. Confirmed update.sh does the same thing (lines 146-159, 253-259), so it's consistent, but both scripts could clean those up at the end since they're only there as a safety net during the run.
- Curious which specific PS 5.1 parse errors the box-drawing-character swap fixed. Was it the file encoding without BOM, the host code-page rendering, or something else? Useful to know in case it bites us elsewhere.

ASCII install/update pattern matches the curl one-liner shape Linux/macOS users already follow, which is the right call. README split into "Linux / macOS / Pi" vs "Windows" peers is also a clarity win.

PR #1060, opened 2026-06-02
[fix] error __satWinToggleMinimize remove unnecessary onclick

MichaelWheeley:
## What does this PR do?

- message = `?log=debug:1 Uncaught TypeError: window.__satWinToggleMinimize is not a function
    at HTMLButtonElement.onclick (?log=debug:1:8)`
- remove unnecessary onclick

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test
1.
2.
3.

## Checklist


accius (approved):
Confirmed via `addMinimizeToggle.js:50-73`, the click event is bound there with `manageButtonEvents: true`, which queries the button by class and attaches its own listener. So the inline `onclick` was both broken (`window.__satWinToggleMinimize` was never exposed) and redundant. Clean fix.

PR #1063, opened 2026-06-04
[refactor] satellite data, fletcher, TLE handling

MichaelWheeley:
## What does this PR do?

satellite data refactor for use of `fletcher` (was `tle-fetcher`), and general improvements to TLE handling and utilities
- add note explaining choice of 24 hour OMM_CACHE_DURATION
- reduce unnecessary global variables to improve readability
- rename functions as TLE where appropriate
- add questionable data comments
- remove exports for TLE utilities except where needed
- apply previously developed state machine for TLE
- add unit test for server/utils/tle-to-omm
- renamed tle-fetcher as fletcher, reason = HTTP relay is not related to TLE/3LE data format
- project .env TLE_FETCHER_URL change to FLETCHER_URL
- tleFetcherUrl to fletcherUrl
- remove debug from parse TLE which was cluttering server logs

misc note, was not able to use the main project `User-Agent` as the version number is not available to the `Fletcher` service, therefore have renamed as hard-coded,
`'User-Agent': inboundUserAgent || 'OpenHamClock-Fletcher/1.0',`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test
.env requires `FLETCHER_URL` (was `TLE_FETCHER_URL`)

## Checklist


accius (changes requested):
Bunch of small stuff, nothing dramatic.

**On the function renames.** Honestly might just be me reading them wrong, you've spent way more time in this corner of the code than I have. But `fetchTleFromAmsat` had me double-taking because the body still returns the OMM-shaped object (`ommArray`, `omm`) and callers destructure those names. Felt like the TLE in the name describes what's on the wire from upstream, but the OMM in the variables describes what the function actually hands back. Wondered if it'd be cleaner to either keep the old `fetchOmm*` names, or rename the returned vars to match too. Curious what your thinking was.

**Actually a real bug at satellites.js:664:**
```js
sat.backoffSatnogsUntil = now + +60 * 60 * 1000; // 1 hour
```
The `+ +` evaluates fine since `+60` is unary, but I'm betting it's a search-replace leftover from `+ SATNOGS_BACKOFF`. Worth tidying.

**Few nits while you're in there:**
- Couple typos: `simulute` in the tle-to-omm.js header, and `they is little accuracy benefit` in the OMM_CACHE_DURATION comment. Also the "50% on a chained distribution" line in that comment went over my head, what does that mean exactly? Previous version ("matches NORAD's daily TLE release cadence") landed clearer for me.
- The commented-out `debugStr` block in the new `parseTleBlock` emit callback. Either commit it clean or leave a TODO so future-you remembers why.
- The new `TleLineParserStateMachine`. Is there a planned second caller or another TLE-shaped feed coming? Asking because the test cases pass against the original 10-line loop too, so the 70-line class machinery is doing roughly the same observable work right now. Totally fine if you're laying groundwork, just curious.

**Stuff I really like:**
- The unit tests for `parseTleBlock`. Big win, locks behavior down for anyone touching this file later.
- Removing the unused exports is fine, confirmed nothing else imports `parseImpliedDecimal` / `parseExpField` / `epochToIso`.
- `logWarn` on AMSAT/SatNOGS backoff trips is genuinely useful when something's going sideways live.
- "Accuracy of data questionable" on SatNOGS, saves a future maintainer a head-scratch.

Once the `+ +` and the typos are sorted, plus whatever you decide on the function names, I think this is good to go.

Side note on the rename. I already prepped Staging for it. Railway service is renamed, private domain is renamed, both `FLETCHER_URL` and `TLE_FETCHER_URL` env vars point at `fletcher.railway.internal:8080` so nothing breaks during the swap. Only remaining dashboard step at merge time is flipping the service rootDirectory from `tle-fetcher` to `fletcher`. About a 30s window where the auto-deploy fails until that's clicked, just so you know.

MichaelWheeley:
> The new TleLineParserStateMachine. Is there a planned second caller or another TLE-shaped feed coming? Asking because the test cases pass against the original 10-line loop too, so the 70-line class machinery is doing roughly the same observable work right now. Totally fine if you're laying groundwork, just curious.

Yes, the algorithm was developed previously to handle both TLE and 3LE formats but was never merged into mainline since TLE/3LE formats were replaced by OMM

accius (approved):
Reading it again with the rename suffix in mind, `fetchOmmFromAmsat_TleBased` and the AMSAT/SatNOGS twins make more sense to me now. The name carries both axes (return shape: OMM, upstream wire shape: TLE-derived) without losing either. The OMM_CACHE_DURATION analysis comment is more than I'd have written but it explains the choice cleanly and won't hurt the reader.

State machine: I get it now after your reply. Reusable groundwork for the TLE/3LE path is fine to land even if the immediate observable behaviour matches the previous loop. The tests do cover the corrupt-block and no-name-line branches that the old loop wouldn't have caught cleanly.

One deploy thing before we merge so we don't accidentally route around fletcher: the env var rename `TLE_FETCHER_URL → FLETCHER_URL` means I need to set `FLETCHER_URL` on the `openhamclock-staging` Railway service before this lands (and on `openhamclock-production` when we cut the next release). If we forget, the satellite routes silently fall back to direct upstream and CelesTrak goes quiet again. I'll handle that side. The Railway private domain rename `tle-fetcher.railway.internal → fletcher.railway.internal` is the matching step on the fletcher service itself.

Approving.

73 de K0CJH

PR #1064, opened 2026-06-05
fix: restore title= hover tooltips on icon-only buttons (#1051 follow-up)

w9fyi:
## Summary

Follow-up to #1051 (WCAG 2.1 AA accessibility pass). As @accius noted in the review, the PR added `aria-label` to all icon-only interactive elements but removed the pre-existing `title=` attributes in the process. `aria-label` is sufficient for AT users, but sighted mouse users rely on the hover tooltip to discover what icon-only controls do — especially first-time users who don't know what each icon means.

The `DXClusterPanel` map toggle was already correct (kept both `title=` and `aria-label=`). This PR brings everything else in line with that pattern.

**Files changed (6):**

| File | Buttons fixed |
|------|--------------|
| `CallsignLink.jsx` | Callbook lookup `<a>` |
| `DonateButton.jsx` | Support trigger + modal close |
| `KeybindingsPanel.jsx` | Modal close |
| `RigControlPanel.jsx` | Status LED |
| `SidebarMenu.jsx` | Mode toggle, 9 nav items, layout lock, reset layout, update, fullscreen, settings |
| `ThemeSelector.jsx` | All theme selector buttons |

Every `title=` added matches the corresponding `aria-label` (or the original tooltip text where they differ slightly). No visual changes; build passes `vite build ✓`.

## Test plan

- [ ] Hover over each sidebar icon-only button — verify tooltip appears
- [ ] Hover over the status LED in RigControl — verify connection state tooltip
- [ ] Hover over a callsign link — verify "Look up {call} in callbook" tooltip
- [ ] Hover over the Support button and its modal close — verify tooltips
- [ ] Hover over the Keybindings modal close — verify "Close" tooltip
- [ ] Hover over each ThemeSelector button — verify "{Theme} theme" tooltip
- [ ] Confirm DXClusterPanel map toggle still has tooltip (regression check)

### DIFF
--- a/src/components/CallsignLink.jsx
+++ b/src/components/CallsignLink.jsx
@@ -118,6 +118,7 @@ export default function CallsignLink({
         rel="noopener noreferrer"
         onClick={(e) => e.stopPropagation()}
         aria-label={`Look up ${call} in callbook (opens in new tab)`}
+        title={`Look up ${call} in callbook`}
         style={{
           color,
           fontWeight,
--- a/src/components/DonateButton.jsx
+++ b/src/components/DonateButton.jsx
@@ -29,6 +29,7 @@ export default function DonateButton({ compact = false, fontSize = '12px', paddi
         type="button"
         onClick={() => setOpen(true)}
         aria-label="Support OpenHamClock"
+        title="Support OpenHamClock"
         tabIndex={tabIndex}
         style={{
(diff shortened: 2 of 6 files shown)


accius (approved):
Solid follow-up. Title attributes restored exactly where the #1051 review flagged them, and the few short-vs-long title/aria-label pairs are the right call for tooltip readability. Approving as-is.

73 de K0CJH

PR #1066, opened 2026-06-05
Feature map rotation

AnthonyOHC:
## What does this PR do?

Adds optional rotation controls for the main map/base-layer selector.

This lets the user rotate between selected base maps automatically, choose the rotation interval, or turn rotation off entirely. The controls live under a compact Map Rotation dropdown beside the existing base map selector.

## Type of change

* [ ] Bug fix
* [x] New feature
* [ ] Performance improvement
* [ ] Refactor / code cleanup
* [ ] Documentation
* [ ] Translation
* [ ] Map layer plugin

## How to test

1. Run the app and open the main map view.
2. Open the Map Rotation dropdown beside the base map selector.
3. Toggle Rotation On.
4. Select a rotation interval.
5. Choose which base maps are included using All/None or individual checkboxes.
6. Confirm the base map changes automatically after the selected interval.
7. Manually select a base map from the main map selector and confirm rotation turns off.
8. Leave the Map Rotation dropdown open and confirm it auto-closes after about 25 seconds of inactivity.

## Checklist

* [x] Feature is scoped to the existing base map selector.
* [x] Rotation can be turned on or off.
* [x] Rotation interval is selectable.
* [x] Base maps included in rotation are selectable.
* [x] Preferences are stored locally in browser localStorage.
* [x] Build completed successfully with `npm run build`.

### DIFF
--- a/src/components/WorldMap.jsx
+++ b/src/components/WorldMap.jsx
@@ -578,8 +578,111 @@ export const WorldMap = ({
   const initialStyle = MAP_STYLES[migratedStyle] && !MAP_STYLES[migratedStyle].legacy ? migratedStyle : 'dark';
   const initialProjection = storedSettings.isAzimuthal ? 'azimuthal' : storedSettings.mapProjection || 'mercator';
   const [mapStyle, setMapStyle] = useState(initialStyle);
+  const [mapRotationConfig, setMapRotationConfig] = useState(() => {
+    try {
+      const saved = JSON.parse(localStorage.getItem('ohc_main_map_rotator_v1') || '{}');
+      const allStyleIds = Object.keys(MAP_STYLES).filter((id) => MAP_STYLES[id] && !MAP_STYLES[id].legacy);
+      return {
+        enabled: saved.enabled === true,
+        intervalSeconds: Number(saved.intervalSeconds) || 60,
+        selectedIds: Array.isArray(saved.selectedIds) && saved.selectedIds.length ? saved.selectedIds : allStyleIds,
+      };
+    } catch {
+      const allStyleIds = Object.keys(MAP_STYLES).filter((id) => MAP_STYLES[id] && !MAP_STYLES[id].legacy);
+      return { enabled: false, intervalSeconds: 60, selectedIds: allStyleIds };
+    }
+  });
+  const [showMapRotationMenu, setShowMapRotationMenu] = useState(false);
+  const [mapRotationMenuActivity, setMapRotationMenuActivity] = useState(0);
   const [mapProjection, setMapProjection] = useState(initialProjection);
   const isAzimuthal = mapProjection === 'azimuthal';
+
+  const availableBaseMapIds = useMemo(
+    () => Object.keys(MAP_STYLES).filter((id) => MAP_STYLES[id] && !MAP_STYLES[id].legacy),
+    [],
+  );
+
+  useEffect(() => {
+    try {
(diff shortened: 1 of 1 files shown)


accius (changes requested):
Welcome aboard, and good first PR. The feature is well-scoped and the code follows the existing patterns in `WorldMap.jsx`, which is a chunky file to land your first change in. Two things need a fix before this can merge to Staging, plus a couple of smaller polish items.

### 1. Rebase to clean up the commit history

The PR is sitting on top of an older `main` (pre-v26.3.3), so GitHub is showing 12 commits in this PR but only one of them (`1fd1073`, your feature) is actually yours. The other 11 are old release merges that aren't in our current `Staging` branch. The net file diff is still clean (just `WorldMap.jsx` +309/-0), but if we merge as-is, all those old commits become part of Staging's history and the log gets noisy.

Here's how to clean it up:

```bash
# Add the upstream remote if you haven't already
git remote add upstream https://github.com/accius/openhamclock.git

# Get the latest Staging
git fetch upstream

# Switch to your feature branch
git checkout feature-map-rotation

# Rebase your one commit on top of current Staging
git rebase upstream/Staging

# If you hit a conflict in WorldMap.jsx during the rebase, open the file,
# keep your additions alongside whatever Staging has, then:
#   git add src/components/WorldMap.jsx
#   git rebase --continue

# Push the rebased branch
git push --force-with-lease origin feature-map-rotation
```

`--force-with-lease` is safer than `--force` since it bails if someone else pushed to your branch in the meantime.

After that, the PR should show 1 commit with the same diff.

### 2. Wire manual map selection to turn rotation off

Your test plan step 7 says manual selection should turn rotation off, but the existing code doesn't do that yet. The base map `<select>` at `src/components/WorldMap.jsx:2588-2603` just calls `setMapStyle(e.target.value)`, so the rotation interval keeps running. On the next tick, your effect sees the user's manual pick isn't in the rotation set (`indexOf` returns -1) and snaps to `selectedIds[0]`, blowing away the manual choice.

Change the onChange at line 2591 from:

```jsx
onChange={(e) => setMapStyle(e.target.value)}
```

to:

```jsx
onChange={(e) => {
  setMapStyle(e.target.value);
  if (mapRotationConfig.enabled) {
    setMainMapRotation({ enabled: false });
  }
}}
```

That way a manual override actually stops the rotation, matching your test plan.

### 3. Smaller polish items

**Theme colors on the rotation button.** Inside the dropdown you used `var(--bg-secondary)`, `var(--text-primary)`, etc., which adapt to the user's theme. The toggle button itself uses hardcoded teal (`#00ffd0`, `#001a14`, `rgba(0, 255, 170, 0.95)`). Looks fine in dark mode but will jump out in light/retro themes. Worth swapping to `var(--accent-color)` (or whichever theme variable is closest) for consistency.

**a11y.** The icon button has `title="Map rotation controls"` but no `aria-label`. Screen readers don't always announce `title`, so add `aria-label="Map rotation controls"` for reliable announcement. The `aria-hidden="true"` on the SVG is correct.

That's it. Once 1 and 2 are done, this is mergeable. The items in 3 are good-to-have but I won't block on them if you'd rather land follow-ups.

73 de K0CJH

accius (approved):
Nice turnaround. Rebase is clean, the manual-select wiring matches the test plan exactly, and the button is now theme-aware and has an aria-label. Approving and flipping on auto-merge so this lands as soon as CI goes green. Welcome to the project.

73 de K0CJH

PR #1068, opened 2026-06-05
feat: callsign lookup popup

### LINKED ISSUE #1016: [FEATURE] Ability for quick callsign lookups inside OHC
Currently clicking a callsign opens an external website lookup.
This is nice to get as much information on the user as possible, but a bit overkill to just get some basic information, while also taking the user out of the OHC tab.

I believe it would be nice to be able to get some of the basic user information right in the OHC UI.

A related feature would be the ability to do ad-hoc callsign lookups/searches using a search box of some sort (maybe even callable with keyboard shortcut).

The best way to display this is unclear and input would be appreciated.

My current ideas are either:
- Clicking (or hovering?) the callsign shows a popup similar to when hovering over a contact in the map:

Image

- Have a dedicated map collapsible floating window similar to the one used by map layers or even the band legend. The user could potentially move it around to their preference and hide it when not needed.

As for what information to display, my current candidates are:

- Name
- Country
- Grid Locator
- A button to open the profile in the appropriate website (QRZ/HamQTH/etc..)

I had a quick chat with @alanhargreaves and he also suggested the following:

- Instead of clicking, maybe just hovering on the callsign could trigger the popup similar to the map popups.
- Add a server-side cache system so we dont hit the callbook websites for repeat lookups (and depending on the information we display, its likely stuff that does not need updating regularly)

## Use Case

- [x] General HF operation
- [x] VHF/UHF operation
- [x] Satellite operation
- [x] Contesting
- [x] POTA/SOTA
- [x] DXing

- [ ] Yes, I can submit a PR
- [x] Yes, I can help test
- [ ] No, but I'd use it!

## Additional Context

I will attempt to create some mockups soon of what it could look like.

TBD

---

73!

lbatalha:
## What does this PR do?

Should resolve #1016 

### Features

 - Callsign info popup: Replace tab-opening behavior with a floating info popup showing name, grid, country, local time, and a globe link to the configured callbook
 - Local time display: Show 24h formatted local time next to the callsign using grid/lat-lon timezone API with LRU cache (24h TTL, max 500 entries)
 - DX panel shared timezone cache: Reuse useTimezone hook across popup and DX panel to eliminate redundant geo-time fetches
 - Clickable callsigns in WorldMap popups

### Bug Fixes

 - HamQTH DXCC JSON migration: Replaced XML dxcc.php with JSON dxcc_json.php, added response validation and retry logic, 2s timeout for faster fallback
 - Cache poisoning fix: Migrated wsjtx.js and dxcluster.js from direct dxcc.php calls to hamqthLookup(), eliminating shared cache corruption from flaky XML API
 - Empty country fix: HamQTH DXCC in wsjtx.js and dxcluster.js now sets country: 'Unknown' instead of empty string on missing name to match callsign.js impl
 - Popup animation: Fixed initial height estimation and expanded animation to prevent content flashing/reflow
 - Client-side cache: Added LRU eviction (500 entries max) to callsign lookup cache

### Cleanup

 - Removed QRZ toggle and legacy callbook tab feature
 - Standardized source field values ('qrz', 'hamqth-xml', 'hamqth-dxcc', 'prefix', 'cty-dat')

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Click any spot callsign in any panel 
2. See that popup shows up and tries to load callbook data
3. Check that timezone is correct (should usually resolve to the spot location in all panels)
4. Check that caching works by opening the same callsign again and noticing there is no loading delay for the data

## Checklist

Popup with API error:
image

Popup with just basic DXCC info:
image

Popup with extra callbook data:
image

(the tooltips are the mouse hovering the local time display, its used to debug or just show info on the way the timezone was calculated)


accius (changes requested):
Reviewed the full diff. The feature is wired end-to-end cleanly and the security surface holds up (more on that below), but there is one real regression I want fixed before this lands, plus a few correctness nits in the new hooks.

### Blocking

1. **`server/routes/wsjtx.js:665` regresses the WSJT-X background lookup throttle.** The refactor to `hamqthLookup()` dropped the `.finally()` cleanup that the base branch had at the old lines 689-690:

   ```js
   wsjtxHamqthInflight.add(targetCall);
   hamqthLookup(targetCall).then((result) => {
     if (result) { cacheCallsignLookup(targetCall, { data: result, timestamp: Date.now() }); }
   });
   ```

   `wsjtxHamqthInflight` (line 368) is a long-lived Set, and nothing ever deletes from it now. The guard on line 663 is `!wsjtxHamqthInflight.has(targetCall) && wsjtxHamqthInflight.size < 5`. Once five unique callsigns have been seen the set is permanently full, `size < 5` is forever false, and **all** WSJT-X background HamQTH grid lookups stop for the life of the process. Those five calls also can never be re-looked-up. Restore the cleanup:

   ```js
   hamqthLookup(targetCall)
     .then((result) => { if (result) cacheCallsignLookup(...); })
     .finally(() => wsjtxHamqthInflight.delete(targetCall));
   ```

### Non-blocking, would be good to fix in this PR

2. **`src/hooks/app/useTimezone.js` `mountedRef` is stuck false after the first grid change.** The mount effect sets `mountedRef.current = true` with `[]` deps, but the per-`grid` effect's cleanup sets it to `false` on every grid change, not just unmount. After the first change the ref never goes back true, so `setLocalTime` is gated off for all later grids. In `CallsignPopup` the grid commonly changes mid-life (cty centroid grid, then callbook lat/lon grid), so the local time can stick to the first grid's zone. Usually the same zone so invisible, wrong for large multi-zone entities. The `AbortController` cleanup already prevents stale `setState`, so you can drop the `mountedRef` gating, or set it true at the top of the per-grid effect.

3. **`src/hooks/app/useCallsignLookup.js` drops lookups when `call` changes mid-flight.** `fetchInProgress.current` short-circuits the new fetch, and there is no abort tied to `call` change, and `data` is not reset, so a stale entry shows briefly. The popup remounts on each click so this is mostly masked today, but it bites any reuse path (programmatic `showPopup`, WorldMap). Prefer abort-on-change with a last-write-wins guard over the in-progress ref.

4. **`src/components/CallsignPopup.jsx` double-escapes via `esc()` in JSX.** `esc()` from `escapeHtml.js` is for raw-HTML / Leaflet string contexts. Here the values are JSX children and attributes (`{esc(call)}`, `title={esc(error)}`), which React already escapes, so "Trinidad & Tobago" renders as "Trinidad &amp; Tobago" and "O'Brien" as "O&#39;Brien". Drop `esc()` on the JSX values.

5. `CallsignPopup.jsx:8` comment says "Auto-dismisses after 15s" but there is no timer. Add it or fix the comment.

6. The shared `hamqthLookup()` no longer sends the `User-Agent: OpenHamClock/<ver>` header the old `dxcc.php` callers used. Minor good-neighbor item given the fleet size hitting HamQTH, worth adding back inside `hamqthLookup`.

### Security: clean

I specifically checked the new `/api/callsign/:call` route and the background callers. Every external URL interpolates the callsign through `encodeURIComponent`, and the route validates `^[A-Z0-9\/\-]{1,20}$` before any fetch, so no SSRF or URL injection. Hosts are hardcoded. QRZ/HamQTH credentials are written `0o600`, never returned by the status endpoints, and the configure endpoints sit behind `writeLimiter` + `requireWriteAuth`. The deleted `.github/workflows/docker-dxspider-proxy 2.yml` is a macOS "copy 2" duplicate of the still-present `docker-dxspider-proxy.yml`, so no CI is lost.

### What is good

- The HamQTH DXCC XML to JSON migration is a genuine fix. The old `dxcc.php` regex had mismatched `<n>`/`</name>` tags and could never match a country name. `dxcc_json.php` with response validation and a one-shot retry is correct.
- Server-side cache hygiene is careful: insert-time cap, 30-minute TTL sweep, oldest-first eviction over cap, and the client LRU re-inserts on access for true ordering.
- The popup replaces the old tab-opening behavior everywhere (DXCluster, PSKReporter, APRS, MeshCom, Activate, DXpedition, WorldMap) with proper keyboard and outside-click handling.

Fix 1 and I will re-review promptly. 2 through 6 I will leave to your judgement but 2 and 4 are user-visible.

K0CJH

accius (approved):
The follow-ups land where I wanted them: inflight set now gets cleaned up in the finally so the background lookup regression is gone, User-Agent is back on the HamQTH calls, and dropping esc() in JSX is right since React handles the escaping. The mountedRef removal is also correct now that the cache is module-level.

One blocker left that isn't on me: this branch has merge conflicts with Staging after today's merges. Resolve those and it can land.

K0CJH

lbatalha:
Fixed merge conflicts, just a few seconds before you responded :D

Issue #1002, opened 2026-05-19
[Accessibility] Non-map alternative view of map data for screen-reader users


accius:
Sub-issue of #997 (accessibility umbrella).

## Problem

The OHC main view is map-centric — DX spots, satellites, lightning, aurora, aircraft, gateways are all visually plotted. None of this is meaningful to a totally blind operator.

## What we want

A toggleable "list view" of map data that's screen-reader-first. Doesn't need to replace the map — sits alongside it. Format:

```
DX SPOTS — 30 visible
1. 14.074 MHz — JA1XYZ (Japan) — bearing 320°, 10,750 km — 2 min ago
2. 7.030 MHz — VK4LWH (Australia) — bearing 254°, 14,200 km — 4 min ago
…

SATELLITES OVERHEAD
1. ISS — elevation 23°, azimuth 178°, working window 6 min
…

LIGHTNING — 3 strikes within 100 km
1. 42 km SE — 30 s ago — +15 kA
…
```

Render as a real `<table>` per data source (relates to the spot-table sub-issue) with headings the screen reader can navigate by column. Each layer the user has *enabled* gets a section.

## Scope

A new panel (probably `MapDataListView.jsx` or similar) that subscribes to the same hooks the map layers use and renders them in textual form. Most data is already in the hooks — this is presentation work, not data-fetching work.

## Tester wanted

This one especially benefits from real screen-reader user feedback on what's useful vs. noise.

accius:
First pass shipped in dec0a23 on Staging. New **"Map Data (text view)"** dockable panel — pulls the same data the map plots, renders it as semantic tables with `<thead>` / `scope="col"` so screen readers can navigate by column, and adds aria-labels that expand numeric values into proper English ("bearing 320 degrees NW, 10,750 km", "5 minutes ago", etc.).

**Three sections in v1**, all matching the format in the issue body:

- **DX Spots** — frequency, callsign, bearing/distance from DE, age. Frequency gets a visually-hidden " megahertz" suffix (same pattern as #1001).
- **Satellites Overhead** — filtered to currently visible + elevation > 0, sorted by elevation descending. Columns: name, elevation, azimuth (with compass direction), mode, next pass minutes.
- **Activations** — combined POTA / SOTA / WWFF / WWBOTA in one section with a Program column. Same bearing/distance/age treatment.

[...]

PR #1069, opened 2026-06-06
feat(a11y): announce satellite rise/set events via aria-live (#997)

chrisdebian:
## What this does

Adds `useSatelliteAnnouncements` — a small hook that diffs the visible satellite set on each update cycle and announces transitions via two `aria-live` regions rendered in `App.jsx`.

- **Rise** (assertive): satellite newly above the user's configured minimum elevation → *"ISS now overhead"*
- **Set** (polite): satellite that was visible drops below minimum elevation → *"ISS passed below horizon"*
- Multiple satellites in one tick are joined naturally: *"ISS and AO-91 now overhead"*
- First render establishes a silent baseline (same pattern as `RigContext` rig-status announcements from #1000)
- Announcement text clears after 4 s so the same satellite can re-trigger on its next pass

## Why

Sub-issue of #997 (screen-reader accessibility umbrella). A blind operator using a screen reader can see the satellite list via the Map Data panel (#1002), but gets no prompt when a satellite becomes workable. This gives them a timely, spoken alert.

## Files changed

| File | Change |
|---|---|
| `src/hooks/app/useSatelliteAnnouncements.js` | New hook |
| `src/hooks/app/useSatelliteAnnouncements.test.jsx` | 8 tests — first render no-op, single rise, single set, multi-satellite phrasing, 4 s cleardown, stable-visible no-op, empty array guard |
| `src/App.jsx` | Import + call hook; two `visually-hidden` `aria-live` divs before `</main>` |

## Test plan

- [ ] Run `npm test -- --run src/hooks/app/useSatelliteAnnouncements.test.jsx` — 8 tests pass
- [ ] Enable a satellite layer (e.g. ISS) and wait for a pass — NVDA/VoiceOver should announce the rise and set
- [ ] Confirm no announcement fires on page load (baseline suppression)
- [ ] Confirm existing rig-status aria-live announcements are unaffected

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -54,6 +54,7 @@ import useLocalInstall from './hooks/app/useLocalInstall';
 import useVersionCheck from './hooks/app/useVersionCheck';
 import usePresence from './hooks/app/usePresence';
 import useAudioAlerts from './hooks/app/useAudioAlerts';
+import { useSatelliteAnnouncements } from './hooks/app/useSatelliteAnnouncements';
 import WhatsNew from './components/WhatsNew.jsx';
 import { initCtyLookup } from './utils/ctyLookup.js';
 import { getAllLayers } from './plugins/layerRegistry.js';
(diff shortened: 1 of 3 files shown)


accius (approved):
Clean implementation. The diff-based visible-set tracking with null-baseline-on-first-render mirrors the RigContext pattern exactly, and the 8 tests cover all the cases I'd want including the multi-satellite name concatenation. Nice touch using `createRoot` + `act` directly instead of pulling in `@testing-library/react`.

Heads up: this PR and #1070 both add a new aria-live div at the same spot in `App.jsx` (right after `WhatsNew`). Whichever lands first wins; the other will need a 30-second rebase on that hunk. No code conflict beyond that.

73 de K0CJH

PR #1071, opened 2026-06-06
fix(panel): fix layout and style issues

lbatalha:
## What does this PR do?

This PR fixes some inconsistent layout and styles in the PSKReporter panel:

- PSKReporter vs WSJT-X tab - PSK is missing the color due to variable not existing. We should also migrate all these hardcoded colors to variables but that will require using something like `color-mix` to add in the opacity to the 3 channel color strings
- PSK vs WSJT-X second row buttons had inconsistent styling - fixed incorrect div nesting
- PSKReporter setting buttons overflowed at small screen sizes - Moved all controls to a dedicated row for both PSK and WSJT
- Excessive padding in main body causes unnecessary space usage and is inconsistent with every other panel - remove all extra padding, the parent padding settings are already enough
- WSJT-X body pre-connection message had broken variable string replacement
- Tweaked default right panel flex spacings in modern mode to improve usability of PSK panel at minimum desktop mode screen sizes (could trigger even on 22-24" 1920x1080 panels at common scaling ratios)
- WSJT-X tab text wraps on the hyphen - made it so that cant happen, though it never can now with these changes.
- Moved PSKReporter status light to the tab - thoughts?

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Check that PSKReporter panel operates as before but with improved usability and consistency
2. ???
3. Profit

## Checklist

Screenshots taken at minimum vertical size allowed by modern layout before horizontal shrinking is triggered (vertical black bars), ie minimum panel heights allowed by styles:

### Before:

image

Notice that there are 3 missing control buttons due to lack of horizontal space

image

Notice how you cannot scroll further up in the WSJT-X body, and `pskReporterPanel.wsjtx.udpPath` appears instead of the correct message. Notice the text wrapping in the WSJT-X tab

Overall default bar look:

image

### After:

image

image

Notice correct text

Overall default bar look with WSJT-X in WSPR mode for completeness:

image


lbatalha:
TODO: fix DX Cluster panel control button discrepancies to regain even more vertical space in modern mode

lbatalha:
I have modified the DX Cluster panel so that it is now consistent with other panels:

Before:

image

Besides the multi-line buttons, notice we are overflowing far from being near tablet widths


After:

image

I made the viewport even smaller horizontally for this screenshot, a few pixels away from swapping to tablet mode, and it all fits!

The filter button shows number of active filters just like other panels (do we even need this, isnt just making the button amber enough?
image

One thing I also fixed that was pissing me off is in the filter popup - Clear Spots is in amber (which is used for filters) and clicking it removes all the spots from your list! Thats very destructive and should be red! The red Clear All actually only clears the filters like other panels. The colors are flipped imo. Here is a before and after:

image
image

That being said, all the other panel's filter popups have a red `Clear All Filters` button, so this makes it inconsistent. They dont have the Clear Spots button so its a bit of a conundrum. What should we do about this?

accius (changes requested):
Real improvements packed in here. The visual polish at small panel widths, the row restructure, the Clear Filters / Clear Spots color swap to match destructiveness semantics, the `var(--accent-blue)` / `var(--accent-purple)` migration, and the translation key whitespace bug fix all read well. A few things I want to land before merge.

### Blocking: ARIA tablist regression for WSJT-X

The new sub-tabs section hoists `role="tablist"` to always wrap the buttons:

```jsx
<div role="tablist" aria-label="..." onKeyDown={(e) => ariaTabKeyDown(e, PSK_TABS, ...)}>
  {panelMode === 'psk' ? (
    <>
      <button role="tab" ... />
      <button role="tab" ... />
    </>
  ) : (
    <>
      <button onClick={() => setWsjtxTab('decodes')}>...</button>
      <button onClick={() => setWsjtxTab('wspr')}>...</button>
      <button onClick={() => setWsjtxTab('qsos')}>...</button>
    </>
  )}
</div>
```

When WSJT-X mode is active you've got a `role="tablist"` containing buttons that lack `role="tab"`/`aria-selected`/`aria-controls`. ARIA requires a tablist to contain tabs. Screen readers will announce "tablist with 0 tabs" or skip it entirely. Also the `onKeyDown={ariaTabKeyDown(e, PSK_TABS, ...)}` runs in both modes but only knows about the PSK tabs, so arrow-key nav over WSJT-X tabs is broken.

Two ways to fix:

- **Simpler:** revert the tablist hoist. Keep it conditional, only PSK tabs get the tablist treatment, WSJT-X stays as plain buttons in a flex row. Matches the pre-PR shape with your visual restructure on top.
- **Better long-term:** give the WSJT-X buttons `role="tab"`, an `aria-selected` / `tabIndex` shape that mirrors the PSK ones, and a `WSJTX_TABS` constant + ref map so the onKeyDown picks the right set based on `panelMode`.

I'd land the simpler revert here and file the WSJT-X tablist proper as a follow-up since the #1045 push already covered that pattern for the panels you've touched.

### Nit: commented-out style

```js
//height: '100%',
```

In the WSJT-X waiting message styles. Your checklist (rightly) excludes `.bak` / `console.log` / debug lines and commented-out style hits the same spirit. Just delete it.

### Nit: DXClusterPanel filter button loses discoverability

```jsx
- {t('dxClusterPanel.filtersButton')}
+ {filterCount > 0 ? filterCount : ''}
```

When no filters are active, the button is just `[IconSearch]` with no visible text. It has `aria-label` but no `title=`, so sighted mouse users get nothing on hover either. Two paths: add a `title=` matching the existing aria-label (mirrors the #1064 pattern), or keep the "Filters" text and just append the count when > 0. Either works.

### Approving once those land

The bigger structural reorganization in PSKReporterPanel and the color semantics fix in DXFilterManager are genuinely better than what was there. Status dot in the tab label is a nice touch. Once the ARIA thing is addressed I'll approve.

73 de K0CJH

lbatalha:
Will fix. Thoughts on the filter menu button color issue I mentioned? Should I also change the `Clear All Filters` button in the filter menu for the other panels to be amber, or should I revert the color change (but change the clear filter button text to be consistent)?

lbatalha:
Implemented the tab role et al for the WSJT-X tabs, might as well get it done now. The DXCluster search button issue does not exist, it did have a Title but no aria label, so I added that for consistency.

accius (approved):
Re-reviewed against the 06-07 push. All three items from the last pass are addressed:

- **ARIA tablist regression (was blocking):** resolved, and you took the better path rather than the quick revert. The WSJT-X buttons now carry `role="tab"`, `aria-selected`, `aria-controls`, `tabIndex`, and ref wiring into `wsjtxTabRefs`, the tablist `aria-label` is mode-aware, and `onKeyDown` now drives off `ALL_TABS` / `activeTabKey` / `setActiveTabFn` / `activeTabRefs` so arrow-key nav follows whichever mode is active. That is the correct shape.
- **Commented-out `//height: '100%'`:** gone.
- **DXClusterPanel filter button discoverability:** `title=` plus the added `aria-label` covers both hover and screen-reader paths now.

Two minor things, neither blocking, fine as follow-ups:

1. `aria-controls="panel-psk-content"` / `panel-wsjtx-content` still point at ids that do not exist on the content container (grep finds no matching `id=`). This predates your PR, the original PSK tabs had the same dangling reference, but since you are in here giving WSJT-X the full treatment it would be a good time to add the matching `id` to the content `<div>` so the relationship actually resolves.
2. `WSJT_X_TABS` always includes `'wspr'`, but the WSPR tab button only renders when `isWSPRMode || wsjtxWspr.length > 0`. Arrow-keying can therefore land `wsjtxTab` on `'wspr'` when no WSPR tab is visible. `ariaTabKeyDown`'s `?.focus()` makes the focus call a safe no-op, so worst case is a hidden/empty content state, but you may want to compute the tab list from what is actually rendered.

The PSKReporterPanel row restructure and the Clear Filters / Clear Spots color-semantics swap in DXFilterManager are real improvements. Approving.

K0CJH

lbatalha:
Regarding 1, I have `id={panelMode === 'psk' ? 'panel-psk-content' : 'panel-wsjtx-content'}` [here](https://github.com/accius/openhamclock/pull/1071/changes#diff-c07d9d200ed9371671e89c6a7144cee4c3a9f29d52ab59ec05e624267a40b503R630). It sets the `id` tag for the content div. Is that what you're referring to?

Fixed 2

accius (approved):
Looked over the new commits since my approval. The WSPR tab filtering plus the effect fallback to decodes covers both the arrow-key cycling and the stale-selection case, and the lang file reorder is just alphabetization. Still good to go.

K0CJH

PR #1012, opened 2026-05-24
Update local run documentation for wasm and fix script permissions

alanhargreaves:
## What does this PR do?

Updates the Installation section of the top-level `README.md` with a note to run `scripts/fetch-wasm.sh` if you want clients on your local server to use the client-side propagation calculation module.

Also adds execute permission to `scripts/fetch-wasm.sh`.

I can't be the only one that needed to spend way more time than necessary to figure this out.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [X] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Go to a local client and verify that the VOACAP panel says that it is using "WASM" rather than "EST"

## Checklist

Documentation and permissions change only.

## Screenshots (if visual change)

image

PR #1072, opened 2026-06-06
[fix] convert bash fetch-wash.sh -> node fetch-wasm.js

MichaelWheeley:
## What does this PR do?

- problem identified during a recent push to staging ref https://github.com/accius/openhamclock/commit/2f0c3c7dc785e5abb639508c978cd931d2f387cf where bash command had prevented `npm start` on Powershell
- rewrite `fetch-wash.sh` as `fetch-wasm.js`
- prebuild: `bash fetch-wash.sh` -> `node fetch-wasm.js`
- convert all existing uses of bash script -> `node fetch-wasm.js`
- also fixes problem identified in https://github.com/accius/openhamclock/pull/1012#issuecomment-4529374381

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## Checklist


accius (changes requested):
Good catch on the bash assumption. Git for Windows ships bash.exe but only into the Git Bash environment, not into PowerShell's PATH. My mistake. Node-based is the right call.

The script works end-to-end. I just ran it against an empty `public/wasm/` and both `p533.mjs` and `p533.wasm` landed on disk. But there's one bug worth fixing before merge: the sha256 check is broken in a way that silently false-alarms on every run.

### The bug

`public/wasm/p533.sha256` has two lines:

```
389344e88cf...  p533.mjs
05cf70f9ac7...  p533.wasm
```

The script does:

```js
const expected = raw.split(/\s+/)[0];  // 389344... (which is p533.mjs's hash)
const actual = await sha256File(wasmFile);  // hash of p533.wasm
```

So it compares p533.mjs's expected hash to p533.wasm's actual bytes and always sees a mismatch. Output is:

```
⚠  fetch-wasm: sha256 mismatch on downloaded WASM — skipping (runtime will use REST fallback)
```

The downloaded files are still on disk (download finished before verifyChecksum runs), so Vite bundles them and runtime works. But the integrity check that's supposed to catch tampered or corrupted assets never actually runs, and it's a noisy false alarm on every build.

### Suggested fix

Parse the sha256 file line-by-line and verify each entry, the same way `sha256sum -c p533.sha256` did in the bash original:

```js
async function verifyChecksum(destDir) {
  const checksumFile = path.join(destDir, 'p533.sha256');
  if (!fs.existsSync(checksumFile)) return;

  const lines = fs.readFileSync(checksumFile, 'utf8')
    .split('\n')
    .map((line) => line.trim())
    .filter(Boolean);

  for (const line of lines) {
    const [expected, filename] = line.split(/\s+/);
    if (!expected || !filename) continue;
    const filePath = path.join(destDir, filename);
    if (!fs.existsSync(filePath)) continue;
    const actual = await sha256File(filePath);
    if (expected !== actual) {
      throw new Error(`sha256 mismatch on ${filename}`);
    }
  }
}
```

### Smaller nit

The `download` response stream doesn't have an error handler. If the network drops mid-download the file write might hang instead of rejecting. Adding `response.on('error', reject);` next to the existing `fileStream.on('error', reject);` would tighten it up. Not critical for a release-asset download but worth the two lines.

Once the sha256 thing is sorted I'll approve. Apologies for the Windows-breaking original, that one's on me.

73 de K0CJH

MichaelWheeley:
especially please check Docker (change made to `./Dockerfile`) but not tested
### summary,
- remove existing bash script
- existing executions of bash script -> node.js code
- verifyChecksum rework
- download filestream response protect

previously relied on bash script and displayed only EST, now
image

accius (approved):
Both points fixed cleanly.

`verifyChecksum` now walks each line of `p533.sha256` and pairs the hash with the named file, which is what `sha256sum -c` was doing in the bash original. I ran it locally against a fresh `public/wasm/` and the false-alarm warning is gone — checksum verifies for both `p533.mjs` and `p533.wasm`. If either ever drifted, the throw would surface it instead of getting swallowed.

`response.on('error', reject)` next to the existing `fileStream.on('error', reject)` is exactly right. Mid-download network drops will now reject the promise instead of hanging.

CI is green and the Docker layer still installs bash for `vendor-download.sh`, which is fine. Approving.

73 de K0CJH

PR #1075, opened 2026-06-10
Observability wiring

lbatalha:
## What does this PR do?

This PR adds an initial base implementation of observability features.
It consists of adding a conventional prometheus-compatible metrics endpoint at `/metrics`

Current metrics implemented include:
- internal node.js metrics
  - eventloop, gc, etc..
- dynamic scraping of existing server routes - we dont need to manually configure scraping for new routes in the future
  - API requests total
  - API response duration
  - API request/response body size
- existing stats are also tracked - we should remove the built in stats system later when we have a full monitoring stack implemented as its redundant
- Subsystem health status - uses the same numbering but we should not have OK/Unknown be the same number (and OK should not be zero, thats uncommon) - to change later if we remove the builtin subsystem status

Metrics are lazy-loaded on scrape requests as is standard to optimize resource usage.

Care was taken (and must continue to be taken) to maintain minimal cardinality. URL path parameters are normalized the opentelemetry's standard {param_name} format in the `route` label so that we dont increase cardinality unbounded. Any future metrics we add MUST ensure we dont include high cardinality labels.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Visit `/metrics`
2. Check that expected metrics appear
3. Check that making requests to an API endpoint updates related metrics

## Checklist

- [x] If touching `server.js`: caches have TTLs and size caps (we serve 2,000+ concurrent users)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "mqtt": "^5.15.0",
         "node-fetch": "^2.7.0",
         "papaparse": "^5.5.3",
+        "prom-client": "^15.1.0",
         "react-colorful": "^5.6.1",
         "react-i18next": "^16.5.4",
         "rss-parser": "^3.13.0",
@@ -1587,6 +1588,15 @@
         "node": ">=10"
       }
     },
+    "node_modules/@opentelemetry/api": {
+      "version": "1.9.1",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/api/-/api-1.9.1.tgz",
+      "integrity": "sha512-gLyJlPHPZYdAk1JENA9LeHejZe1Ti77/pTeFm/nMXmQH/HFZlcS/O2XJB+L8fkbrNSqhdtlvjBVjxwUYanNH5Q==",
+      "license": "Apache-2.0",
+      "engines": {
+        "node": ">=8.0.0"
+      }
+    },
     "node_modules/@pkgjs/parseargs": {
       "version": "0.11.0",
       "resolved": "https://registry.npmjs.org/@pkgjs/parseargs/-/parseargs-0.11.0.tgz",
@@ -3054,6 +3064,12 @@
       "bin": {
         "baseline-browser-mapping": "dist/cli.js"
       }
+    },
+    "node_modules/bintrees": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/bintrees/-/bintrees-1.0.2.tgz",
+      "integrity": "sha512-VOMgTMwjAaUG580SXn3LacVgjurrbMme7ZZNYGSSV7mmtY6QQRh0Eg3pwIcntQ77DErK1L0NxkbetjcoXzVwKw==",
+      "license": "MIT"
     },
     "node_modules/bl": {
       "version": "6.1.6",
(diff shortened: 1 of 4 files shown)


accius (commented):
Solid first cut at observability and the lazy default-metrics collection plus the dynamic route-pattern extraction is a nice touch, it means new routes get scraped without anyone remembering to wire them. CI is green and the prom-client dependency tree is clean. A couple of things I would want sorted before this endpoint sees real production traffic, then I am happy.

### Should address before heavy traffic

1. **Unbounded label cardinality on unmatched routes.** `normalizeRoute()` falls back to returning the raw `reqPath` when nothing matches, and that value becomes the `route` label on `ohc_api_requests_total`, `ohc_api_duration_seconds`, and both size histograms. On a public-facing Railway host you get scanned constantly (`/wp-login.php`, `/.env`, random paths), and every distinct path spins up a new metric series that prom-client keeps in memory for the life of the process. That is a slow memory leak plus scrape bloat. Bucket anything that does not match a known pattern into a single `route="unknown"` label (or only record matched routes).

2. **`/metrics` is unauthenticated and public.** It exposes process internals, the full route list, visitor totals, and active session counts to anyone. That may be an acceptable trade for this app, but it should be a conscious decision rather than a default. Options: bind it to the Railway private network only, gate it behind the existing `requireWriteAuth` / a bearer token, or an allowlist. Let me know which way you want to go.

### Nits

3. `apiMetricsMiddleware` uses `console.log('[Prometheus] ...')` for the one-time pattern-extraction line. The rest of the server goes through `logInfo` / `logDebug` from `server/utils/logging` (you already use `logErrorOnce` for the endpoint handler). Route it through the same logger so log levels and rate limiting apply.

4. `extractRoutePatterns` reads `app._router.stack`, which is a private Express internal that moves to `app.router` in Express 5. Fine on the current 4.x pin, just leaving a marker for whenever the framework bump happens.

5. The `setInterval` in `setSubsystemsHealth` is never cleared and not `unref()`-ed. Harmless for a singleton process, but an `.unref()` keeps it from holding the event loop open on shutdown.

None of this is a hard blocker, but I would like 1 and 2 resolved before we point Prometheus at production. Happy to re-review quickly once they are in.

K0CJH

accius:
You're right that open /metrics is the norm when there's an ingress doing the access control. The snag is Railway specifically doesn't give us that layer. There's no inbound IP allowlist or request-rule config on Railway's public networking the way nginx or an ALB has. Their own guidance is "if you need to restrict inbound, put Cloudflare in front and filter there." So "the proxy handles it" only holds where a proxy is actually configured to.

And right now it isn't. I just hit `https://stagingbranch26.openhamclock.com/metrics` and it returns 200 with the full `ohc_*` dump, served straight through Cloudflare with no rule in front of it. So as it stands the endpoint is wide open on a host that's public by default and serving visitor and session counts plus the whole route map in the payload. Prod isn't affected yet since this only landed on Staging.

Two ways I'd be happy with:

1. Railway private networking. Don't publish /metrics on the public domain at all. Keep it internal and scrape it from a Prometheus service running inside the same Railway project over `*.railway.internal`. That's the [pattern Railway themselves recommend](https://docs.railway.com/reference/private-networking) and it removes the public entry point entirely. Cleanest if the scraper lives in Railway.
2. Cloudflare rule. Prod sits behind Cloudflare, so a firewall rule that denies /metrics except from the scraper (by IP, or a shared header/token) gives you exactly the ingress-level control you're describing, no app code. Downside is it only covers prod, not Staging and not self-hosters, who run this with nothing in front of it at all.

If the scraper is going to live outside Railway (Grafana Cloud etc.), I'd lean option 2 or just a bearer token on the route. If it's in-Railway, option 1 is strictly better. No strong preference, I just don't want it sitting wide open by default.

Separate thing, unrelated to the auth question: the unmatched-route label still wants bucketing to `unknown` regardless of how we gate access, since a scanner walking random paths grows the series set unbounded. Happy to pair on either of these.

K0CJH

PR #1077, opened 2026-06-10
fix(prometheus): restrict /metrics access & limit unknown route cardinality

lbatalha:
## What does this PR do?
- Guard /metrics behind optional METRICS_AUTH_KEY env var — public by default, useful for Railway private-network and other deployments
- Collapse all unmatched API paths into {unknown} label so random requests no longer create unbounded time series

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. Verify that requests that are not part of our routes show up as `route={unknown}` in metrics
2. Verify that no histogram metrics are created for this type of route
3. Verify that setting the env var blocks access unless correct header is set

## Checklist

- [x] If touching `server.js`: caches have TTLs and size caps (we serve 2,000+ concurrent users)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,6 +52,13 @@ HOST=localhost
 # REQUIRED for cloud/public deployments to prevent unauthorized access.
 # API_WRITE_KEY=your-secret-key-here
 
+# API key for the Prometheus /metrics endpoint.
+# When set, requires a matching ?key= query param or Authorization: Bearer <key> header.
+# Leave unset for public metrics (default, fine for local/private installs).
+# Useful for Railway deployments behind private networking — set this key so only
+# internal scrapers (e.g. Prometheus in the same Railway project) can access metrics.
+# METRICS_AUTH_KEY=your-secret-key-here
+
 # CORS allowed origins (comma-separated)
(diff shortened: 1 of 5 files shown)


lbatalha:
just noticed that in a normal deployment the frontend vite server proxies API endpoints to the backend, I need to fix that for the /metrics endpoint as well.

accius (approved):
This closes out both items I flagged on #1075. The METRICS_AUTH_KEY gate matches the API_WRITE_KEY pattern, supporting both the query param and Bearer header is the right call for Prometheus scrape configs, and collapsing unmatched paths into {unknown} kills the cardinality risk while keeping the counter so scanners still show up in rate(). Skipping the histograms for unknown routes is the right instinct too, that is where the label explosion actually hurts.

One non-blocking thought: the ?key= variant will end up in access logs wherever this runs behind a proxy that logs query strings. The Bearer header avoids that, might be worth a one-liner in .env.example nudging people toward it.

The remaining nits from the #1075 review (project logger, app._router on Express 5, unref on the interval) are still open but none of them block this.

K0CJH

PR #1079, opened 2026-06-11
azimuth fix

AntDiClementi:
AzimuthalMap.jsx:

pass browser canvas pixel scale into the tile reprojector
draw the reprojected image back into CSS-sized canvas space
tileReproject.js:

choose higher source tile zoom based on Azimuthal zoom level
clamp Web Mercator latitude sampling
add bilinear sampling
render reprojected ImageData at scaled canvas resolution

### DIFF
--- a/src/components/AzimuthalMap.jsx
+++ b/src/components/AzimuthalMap.jsx
@@ -317,6 +317,7 @@ export default function AzimuthalMap({
       panX: vs.cx - size.w / 2,
       panY: vs.cy - size.h / 2,
       lowMemory: lowMemoryMode,
+      pixelRatio: Math.max(1, window.devicePixelRatio || 1),
     })
       .then(() => setTilesReady(true))
       .catch(() => {});
@@ -377,7 +378,7 @@ export default function AzimuthalMap({
     const canvas = canvasRef.current;
     if (!canvas) return;
     const ctx = canvas.getContext('2d');
-    const dpr = window.devicePixelRatio || 1;
+    const dpr = Math.max(1, window.devicePixelRatio || 1);
     canvas.width = size.w * dpr;
     canvas.height = size.h * dpr;
     ctx.setTransform(dpr, 0, 0, dpr, 0, 0);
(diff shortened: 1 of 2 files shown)


accius (changes requested):
This is the real deal this time, and the rendering work itself is right: the bilinear sampler is correctly written (X wrap, Y clamp, per-channel lerp), the Mercator latitude clamp kills the pole artifacts, and the DPR-scaled ImageData drawn back into CSS space through the transform is the correct way to get crisp output on retina. Labels and borders will look dramatically better.

What I cannot take as-is is the tile zoom table, because the memory and fetch math got steep:

- Normal mode base tier went 3 to 4: the equirectangular canvas goes 2048px to 4096px, which is 67 MB of RGBA plus the canvas itself, built from 256 tile fetches.
- The new zoom 5 tier (azZoom > 1.75) is an 8192px canvas, roughly 268 MB of pixel data and 1024 tile fetches per rebuild. That is a lot to ask of the tile provider and of any machine that is not a gaming desktop.
- Low-memory mode went 2 to 3/4. That mode exists specifically for the Pi kiosk fleet, and a 4096px buffer at azZoom > 1.75 on a Pi 3 with 1 GB of RAM shared with Chromium is asking for an OOM kill.

Also a small one: chooseTileZoom is now called with pixelRatio as a third argument but the function ignores it.

Concrete asks:
1. Cap normal mode at 4 and gate it on pixelRatio >= 2 (you are already passing it in, use it): dpr 1 displays cannot show the extra detail anyway, so they keep today's 3/4 cost.
2. Leave lowMemory at 2, or at most 2/3 with the 3 gated on zoom. The bilinear sampling alone will already improve Pi output at the existing tile zoom.
3. Either use or drop the dead pixelRatio param in chooseTileZoom.

With those, this merges fast. The sampling and DPR plumbing do not need to change at all.

K0CJH

accius (approved):
Re-reviewed after the updates and all three asks are addressed: lowMemory stays at tile zoom 2 so the Pi fleet keeps its current memory ceiling, normal mode is capped at 4 with the 4 gated on pixelRatio >= 2 so dpr-1 displays keep today's 3/4 cost, and chooseTileZoom now actually consumes the pixelRatio it is handed. The zoom 5 tier and its 268 MB buffer are gone.

The rest I already signed off on last pass and it has not changed: bilinear sampler, Mercator latitude clamp, and the dpr-scaled ImageData drawn back through the transform. The cache key picking up outputScale is the right touch so a dpr change invalidates correctly.

Approved. Note the docker build-and-push job was still running when I posted this, so let that finish before merging.

K0CJH

PR #1081, opened 2026-06-11
fix(makeDraggable): floating windows drift on click in Firefox

### LINKED ISSUE #1017: [BUG] Floating windows move after being dropped
## Describe the Bug

In Modern Display mode, when the user grabs and lets go of the map floating windows, they do not remain in the location they are dropped in. They slide to the right every time they are dropped, even if they are not moved.

## To Reproduce

1. Enable Modern Display mode
2. Enable at least one map layer that has a floating info window (eg. gray line)
3. Use the mouse left-click hold to grab the window and release
4. See that the window moves to the right instead of staying where it is.

## Expected Behavior

It is expected that the window stays where it is dropped - this is the case in the Dockable display mode for example.

## Screenshots

N/A

## Environment

- **OS**: MacOS 26.5
- **Browser**: Firefox 151.0.2
- **OpenHamClock Version**: Latest docker image (26.3.3?)
- **Running As**: Docker

## Console Errors

N/A

## Additional Context

N/A

---

73!

ceotjoe:
## Summary

- `offsetLeft`/`offsetTop` returns `0` for `position: fixed` elements in Firefox (spec-compliant behaviour when `offsetParent` is `null`), causing every click-release to snap panels to the viewport origin and save position `0,0`
- Replaced all three usages with `parseFloat(el.style.left/top)`, which is reliable because `makeDraggable` always sets those values explicitly before any drag interaction
- Chrome was unaffected because it returns the actual visual position for fixed elements (non-spec behaviour)

## Test plan

- [ ] Enable Modern Display mode
- [ ] Enable a map layer with a floating panel (e.g. Gray Line)
- [ ] Click and release the panel title without moving — panel should stay in place (was drifting to left edge in Firefox)
- [ ] Drag the panel to a new position and release — panel should stay where dropped
- [ ] Reload the page — panel should restore to the saved position
- [ ] Double-click the panel title — panel should reset to its default position
- [ ] Verify behaviour is unchanged in Chrome/Chromium

Fixes #1017

### DIFF
--- a/src/plugins/layers/makeDraggable.js
+++ b/src/plugins/layers/makeDraggable.js
@@ -161,8 +161,8 @@ export function makeDraggable(
       didDrag = false;
       startX = e.clientX;
       startY = e.clientY;
-      startLeft = el.offsetLeft;
-      startTop = el.offsetTop;
+      startLeft = parseFloat(el.style.left) || 0;
+      startTop = parseFloat(el.style.top) || 0;
       previousTransition = el.style.transition;
       el.style.transition = 'none';
       dragHandle.style.cursor = 'grabbing';
@@ -208,14 +208,14 @@ export function makeDraggable(
       suppressClick = didDrag;
 
       if (snap) {
-        el.style.left = snapToGrid(el.offsetLeft, snap) + 'px';
-        el.style.top = snapToGrid(el.offsetTop, snap) + 'px';
+        el.style.left = snapToGrid(parseFloat(el.style.left), snap) + 'px';
+        el.style.top = snapToGrid(parseFloat(el.style.top), snap) + 'px';
       }
 
       clampToViewport(el);
 
-      const topPercent = (el.offsetTop / window.innerHeight) * 100;
-      const leftPercent = (el.offsetLeft / window.innerWidth) * 100;
+      const topPercent = (parseFloat(el.style.top) / window.innerHeight) * 100;
+      const leftPercent = (parseFloat(el.style.left) / window.innerWidth) * 100;
       localStorage.setItem(
         storageKey,
         JSON.stringify({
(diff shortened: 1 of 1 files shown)


accius (commented):
Diagnosis is right (Firefox returns 0 for offsetLeft/offsetTop when offsetParent is null on fixed elements, and that is spec-conformant) and this should finally close #1017. But parseFloat(el.style.left) has a unit trap: the restore path in makeDraggable sets saved positions as percentages (el.style.left = data.leftPercent + '%'), so after a reload parseFloat reads '37.2%' as 37.2 and the first click-release snaps the panel toward the viewport origin in every browser, not just Firefox. Same bug class you are fixing, different trigger.

Suggest reading position from getBoundingClientRect() instead of style at all three sites:

- mousedown: const rect = el.getBoundingClientRect(); startLeft = rect.left; startTop = rect.top;
- mouseup snap + percent save: same thing, fresh rect

For a fixed-position element, rect.left/top is viewport-true regardless of whether style.left is px, %, or unset, and it behaves identically in Firefox and Chrome. As a bonus it also handles the skipPositionLoad callers that never set style.left themselves.

K0CJH

ceotjoe:
Good catch — updated in 26ebe71. All three sites (mousedown start capture, mouseup snap, mouseup save) now use `getBoundingClientRect()` instead. That's viewport-true regardless of whether `style.left` is `px`, `%`, or unset, so the percentage unit trap is gone and the `skipPositionLoad` callers are covered too. 73!

accius (approved):
This is exactly the fix. getBoundingClientRect at all three sites (drag start, snap, percent save) is viewport-true for fixed elements no matter whether style.left holds px, %, or nothing at all, so this closes the Firefox offsetParent quirk and the percent-restore snap trap in one move. I checked the callers for the one thing rect-based reads are sensitive to, CSS transforms, and we are clean: the restore path sets transform to none, the lightning proximity panel does the same in every branch and deliberately avoids translate centering, and no other draggable panel uses a transform.

CI is green. Approving, and this should finally put #1017 to bed.

K0CJH

PR #1082, opened 2026-06-11
fix(prometheus): small nit tweaks and add README documentation

lbatalha:
## What does this PR do?

This fixes a few nits raised in #1075 as well as adds some documentation to the README

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [x] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,8 @@ HOST=localhost
 # Leave unset for public metrics (default, fine for local/private installs).
 # Useful for Railway deployments behind private networking — set this key so only
 # internal scrapers (e.g. Prometheus in the same Railway project) can access metrics.
+# NOTE: Prefer Authorization: Bearer over ?key= — the latter leaks into access logs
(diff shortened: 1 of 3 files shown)


accius (approved):
All four changes are right: the .unref() on the subsystems interval means the process can exit cleanly instead of being held open by a timer nobody cancels, logInfo keeps the route-pattern line subject to LOG_LEVEL like the rest of the server, and the Bearer-over-query-string note in .env.example is good advice since ?key= lands in proxy access logs verbatim. Verified the docs match the implementation: server.js accepts both Authorization: Bearer and ?key=, so describing Bearer as the preferred path is accurate.

One take-it-or-leave-it nit: "prometheus compatible" in the README section should be "Prometheus-compatible". Not holding the merge on capitalization.

K0CJH

PR #1084, opened 2026-06-11
[refactor] fetch-wasm.js checksum

MichaelWheeley:
## What does this PR do?

- refactor `scripts\fetch-wasm.js`
- fetch WASM checksum, if ok skip full download
- improved checksum verification, checks both checksum file list against expected file list, and expected files against checksum list

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test
- `npm run prebuild`

## Checklist

### DIFF
--- a/scripts/fetch-wasm.js
+++ b/scripts/fetch-wasm.js
@@ -51,7 +51,7 @@ const sha256File = async (filePath) => {
 };
 
 /**
- * Verifiy checksum of files in destination folder. File p533.sha256 should exist and contain checksum information.
+ * Verify checksum of files in destination folder. File p533.sha256 should exist and contain checksum information.
  * @param {*} destDir Destination directory where files are downloaded.
  * @param {*} filesExpected List of expected files to verify, the first file in the list should be the checksum file (e.g. p533.sha256) and the rest are files to verify (e.g. p533.mjs, p533.wasm).
  * @returns returns true if checksum file exists and matches, false if checksum file is missing or invalid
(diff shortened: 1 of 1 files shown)


accius (approved):
The checksum-first skip is a good optimization: builds with a current bundle now download a few hundred bytes instead of the full wasm every time, and the two-way verification (expected files present in the checksum list AND every listed file hashing clean) closes the gap where a truncated sha256 file would pass silently. The response.resume() on redirects is a nice catch too, that was leaking sockets on every GitHub release redirect.

Fail-open semantics are preserved (any throw lands in warn() which exits 0 so builds never break on network trouble), which was the property I cared most about from the original review. Approving with two non-blocking nits if you feel like a follow-up:

- The mismatch error message interpolates the outer filename const, which is always 'p533.sha256', so a hash failure on p533.wasm reports the wrong file. Capturing which file failed inside verifyChecksum would make that diagnostic honest.
- fs.rmSync deletes the existing checksum file before the new download is known to succeed. Offline builds still ship the existing wasm (good), but they lose the checksum file, so every later build does a full re-download until the network is back. Downloading to a temp name and renaming on success would keep the fast path intact.

K0CJH

MichaelWheeley:
follow up changes made in https://github.com/accius/openhamclock/pull/1089

PR #1085, opened 2026-06-11
feat: add DX callsign lookup popup, rearrange DX target controls, simplify DX time

lbatalha:
## What does this PR do?

- Add info button (i) in ModernLayout DX target panel to open the callsign lookup popup via useCallsignPopup, with proper title, aria-label, and aria-disabled attributes
- Move DXCC toggle, favorites (☆), and lock (🔒/🔓) buttons into the header row of the DX target panel (Modern layout), with matching 2px 6px padding
- Replace DXLocalTime toggle with a simple static local-only time display — remove UTC/Local toggle state, localStorage persistence, and UTC formatting;
- Remove dead i18n keys (showLocal, showUtc, etc.), add new keys (callsignLookupTitle, callsignLookupAriaLabel, callsignLookupEmptyTitle, solarFallbackTitle)
- Add aria-label to DXCC toggle button
- Fix overflowing Grid/Callsign line

## Type of change

- [x] Bug fix
- [x] New feature

## How to test

1. Check that all existing buttons work
2. Check that info button shows the callsign lookup popup as per rest of panels

## Checklist

Minimum desktop width looks like:

image

Normal size looks like:

image


accius (approved):
Went through all three pieces. The lookup button wiring is right in both layouts (useCallsignPopup from #1068, anchor ref for positioning, disabled state when no callsign with proper title/aria variants), the header-row consolidation of DXCC/favorites/lock declutters the value row nicely, and letting the grid/callsign inputs flex instead of hard ch widths fixes the cramped feel at narrow panel sizes. Lang files are complete across all 16 locales and check-json-keys agrees.

One thing I want to call out explicitly so it's a conscious decision and not a drive-by: dropping the UTC/local toggle from DXLocalTime removes a piece of the #1038 feature that shipped in 26.4.0, along with its localStorage preference. I think the simplification is defensible since UTC is permanently visible in the header clock and the DX-local reading is the part with actual information content, and the solar-fallback ⚠ marker survives. But if anyone was using the toggle the old key just orphans silently. Fine with it going out this way; if we get complaints we know what to restore.

Tiny nit, take it or leave it: with disabled set on the lookup button, aria-disabled is redundant, and a disabled button drops out of the tab order entirely. Keyboard users with no DX callsign set won't discover the control exists. aria-disabled alone with a no-op click would keep it focusable. Not worth blocking a UI PR over.

K0CJH

PR #1089, opened 2026-06-12
[fix] WASM checksum PR follow-up

MichaelWheeley:
## What does this PR do?
- follow-up to https://github.com/accius/openhamclock/pull/1084
- use temporary file for download, only replace existing checksum file if download successful
- correct mismatch warning message

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [X] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

### DIFF
--- a/scripts/fetch-wasm.js
+++ b/scripts/fetch-wasm.js
@@ -126,7 +126,7 @@ const verifyChecksum = async (destDir, filesExpected) => {
       }
 
       if ((await verifyChecksum(DEST_DIR, filesExpected)) === false)
-        throw new Error(`sha256 mismatch, verifyChecksum failed after download`);
+        throw new Error(`sha256 mismatch, verifyChecksum failed after WASM package download`);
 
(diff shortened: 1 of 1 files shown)


accius (approved):
Both follow-up items from the #1084 review are handled cleanly. The temp-file-then-rename keeps the existing checksum file intact when a download fails, so an offline build still verifies against what it already has on disk instead of wiping the checksum and forcing a full re-download on every subsequent build. The error cleanup on the temp file is the right touch.

The mismatch message no longer interpolates the outer filename const, so it stops naming the wrong file. It is generic now rather than naming which file actually failed, which is a fair trade for not being misleading. If you ever want the per-file detail back, returning the failing filename out of verifyChecksum would get you there, but that is firmly optional.

Fail-open semantics still intact (everything lands in warn() which exits 0). CI green.

K0CJH

PR #1090, opened 2026-06-12
a11y: aria-live announcements for BandConditionsPanel (#997)

chrisdebian:
## Summary

Closes part of #997.

The `BandConditionsPanel` had no ARIA attributes, making it silent to screen readers even though band conditions update hourly.

**Changes in `BandConditionsPanel.jsx`:**

- `aria-live="polite"` + `aria-label` on the conditions grid container — screen readers announce condition changes when the hourly data refreshes
- `aria-label="{band}: {condition}"` on each band cell (e.g. `"40m: Good"`) so the full meaning is conveyed; visual child divs marked `aria-hidden="true"` to prevent double-reading
- `role="status"` + `aria-label="Loading band conditions"` on the spinner wrapper; spinner itself is `aria-hidden="true"`
- `role="alert"` on the stale-data badge so assistive technology announces immediately when the server is serving error-fallback data

**New `BandConditionsPanel.test.jsx`:** 7 tests covering each aria attribute.

## Test plan

- [x] All 7 new component tests pass (`BandConditionsPanel.test.jsx`)
- [x] Full suite: 26 test files, 503 tests — no regressions
- [x] Visual output unchanged — all aria changes are additive attributes / markup additions only
- [ ] Manual screen reader test (NVDA / JAWS / VoiceOver) recommended before merge

### DIFF
--- a/src/components/BandConditionsPanel.jsx
+++ b/src/components/BandConditionsPanel.jsx
@@ -30,6 +30,7 @@ export const BandConditionsPanel = ({ data, loading, extras }) => {
         {t('band.conditions')}
         {staleMinutes != null && (
           <span
+            role="alert"
             title={t('band.conditions.stale.tooltip', { mins: staleMinutes })}
             style={{
               fontSize: '10px',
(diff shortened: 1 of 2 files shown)


accius (approved):
Good, focused panel pass. The pattern is right: aria-label on each cell carrying the full "40m: Good" meaning, visual children marked aria-hidden so the band name and condition do not get read twice, role=status on the loading spinner, and the polite live region on the grid so the hourly refresh announces without interrupting. The seven tests pin all of it, including the no-double-read assertion.

One non-blocking nit: role=alert on the stale badge is assertive, so it interrupts whatever the user is doing the moment data goes stale. Staleness is informational rather than urgent, and role=alert on a node that stays mounted while stale can re-announce on re-render in some screen readers. role=status (polite) would be the gentler and more conventional choice there. Not worth holding the merge.

CI green.

K0CJH

PR #1092, opened 2026-06-19
Update to support Chinese language (adding registration in i18n.js)

Oukagen:
Hi,

I previously submitted the zh.json file, but it didn't show up as an option in the UI because I hadn't registered it in src/lang/i18n.js.

I have now updated i18n.js to correctly import the Chinese translation file and register it within the LANGUAGES array and resources object. This should now enable the Chinese language option in the settings.

Thank you for your time and for maintaining this project!


MichaelWheeley:
@Oukagen please change target of PR from accius:main to accius:Staging
p.s. click on the pencil/edit button as if you were going to edit the title of this PR, then the target becomes available to change

accius (commented):
The registration itself is correct: the import, the LANGUAGES entry, and the resources entry all line up, and zh.json is present in the tree so the import resolves. This is the missing half of the earlier zh.json submission and it does enable the option in settings.

One process note rather than a code issue: this PR targets `main`. Contributions to this project go through the `Staging` branch first and reach main on the normal release schedule. Could you retarget this to `Staging`. The build-and-push check also shows cancelled here, which a fresh run against Staging would clear up.

K0CJH

accius:
this PR has been updated to target staging branch correctly, however there are conflicts now.  can you look at the conflicts and update the PR?  Thank you! once that's done we will be able to push this to staging.

Oukagen:
Thank you both for your help! I'm not very familiar with GitHub and coding, so I really appreciate your patience. I've resolved the conflicts as you asked. Please let me know if there's anything else that needs to be fixed.😸 @MichaelWheeley @accius

PR #1096, opened 2026-06-21
Any time we use pass a callsign over http we need to encodeURIComponent()

### LINKED ISSUE #1095: [BUG] RBN should query spot/cluster caches before querying qrz for location
## Describe the Bug

I just went to check the RBN as I had VK2IO/P operating from a WWF Park (spotted for WWFF on spot line), but the display of the RBN went directly to QRZ and showed the location of his QTH rather than the spot.

## To Reproduce

FInd  CW operator who has a spot out (any spot) and ch

## Expected Behavior

Before looking up QRZ dot COM, the RBN code should query the spot, and dx cluster caches for this callsign and band/frequency.

## Screenshots

Image

Image

The inverted triangle is his 160m CW spot from VKFF-1318. RBN is showing his home QTH at QF56ob

## Additional Context

Add any other context about the problem here. Include your callsign if you're comfortable sharing it!

---

73!

alanhargreaves:
## What does this PR do?

Any time we pass a callsign to the server API in `useRBN.js` we need to encode the callsign so callsigns with suffixes like `/P` don't screw things up.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Watch an RBN skimmer for a known CW op operating with a suffix like `/P` that has also spotted in either DX Cluster or an ActivatorPanel type spot and verify that we get the proper location.

## Checklist

### DIFF
--- a/src/plugins/layers/useRBN.js
+++ b/src/plugins/layers/useRBN.js
@@ -277,7 +277,7 @@ export function useLayer({
           mySpots.map(async (spot) => {
             if (spot.grid && spot.skimmerLat != null && spot.skimmerLon != null) return spot;
             try {
-              const locationResponse = await fetch(`/api/rbn/location/${spot.callsign}`);
+              const locationResponse = await fetch(`/api/rbn/location/${encodeURIComponent(spot.callsign)}`);
               if (locationResponse.ok) {
                 const loc = await locationResponse.json();
                 return {


alanhargreaves:
I'm running this locally to verify that it addresses the issue. If to this evening, I expect to see VK2IO/P activating in the morning on CW.

alanhargreaves:
ok it looks like `/api/callsign` does not check our caches. It probably should.

accius (commented):
Correct fix. Line 280 was the only callsign-bearing fetch in this file still interpolating raw, the user-callsign spots fetch at line 244 already runs through encodeURIComponent, so this brings the location lookup in line with it. A skimmer callsign with a `/` portable suffix would otherwise break the path segment.

One thing: this is still marked draft. Flip it to ready when you are happy with it and it can go in.

K0CJH

accius:
@alanhargreaves checking in on this one — the fix itself is plainly correct (portable suffixes like `VK2IO/P` break the unencoded path segment) and CI is green. Did your morning VK2IO/P test pan out? If so, feel free to mark it ready for review and we'll get it merged.

Your side-observation about `/api/callsign` not checking our caches was a good catch — I've split it out into its own issue so it doesn't get lost here.

PR #1098, opened 2026-06-21
Add SatNOGS transmitter metadata for satellites

AntDiClementi:
Adds satellite radio metadata from the SatNOGS Transmitter DB for tracked satellites.

Summary:
- Fetches active/alive SatNOGS transmitter records by NORAD ID
- Caches transmitter metadata for 24 hours
- Applies metadata to existing tracked satellites without replacing orbital data
- Selects one best/primary transmitter for the compact satellite popup
- Preserves curated fallback behavior when SatNOGS has no usable record
- Adds SatNOGS attribution text for CC BY-SA metadata
- Adds SatNOGS refresh status to the satellite debug endpoint

Tested locally:
- ISS shows FM downlink/uplink/tone metadata
- RS-44 shows Linear mode with uplink/downlink ranges
- TEVEL satellites avoid being mislabeled as Linear when they are FM single-frequency entries
- npm run build passes

### DIFF
--- a/server/routes/satellites.js
+++ b/server/routes/satellites.js
@@ -88,6 +88,8 @@ module.exports = function (app, ctx) {
   // SatNOGS Transmitter DB radio metadata.
   const SATNOGS_TRANSMITTER_CACHE_DURATION = 24 * 60 * 60 * 1000; // 24 hours
   const SATNOGS_TRANSMITTER_MAX_RECORDS = 500;
+  const SATNOGS_TRANSMITTER_REQUEST_TIMEOUT = 8000;
+  const SATNOGS_TRANSMITTER_CONCURRENCY = 4;
   const SATNOGS_TRANSMITTER_ATTRIBUTION = 'Radio metadata from SatNOGS Transmitter DB (CC BY-SA 4.0)';
 
   let satnogsTransmitterCache = {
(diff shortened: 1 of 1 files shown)


server/routes/satellites.js
```diff
@@ -890,6 +1211,12 @@ module.exports = function (app, ctx) {
       .sort((a, b) => a.norad - b.norad);
 
     res.json({
+      satnogsTransmitters: {
+        lastFetchAt: satnogsTransmitterCache.timestamp || null,
```
MichaelWheeley:
```suggestion
                ...(Number.isFinite(satnogsTransmitterCache.timestamp) && {
          lastFetch: formatSimpleAge(Date.now() - satnogsTransmitterCache.timestamp),
        }),
```

server/routes/satellites.js
```diff
@@ -85,6 +85,319 @@ module.exports = function (app, ctx) {
     logInfo(`[Satellites] Routing CelesTrak/AMSAT/SatNOGS fetches via ${FLETCHER_URL}`);
   }
 
+  // SatNOGS Transmitter DB radio metadata.
+  // OpenHamClock issue: https://github.com/accius/openhamclock/issues/1009
```
MichaelWheeley: for production code line probably not needed (`OpenHamClock issue: ...`)

AntDiClementi:
@MichaelWheeley Thank you, that line has been removed.

accius (commented):
Nice addition, the SatNOGS transmitter DB fills a real gap in the popup radio data. CI is green and the scoring/mode heuristics read well. A few things worth a look before this goes out:

1. The `/api/satellites/status` change has a guard bug. `satnogsTransmitterCache.timestamp` initializes to `0`, and `Number.isFinite(0)` is `true`, so `lastFetch` will be emitted even before the first refresh completes, reporting a bogus age (Date.now() minus 0). The earlier `lastFetchAt: timestamp || null` handled this correctly. Suggest guarding on `satnogsTransmitterCache.timestamp > 0` instead.

2. `refreshSatnogsTransmitterMetadata` fetches per-NORAD sequentially, each with a 20s abort timeout. On a cold start where the upstream is slow or failing, that loop can run for a long while across the whole registry. It is background and non-blocking so not a release blocker, but a short per-request timeout or a small concurrency batch would make the first populate a lot snappier.

3. `applySatnogsMetadataToRegistry` uses `Object.assign` onto the shared registry and ommCache entries but never clears prior SatNOGS fields. If a satellite later loses its active transmitter, the stale `mode`/`downlink`/`uplink` stick around until restart. Low priority given how rarely this data changes, but worth a comment so the next person knows it is intentional.

None of these block merge on their own. Item 1 is the one I would fix before this reaches users since it makes the status endpoint misreport.

K0CJH

AntDiClementi:
Thanks for the detailed review — I addressed the SatNOGS follow-ups.

Changes made:

* Fixed the status endpoint timestamp guard so `lastFetch` is only reported after a real successful SatNOGS transmitter refresh.
* Reduced the SatNOGS transmitter request timeout and changed the cold refresh path to fetch NORAD transmitter records in small concurrency batches instead of strictly sequentially.
* Updated the registry/OMM overlay behavior so previously applied SatNOGS radio fields are restored/cleared if a later refresh no longer finds usable active transmitter metadata for that satellite.

I also re-ran the local checks:

* `git diff --check`
* `node -c server/routes/satellites.js`
* `npm run build`

accius (commented):
Re-reviewed after the latest push and all three notes from the prior round are handled:

1. The status guard now keys off `satnogsTransmitterCache.timestamp > 0`, so `lastFetch` no longer reports a bogus age before the first refresh. Good.

2. The per-NORAD fetch is now batched with `SATNOGS_TRANSMITTER_CONCURRENCY = 4` and the per-request timeout dropped to 8s, so a cold start against a slow upstream will not crawl through the whole registry serially anymore.

3. Stale overlay clearing is in place now via `restoreSatnogsBaseMetadata` and the snapshot map, and the `cleared` counter gives a log signal when an overlay is removed. That was the one I figured would just get a comment, nice to see it actually handled.

Scoring and mode heuristics still read well and the attribution is carried through. CI is green and it is a clean merge into Staging. No further changes needed from me.

K0CJH

PR #1105, opened 2026-07-03
[feat] satellite backup source env enable/disable

MichaelWheeley:
## What does this PR do?

- provides enable / disable flags in `.env` for AMSAT TLE and SatNOGS TLE sources
- these backup sources are questionable data sources, they are only utilized as secondary sources when CelesTrak data source is enabled
- by default, both backup sources are enabled unless explicitly disabled in `.env`
- `.env.example` updated to include `AMSAT_TLE_ENABLED` and `SATNOGS_TLE_ENABLED` flags

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

in `.env`,
`CELESTRAK_ENABLED=true`
`AMSAT_TLE_ENABLED=false`
`SATNOGS_TLE_ENABLED=false`

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -186,6 +186,12 @@ LAYOUT=modern
 # CelesTrak.org data source is enabled by default, no username or password is needed.
 # CelesTrak can be disabled by setting CELESTRAK_ENABLED=false, but it is recommended to keep it enabled as a backup source.
 CELESTRAK_ENABLED=true
+# AMSAT TLE data source is enabled by default, no username or password is needed.
+# AMSAT TLE can be disabled by setting AMSAT_TLE_ENABLED=false, but it is recommended to keep it enabled as a backup source.
+#AMSAT_TLE_ENABLED=true
+# SATNOGS TLE data source is enabled by default, no username or password is needed.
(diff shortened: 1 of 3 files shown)


accius (approved):
Reviewed the fallback-chain gating end to end. All transitions into AMSAT_INIT (the two returns in CELESTRAK_INDIVIDUAL_INIT) and SATNOGS_INDIVIDUAL_INIT (CelesTrak chain, AMSAT backoff/finally, and the SATNOGS fetch loop-back) are correctly guarded, and with both sources disabled the machine falls back to START cleanly. config.js getters match the existing CELESTRAK_ENABLED pattern exactly.

Non-blocking nit: in .env.example the new lines are commented out (`#AMSAT_TLE_ENABLED=true`) while the sibling `CELESTRAK_ENABLED=true` is active — fine either way since default-enabled, just slightly inconsistent.

Merging for today's release. 73, K0CJH

PR #1136, opened 2026-07-23
feat: add PSK Reporter Band Activity overlay and Dockable panel

### LINKED ISSUE #1134: [FEATURE] - A real replacement for the classic HamClock Live Spots view

There's so much great and useful data in OpenHamClock, but the one thing I really used the Classic HamClock for is missing: live spots, that quickly at a glance show me which band is most likely to be successful for activation.

I'm not looking to see who is hearing me, I'm looking for easy band condition indicators that can inform me which band I should configure my home antenna (or which components to pack in my POTA/SOTA bag) before I even start calling. The data is out there and the problem has been solved by what is now a dead software stack.

I left some details in a discussion post here: https://github.com/accius/openhamclock/discussions/939

But the basic gist is this: an overlay on the map that shows me RBN/PSK spots based on my QTH grid square, broken down by band - and a count of live spots per band from my QTH. Classic HamClock has had this for years and it's the single most useful tool in my opinion.

## Describe Alternatives You've Considered

I've tried using the available map overlays, PSK Reporter integration, and have not come up with a solution that combines the tools present already to satisfy this need.

## Use Case

- [X] General HF operation
- [X] VHF/UHF operation
- [ ] Satellite operation
- [X] Contesting
- [X] POTA/SOTA
- [X] DXing

- [ ] Yes, I can submit a PR
- [X] Yes, I can help test
- [ ] No, but I'd use it!

I'm not familiar enough with this codebase to help implement the functionality - much as I'd like to make the time to get there, I have too many other projects ongoing. However, I would definitely test and provide real time feedback.

## Additional Context

You can see some screenshots in the linked discussion post above

Classic HamClock had this and it worked a treat.

---

73!

lbatalha:
Compact horizontal bar chart showing PSKReporter spot counts per HF band.

Features:
- Client-side counting from usePSKReporter data (filtered by time window)
- Real-time updates via custom events (no polling)
- Fixed panel size, all 15 HF bands always visible
- Inactive bands dimmed with reduced opacity
- Minimizeable with footer that shows total spots and time window
- Draggable panel with saved position
- Styled with band colors and consistent with other overlays

This should fix #1134

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [x] Map layer plugin

## How to test

1. Enable Map layer
2. See that spot counts show up per band and that total matches PSKReporter panel results

## Checklist

image


accius (changes requested):
Really nice work overall — this is exactly the right way to build on the PSK Reporter data: the overlay rides the existing MQTT→SSE pipeline with zero new upstream traffic, and the panel follows the layer-plugin conventions (draggable/minimize, storage keys, escaped interpolation). One required change and two optional ones:

**Required — please remove the `GET /api/pskreporter/band-activity` endpoint** (server/routes/pskreporter.js). Nothing in this PR calls it — the client gets its counts from the `psk-band-activity-changed` CustomEvent — and as written it's a footgun on shared/hosted instances: uncached, it iterates every subscriber's `recentSpots` buffer per unauthenticated request, and it double-counts spots that appear under multiple subscription keys, so the numbers it returns are wrong on any multi-user server anyway. (Your own checklist flags the missing caching.) The feature works perfectly without it — if we later want a server-side aggregate, it needs caching and per-subscriber dedup, which can be its own PR.

**Optional:**
1. Stale counts on quiet bands: the event only fires inside `processSpots`, so when no new spots arrive, expired spots stay counted in the panel until the next batch. A periodic re-emit (or filtering by spot age at render time) would keep the bars honest.
2. In the content-update effect, the `insertBefore(contentTarget, container.querySelector('.psk-band-footer'))` fallback would throw if ever reached — after `addMinimizeToggle` wraps children into `.psk-panel-content`, `.psk-band-footer` is no longer a direct child of `container`. Unreachable today, but worth fixing while you're in there.

Drop the endpoint and this is good to go.

lbatalha:
The API endpoint call was originally needed because more info was going to be displayed, but for just this change its definitely not needed, removed it.

The stale counts problem does happen if pskreporter has no data. I tried fixing it in various ways but could not figure it out, even converting the event to a useEffect, etc... so it might be out of my skill level and understanding of how React works.

lbatalha:
As per conversation in the Issue, I've added a Dockable mode panel that does the same as the overlay.

accius (approved):
Endpoint removed as requested, and the dockable panel follows the conventions cleanly with zero new server surface — the condition from my earlier review is satisfied. The overlay stale-count item is moving to a follow-up issue so it doesn't hold this up. Nice work, and thanks for the fast turnaround on the endpoint.

PR #1140, opened 2026-08-08
fix: remove container_name to prevent deployment conflicts

9M2PJU:
Removing the hardcoded `container_name` allows docker compose to generate a unique name (e.g. `openhamclock-openhamclock-1`), which prevents orphaned containers from blocking new deployments. This fixes an issue where an interrupted `docker compose down` leaves an orphaned container occupying the `openhamclock` name, breaking subsequent `docker compose up` commands with a name conflict error.


accius:
Thanks for this fix, @9M2PJU — the reasoning is spot-on. This PR branched from `main`, which has diverged from `Staging` (our release line), so it couldn't retarget cleanly. I've rebased your exact one-line change onto `Staging` with your authorship preserved and opened #1145 to carry it forward. Closing this in favor of that one — full credit to you. 🙏

PR #1142, opened 2026-08-09
feat: add opt-in 630m WSPR and RBN support

### LINKED ISSUE #1091: [FEATURE] Bands
No

No problem,  everything is working fine.  I would like to see if it possible to add 630 meters to the bands? There is daily WSPR  and FST4 activity in the US overnight and in the mornings..

To see WSPR and RBN spots on 630 meters

## Describe Alternatives You've Considered
Just an enhancement.

## Use Case

- [ ] General HF operation
- [ ] VHF/UHF operation
- [ ] Satellite operation
- [ ] Contesting
- [ ] POTA/SOTA
- [ ] DXing
- [x ]  it would increase the overall usefulness.

- [ ] Yes, I can submit a PR
- [ x] Yes, I can help test
- [ x] No, but I'd use it!

## Additional Context
Thanks for a great program
Add any other context, screenshots, or mockups about the feature request here.

Are there similar features in other amateur radio software? If so, what do you like/dislike about their implementation?

---

73!
Tom
N9RU

AntDiClementi:
PR title:
Add opt-in 630m WSPR/FST4W and RBN support

PR body:

Closes #1091

## Summary

Adds opt-in 630m (472–479 kHz) support while preserving the existing default behavior for users who do not enable it.

- Adds 630m frequency classification for MHz, kHz, and Hz inputs.
- Fixes sub-1 MHz normalization so values such as 474.2 kHz are not interpreted as 474.2 MHz.
- Adds 630m WSPR/FST4W activity using the MF data available from WSPR.live while retaining the existing PSK Reporter path for normal WSPR bands.
- Adds 630m support to RBN classification/filtering.
- Keeps upstream WSPR sources failure-isolated so 630m data cannot blank normal-band WSPR data, or vice versa.
- Preserves propagation paths per band so 630m paths remain visible when the 630m map legend chip is selected.
- Adds a customizable 630m band color and conditionally displays the 630m chip before 160m in the bottom map legend.
- Uses the bottom map legend as the WSPR band selector rather than maintaining a redundant Band dropdown in the WSPR panel.
- Adds conditional 630m awareness to the relevant band-filter surfaces.
- Keeps 630m disabled by default.
- Does not add 2200m support and does not include 630m in P.533/VOACAP calculations.

## Testing

- `git diff --check`
- `npm run format:check` / repository-pinned Prettier
- `npx --no-install vitest run src/utils/band630m.test.js`
- `npm run build`
- Server-side 630m boundary tests
- Manual WSPR testing at 15 min, 30 min, 1 hour, 2 hours, and 6 hours
- Manual verification of 630m propagation lines and bottom-legend filtering
- Manual verification that the default state remains unchanged when 630m is disabled

## Notes

630m is intentionally opt-in so the existing band list and map behavior remain unchanged by default.


accius:
Thanks for this, @AntDiClementi — the 630m work is genuinely well done. The per-source failure isolation (`wspr` vs `wspr630m` backoff buckets) and the per-band path keying are exactly right, and I appreciate the boundary tests. Opt-in/default-off is the correct call.

One thing I'd like to sort out before merging: the PR bundles a sizeable **refactor of the WSPR plugin UI** in with the band feature — the four separate Leaflet controls (filter / stats / legend / chart) get consolidated into a single combined `WSPRControl` panel in `useWSPR.js`. That's ~350 lines and it changes the default map layout for **every** user, not just those who enable 630m. A couple of knock-on effects:

- It silently orphans three saved layout keys (`wspr-stats-position`, `wspr-legend-position`, `wspr-chart-position`), so anyone who's dragged those panels loses their positions with no migration.
- Because this lands on `Staging` and ships in the next release, a UX change here reaches everyone immediately — and if something regresses, it's harder to bisect when it's tangled up with the band classification work.

Could we **split the panel consolidation into its own PR**? That keeps #1142 as a clean, reviewable "opt-in 630m" change that's easy to reason about and revert if needed, and lets the UI refactor stand on its own merits (I'm not against the consolidation — it looks clean — it just deserves its own review and a WhatsNew line).

If splitting is more trouble than it's worth, the fallback would be to at least call the panel refactor out explicitly in the PR description and note the orphaned position keys so we can decide whether to migrate them. Happy to help either way.

A few small non-blockers I noticed, for whenever (not gating): `normalizeFrequencyToMHz` is now triplicated — `bandColors.js` could import it from `callsign.js`; and in `fetchWsprLive630m` the `modeByCode` map has codes 5 and 16 both → `FST4W-900`, worth a quick double-check.

Thanks again — this is close.

accius:
This is exactly what I was hoping for. The descope landed clean, the Filters panel keeps its layout keys, and the 630m checkbox living in the existing panel is the right home for it. Merging to Staging now, it'll ride the next release.

If you still want to pursue the panel consolidation, open it as its own PR and we'll give it a proper look on its own merits.

K0CJH

PR #1146, opened 2026-08-13
Update Malay translations

9M2PJU:
Translated missing UI strings and refined existing translations in `src/lang/ms.json`.


MichaelWheeley:
@9M2PJU 
from the command line please run
```npm run lang:sort```
and then check in changes to `src/lang/ms.json` again

src/lang/ms.json
MichaelWheeley, line 1: key order modified, will not pass CI workflow, use
```npm run lang:sort```
prior to checking in changes

MichaelWheeley (changes requested):

PR #1150, opened 2026-08-13
scripts: Rename setup-linux.sh to setup.sh

9M2PJU:
setup.sh is a simpler and more standard name, especially since it supports macOS, FreeBSD, and generic Unix setups in addition to Linux. This updates all references to it.


accius:
The rename itself is fine by me, setup.sh is the better name now that the script covers more than Linux. Two things before this goes in:

1. The old raw URL is load-bearing. There are blog posts, forum threads, and video descriptions out there piping `scripts/setup-linux.sh` straight into bash, and a pure rename 404s all of them the moment this ships to main. Can you keep `scripts/setup-linux.sh` as a tiny shim that just execs `setup.sh` (a one-line deprecation echo would be a nice touch)?

2. Merge order: #1148 just went into Staging and touches this same script, so you'll want to merge Staging into this branch so the rename carries the FreeBSD changes along.

The `setup-linux.sh` mentions still in CHANGELOG.md and WhatsNew.jsx are historical entries, leave those as is.

K0CJH

9M2PJU:
Thanks @accius! Updated:

1. Added `scripts/setup-linux.sh` as a backwards-compatibility shim with a deprecation notice that delegates execution to `scripts/setup.sh` (supporting both local execution and piped `curl`/`wget` invocations with arguments passed through).
2. Merged latest `Staging` into this branch so FreeBSD support from #1148 is carried over into `scripts/setup.sh`.
3. Kept historical references in `CHANGELOG.md` and `WhatsNew.jsx` intact.

PR #1151, opened 2026-08-16
Feature/3d globe

KoenVdH:
## What does this PR do?

Adds a third map projection alongside Flat (Mercator) and Azimuthal: a **3D globe** rendered with three.js.

The projection toggle becomes `Flat · Azimuthal · 3D`, and the choice persists in `localStorage` like the existing two. Everything the other projections show is carried over — DX cluster spots and spotter paths, POTA/WWFF/SOTA/WWBOTA activations, PSKReporter, WSJT-X, the DE/DX markers and the band filter.

Because this projection is not backed by Leaflet, two things are done natively rather than reprojected:

- **Great circles are true 3D arcs** (slerp between unit vectors), bowed proportionally to path length. No antimeridian special-casing is needed, which is most of the complexity in the flat-map version.
- **The day/night terminator is a fragment shader** on the sphere, driven by the subsolar point from `getSunPosition()`. It respects the existing night-darkness slider, sharing that slider's state and its `ohc_nightDarkness` key with the flat map — so the same percentage means the same thing in both, and changing it in one carries to the other.

### Notable implementation details

**Mercator tiles cannot wrap a sphere.** `SphereGeometry` maps texture V linearly to latitude, while tiles are Mercator. `tileReproject.js` already handles this per-pixel while sampling for the azimuthal canvas; for a GPU texture it has to be baked once, so `src/utils/globeTexture.js` composites the tile mosaic and row-remaps it to equirectangular.

**Dark basemaps need a lift on a sphere.** CARTO's dark style measures a mean luminance of ~0.08 against satellite's ~0.31, which renders as a near-black ball. The texture builder reports its mean luminance and the shader lifts it toward a legible range. The multiplier is clamped to a floor of 1, so it only ever brightens and leaves already-bright imagery untouched.

**Tile fetches are throttled to 6 concurrent**, matching what `tileReproject.js` settled on. Requesting all 64–256 at once drew HTTP 429s from the providers, and a rate-limited tile is a hole in the texture.

**three.js is lazy-loaded** via `React.lazy`, landing in its own ~557 kB chunk (142 kB gzipped). Operators who never open 3D mode do not pay for it; the main bundle is unchanged.

**The Countries style** ships transparent overlay tiles rather than a basemap, so it gets the same blue backdrop the flat map paints behind it.

### Known limitation

**Leaflet-bound plugin layers — satellites, aurora, lightning — do not render in 3D mode.** They attach to a Leaflet map instance and there isn't one. `WorldMap` suppresses them while this projection is active rather than binding them to a hidden map, so nothing breaks silently, but the capability genuinely isn't there. Porting them would mean reimplementing each as a three.js overlay, which felt like a separate piece of work rather than something to bundle in here.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. `npm ci && node server.js` (backend) and `npm run dev` (frontend), then open the dashboard.
2. Click **3D** in the projection toggle at the top of the map. The globe opens centred on your QTH and slowly auto-rotates (~84 s per revolution).
3. **Drag** to rotate, **scroll** to zoom, **click** the globe to set DX, **hover** a spot for a tooltip with callsign, frequency, band and spotter.
4. Use the control column: **DE** re-centres on your QTH keeping the current zoom, **DX** centres on the DX target, the **house icon** resets to the default view, the **arrows icon** toggles auto-rotate. The vertical slider is the night-overlay darkness, shared with the flat map.
5. Switch basemaps in the style dropdown (Dark, Satellite, Hybrid, Countries, Streets…) — each rebuilds the globe texture with a progress readout.
6. Filter by band in the legend and confirm markers and arcs thin accordingly.
7. Switch themes (Dark / Light / Retro) and confirm the backdrop follows: white on Light, grey on Retro, space with stars on Dark.

## Checklist

- [ ] If touching `server.js`: caches have TTLs and size caps (we serve 2,000+ concurrent users) — *n/a, `server.js` untouched*
- [ ] If adding an API route: includes caching and error handling — *n/a, no API routes added*
- [ ] If adding a panel: wired into Modern, Classic, and Dockable layouts — *n/a, this extends the existing map panel rather than adding one*
- [x] No hardcoded colors — uses CSS variables (`var(--accent-cyan)`, etc.) — **with one deliberate exception, see below**

### On the colour checkbox

All UI chrome (buttons, loading indicator, readout, tooltip, slider label) and the DE/DX station accents come from CSS variables. WebGL materials cannot reference `var()`, so those read the variables at runtime and rebuild when the theme changes.

The **activity-type marker palette** (POTA / WWFF / SOTA / WWBOTA / PSK / WSJT-X) is deliberately left unthemed, in a documented constant with the reasoning inline. Those colours identify *what a spot is*, and `WorldMap` and `AzimuthalMap` use the same literals — theming them here would make the same POTA spot render differently depending on which projection you're viewing. Band colours are centralised in `bandColors.js` and left alone for the same reason. Happy to change this if you'd rather it follow the theme.

## Screenshots
image
image
image
image


KoenVdH:
Hi @accius , hi Chris,
are you willing to add this feature to OpenHamClock?
If you do, I will further refine and extend the implementation.
For now I wait.
73, Koen ON9PK

MichaelWheeley:
@KoenVdH satellite tracking you just added looking good! is it possible to include this dialog in the view, it disappears in 3D view,
image

accius:
@KoenVdH Short answer: yes. A 3D globe has been on the wish list for a long time and what you've built here looks genuinely well thought out, the shared night-darkness slider and the true great-circle arcs especially. Keep refining.

I'll do a full technical review pass on the diff, so expect some inline comments. The main things I'll be looking at are the three.js bundle weight, teardown behavior when switching projections (leaks, orphaned WebGL contexts), and how it behaves in low-memory mode on Pi-class hardware. Since you're still pushing to the branch, ping me when you consider it review-ready, otherwise I'll just review what's there.

73, K0CJH

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+  }, [dxPaths, showDXPaths, bandPassesMapFilter, dxLocation, lat0, lon0, themeTick, showDeDxMarkers]);
+
+  // ── Scene setup (once) ───────────────────────────────────
+  useEffect(() => {
+    const container = containerRef.current;
+    if (!container) return undefined;
+
+    const scene = new THREE.Scene();
+    const camera = new THREE.PerspectiveCamera(45, 1, 0.01, 100);
+    // Open looking straight down on the operator's QTH. If the location has not
```
accius, line 611: **Blocker.** If the WebGLRenderer constructor throws, this catch logs and returns, but nothing throws during render, so the AzimuthalErrorBoundary in WorldMap never fires and its mercator fallback never runs. The texture effect still downloads the full tile set, then bails at the `if (!s.earthMat) return;` guard before `setTextureLoading(false)`, so the user sits on "Loading globe 100%" forever. And because the projection choice is persisted, the dead panel comes back on every reload. This is exactly what the Pi fleet will hit (kiosk without GPU accel, VNC sessions, GPU-blocklisted browsers). Suggest setting an error state here that rethrows during render so the boundary catches it, or an onUnsupported callback so WorldMap can fall back, and gate the texture fetch on the renderer existing.

src/components/WorldMap.jsx
```diff
@@ -2408,62 +2413,124 @@ export const WorldMap = ({
         </AzimuthalErrorBoundary>
       )}
 
+      {/* 3D globe (three.js / WebGL) */}
+      {isGlobe3D && (
+        <AzimuthalErrorBoundary label="Globe3D" onFallback={() => setMapProjection('mercator')}>
```
accius: **Blocker.** Two problems stacked here. React 18 permanently caches a rejected `lazy()` import, so if the Globe3D chunk fails to load once (stale HTML after a redeploy is the classic case), every later click of 3D re-throws instantly from the cached rejection until a hard reload. And this onFallback writes mercator into the persisted map settings via the save effect, so one network blip silently clobbers the user's projection preference with no retry and no visible error. Suggest a retry-aware lazy wrapper (re-import on next mount after a failure) and having the fallback switch the session's projection without persisting it.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+      // On a narrow panel WorldMap's projection toggle spans the full width and
+      // would sit on top of our control column, so the column drops below it.
+      setPanelWidth(w);
+    });
+    ro.observe(container);
+
+    // A fresh scene starts still; the screensaver clock decides when it turns.
+    kickIdleTimer();
+
+    return () => {
```
accius, line 768: **Blocker.** This effect tears down the renderer and rebuilds the canvas and groups when lowMemoryMode changes, but the markers/arcs effect, the satellites effect, and the pointer effect do not depend on lowMem. Config loads async, so lowMemoryMode flips after mount on every Pi that has it saved: the new scene comes up empty and the click/hover handlers stay bound to the detached old canvas. Either thread lowMem into the downstream deps or key the component on lowMemoryMode so a change remounts everything cleanly.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+        });
+        return;
+      }
+
+      el.style.cursor = downAt ? 'grabbing' : 'grab';
+      setTooltip(null);
+    };
+
+    const onUp = (ev) => {
+      el.style.cursor = 'grab';
```
accius, line 1321: **Blocker.** The point-cloud hit test never compares against the earth intersection distance, so markers on the far side of the globe steal hovers and clicks straight through the sphere. Click ocean whose antipode holds a POTA spot and onSpotClick/onDXChange fires for a marker you cannot see. Cheapest fix: intersect the earth mesh first and discard any point hit whose distance is greater than the earth hit. Same treatment for the hover path above and the satellite hits.

src/components/WorldMap.jsx
```diff
@@ -2408,62 +2413,124 @@ export const WorldMap = ({
...
+              showWWBOTA={showWWBOTA}
+              showPSKReporter={showPSKReporter}
+              showWSJTX={showWSJTX}
+              onSpotClick={onSpotClick}
+              callsign={callsign}
+              showDeDxMarkers={showDeDxMarkers}
+              satellites={satellites}
+              satellitesEnabled={pluginLayerStates.satellites?.enabled ?? true}
+              hideUi={mapUiHidden}
+              tileStyle={mapStyle}
```
accius, line 2520: Base-map rotation keeps cycling mapStyle while the globe is up, and every tick makes Globe3D refetch the full tile set (up to 256 tiles at retina zoom) with a loading-overlay flash. Since the abort path only stops dequeuing, back-to-back rotations overlap fetch bursts, the provider starts throttling, and throttled tiles render as permanent holes in the sphere (globeTexture's own comment warns about exactly that). Please pause rotation while isGlobe3D, or at minimum cache built textures per style so a revisit is free. We try hard to be good neighbors to the tile providers.

src/components/WorldMap.jsx
```diff
@@ -2408,62 +2413,124 @@ export const WorldMap = ({
...
-            locator={deLocator}
-            deLat={deLocation?.lat ?? null}
-            deLon={deLocation?.lon ?? null}
-            lowMemoryMode={lowMemoryMode}
-          />
-        );
-      })}
+      {/* Plugin layers attach to a Leaflet map instance, so they cannot render on
+          the 3D globe — skip them entirely rather than binding to a hidden map. */}
+      {!isGlobe3D &&
```
accius, line 2572: Everything except satellites is skipped under !isGlobe3D, but the layer toggles in Settings stay live and persist enabled:true, with nothing telling the user 3D does not support them. They flip Lightning on, nothing happens, and the layer then pops in "unexpectedly" when they switch back to Flat. For this PR I would settle for a visible note when 3D is active (layer is 2D-only for now); actually porting layers can be follow-up work.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+        if (!Number.isFinite(lat) || !Number.isFinite(lon)) return;
+        // PSKReporter reports freq in Hz; freqMHz is the pre-converted variant.
+        const freqMHz = s.freqMHz || (s.freq ? s.freq / 1e6 : null);
+        const band = normalizeBandKey(s.band) || bandFromAnyFrequency(freqMHz || s.freq);
+        if (!bandPassesMapFilter(band)) return;
+        const isRx = s.direction === 'rx';
+        out.push({
+          lat,
+          lon,
+          color: isRx ? ACTIVITY_COLORS.pskRx : ACTIVITY_COLORS.pskTx,
```
accius, line 514: Two field mismatches against the actual WSJT-X decode shape. First, `s.freq` is the FT8 audio offset in Hz (the server sets `freq: msg.deltaFreq`), and getBandFromFreq treats values >= 1000 as kHz, so a 1800 Hz offset classifies as 160m and the band filter hides or mislabels 20m decodes. The Leaflet path derives band from `dialFrequency / 1e6`. Second, the label a few lines down reads `s.dxCall || s.call || s.callsign`, but decodes carry caller/deCall/dxCall, so CQ decodes (the majority) all render the literal label "WSJT-X", and QSO decodes plot at the sender's grid while naming the addressee. Mirror WorldMap's expressions for both.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+      localStorage.setItem(AUTOROTATE_KEY, String(autoRotate));
+    } catch {}
+    // Enabling arms the 30 s clock; disabling stops any rotation immediately.
+    kickIdleTimer();
+    return () => clearTimeout(idleTimerRef.current);
+  }, [autoRotate, kickIdleTimer]);
+
+  // ── Texture: rebuild when the map style changes ──────────
+  useEffect(() => {
+    const style = MAP_STYLES[tileStyle]?.url ? tileStyle : 'dark';
```
accius, line 822: When the style has no static url (MODIS is built dynamically by the other two projections), this silently swaps in the dark basemap while the dropdown still claims MODIS, and the GIBS day-offset slider still renders over the globe as a dead control (its gate has no isGlobe3D exclusion). Either pass the dynamic GIBS url and gibsOffset the way AzimuthalMap gets them, or exclude MODIS from the style picker in 3D and hide the slider.

src/components/WorldMap.jsx
```diff
@@ -2408,62 +2413,124 @@ export const WorldMap = ({
...
+            <Globe3D
+              deLocation={deLocation}
+              dxLocation={dxLocation}
+              onDXChange={onDXChange}
+              dxLocked={dxLocked}
+              potaSpots={potaSpots}
+              wwffSpots={wwffSpots}
+              sotaSpots={sotaSpots}
+              wwbotaSpots={wwbotaSpots}
+              dxPaths={dxPaths}
```
accius: The Leaflet path renders `filterDXPaths(dxPaths, dxFilters)`; the globe gets raw dxPaths and only applies the band filter. So mode/continent/watchlist filters silently stop working the moment the user switches to 3D and every suppressed spot reappears. Pass the filtered list here, or pass dxFilters through and filter inside.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+        blending: THREE.AdditiveBlending,
+        side: THREE.BackSide,
+        transparent: true,
+        depthWrite: false,
+      }),
+    );
+    scene.add(atmosphere);
+
+    const stars = lowMem ? null : makeStarfield();
+    if (stars) {
```
accius, line 692: Fixed 0.02 world-unit threshold against dots that render at constant screen-pixel size: zoomed out to maxDistance the clickable radius subtends 2-3 px while the dot draws 7-9 px, so clicks on a clearly visible spot fall through to the earth intersect and silently retarget DX to whatever coordinate is under the cursor. Zoomed in it balloons to tens of pixels and grabs neighbors. Scale the threshold with camera distance so it tracks the on-screen dot size.

src/utils/globeTexture.js
```diff
@@ -0,0 +1,171 @@
...
+  return new Promise((resolve, reject) => {
+    if (signal?.aborted) {
+      reject(new Error('aborted'));
+      return;
+    }
+    const img = new Image();
+    img.crossOrigin = 'anonymous'; // required, or the canvas taints and WebGL upload fails
+    img.onload = () => resolve(img);
+    img.onerror = () => reject(new Error(`tile failed: ${url}`));
+    img.src = url;
```
accius, line 64: loadTile only checks the signal before setting img.src, so an abort does not cancel anything in flight. A style change mid-load leaves a burst of dead downloads competing with the new set. Attach an abort listener that clears img.src, or tag loads with a generation token and drop stale results.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+    }
+  });
+  // config.lowMemoryMode arrives asynchronously as undefined then false;
+  // normalising stops that flip from rebuilding the entire WebGL scene.
+  const lowMem = !!lowMemoryMode;
+  const [panelWidth, setPanelWidth] = useState(0);
+  // Satellite selection, shared with the Leaflet layer via sessionStorage.
+  const [selectedSats, setSelectedSats] = useState(readSelectedSats);
+  const toggleSatSelection = useCallback((name) => {
+    setSelectedSats((prev) => {
```
accius, line 389: Follow-up: before deLocation arrives, the DE marker and its arcs draw at 0,0 in the Gulf of Guinea. Guard on a real location the way the Leaflet path does.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+  const b = latLonToVec3(lat2, lon2, 1);
+  const angle = a.angleTo(b);
+  const pts = [];
+
+  if (angle < 1e-6) return [a.clone().multiplyScalar(EARTH_R * MARKER_ALT)];
+
+  // Matches the QSO plotter's profile: a floor so short hops still stand off
+  // the surface, ramping to a high arc by ~18000 km.
+  const lift = 0.03 + 0.15 * Math.min(1, (angle * 6371) / 18000);
+  const sinAngle = Math.sin(angle);
```
accius, line 73: Follow-up: sinAngle goes to zero for antipodal pairs, the weights blow up, and the path degenerates to a chord through the middle of the globe. Rare, but antipodal long-path contacts are exactly the ones hams brag about. Pick an arbitrary intermediate waypoint when angle is within epsilon of pi.

src/components/WorldMap.jsx
```diff
@@ -598,6 +600,9 @@ export const WorldMap = ({
   const [mapRotationMenuActivity, setMapRotationMenuActivity] = useState(0);
   const [mapProjection, setMapProjection] = useState(initialProjection);
   const isAzimuthal = mapProjection === 'azimuthal';
+  const isGlobe3D = mapProjection === 'globe3d';
```
accius, line 637: Follow-up: the persisted projection token is trusted as-is. Validate against the known set on load so junk in localStorage cannot wedge the map, and so anyone who tries this build and then downgrades does not leave 'globe3d' behind for older code to trip on.

src/components/WorldMap.jsx
```diff
@@ -598,6 +600,9 @@ export const WorldMap = ({
...
+  // Both non-Leaflet projections hide the Mercator map and its dock.
+  const isLeafletHidden = isAzimuthal || isGlobe3D;
```
accius, line 639: Follow-up: hiding Leaflet does not stop it. All the 2D layer effects keep churning markers underneath while 3D is up, so a Pi pays for two renderers at once. Worth short-circuiting the layer effects when isLeafletHidden.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+  useEffect(() => {
+    const s = gl.current;
+    if (s.stars) s.stars.visible = isDarkBackdrop;
+    if (s.atmosphere) s.atmosphere.visible = isDarkBackdrop;
+  }, [isDarkBackdrop]);
+
+  // ── Auto-rotate toggle ───────────────────────────────────
+  useEffect(() => {
+    autoRotateEnabledRef.current = autoRotate;
+    try {
```
accius, line 809: Follow-up: {lang} never gets substituted here, so non-English users get English basemap labels only in 3D. The other projections pass the language through; do the same.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+      c.set(sat.color || accentCyan);
+      colors[i * 3] = c.r;
+      colors[i * 3 + 1] = c.g;
+      colors[i * 3 + 2] = c.b;
+      sizes[i] = selectedSats.includes(sat.name) ? 13 : 8;
+    });
+
+    const geo = new THREE.BufferGeometry();
+    geo.setAttribute('position', new THREE.BufferAttribute(positions, 3));
+    geo.setAttribute('color', new THREE.BufferAttribute(colors, 3));
```
accius, line 1105: Follow-up: this shader is missing the `#include <colorspace_fragment>` that the spot-dot shader it was copied from has, so satellite dots render in a different color space and visibly drift from the palette.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+}
+
+/**
+ * Activity-type marker palette.
+ *
+ * Deliberately not themed: these identify what a spot *is*, and a POTA marker
+ * has to read identically in Flat, Azimuthal and 3D. WorldMap and AzimuthalMap
+ * use these same values, so re-theming them here would desynchronise the
+ * projections. Band colours are centralised in bandColors.js for the same
+ * reason. Station accents (DE/DX) and all UI chrome do follow the theme.
```
accius, line 162: Follow-up: the comment says these match the other projections, but the WWFF/SOTA/WWBOTA hexes all differ from the Leaflet values. Import the shared constants instead of redeclaring them so they cannot drift again.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+  // Bumped on every theme change so the WebGL overlays — whose colours were
+  // resolved from CSS variables at build time — get rebuilt with the new ones.
+  const [themeTick, setThemeTick] = useState(0);
+  useEffect(() => {
+    const update = () => {
+      setIsDarkBackdrop(backdropIsDark());
+      setThemeTick((n) => n + 1);
+    };
+    update();
+    const mo = new MutationObserver(update);
```
accius, line 430: Follow-up: this observer fires per mutation with no debounce, and a theme color-picker drag mutates style attributes every frame, so themeTick triggers a scene rebuild storm during the drag. Debounce or rAF-coalesce the update.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+      pointer,
+      dotTexture: makeDotTexture(),
+      markerData: [],
+      disposables: [],
+    };
+
+    // ── Render loop ────────────────────────────────────────
+    let raf = 0;
+    const tick = () => {
+      raf = requestAnimationFrame(tick);
```
accius, line 733: Follow-up: this renders at 60 fps unconditionally even when nothing moves. On Pi-class hardware that is real heat for a static globe. Render on demand (controls change, auto-rotate tick, texture and overlay updates), or at least drop to a low idle rate when auto-rotate is off.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+              color: sat.isVisible ? accentGreen : accentCyan,
+              transparent: true,
+              opacity: 0.8,
+              depthWrite: false,
+            }),
+          ),
+        );
+      }
+    });
+  }, [satellites, satellitesEnabled, selectedSats, themeTick, isDarkBackdrop]);
```
accius: Follow-up: satellites updates every 5 s and this effect rebuilds all geometries and materials from scratch each time. Update the position attributes in place and set needsUpdate instead of reallocating.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1532 @@
...
+    const w2 = Math.sin(t * angle) / sinAngle;
+    const p = new THREE.Vector3(a.x * w1 + b.x * w2, a.y * w1 + b.y * w2, a.z * w1 + b.z * w2);
+    p.normalize().multiplyScalar(EARTH_R * (MARKER_ALT + lift * Math.sin(t * Math.PI)));
+    pts.push(p);
+  }
+  return pts;
+}
+
+// ── Band helpers (mirrors AzimuthalMap) ────────────────────
+const normalizeBandKey = (band) => {
```
accius, line 88: Follow-up: this is the fourth verbatim copy of normalizeBandKey/bandFromAnyFrequency in the codebase. Extract to src/utils and import it everywhere.

accius (changes requested):
Full review pass as promised. The bones here are genuinely good: the arc math, the terminator shader, and the texture assembly are clean, and lazy-loading the three.js chunk means people who never click 3D never pay for it, so bundle weight turns out to be a non-issue as structured.

I went deep because this targets Staging and our install base skews toward Pi-class hardware. 22 findings, all verified against the branch head. Grouped by what I need before merge vs what can wait:

**Blockers (fleet safety):**
- Swallowed WebGL init failure leaves a permanently dead, persisted pane (Globe3D.jsx:588)
- Cached lazy-import rejection plus the fallback silently clobbering the saved projection (WorldMap.jsx:2418)
- lowMemoryMode scene rebuild strands the marker/satellite/pointer effects on a detached canvas (Globe3D.jsx:744)
- Far-side markers steal clicks and hovers through the sphere (Globe3D.jsx:1272)

The first two combine badly with projection persistence: a Pi that cannot do WebGL gets wedged on a dead panel across reloads with no way back.

**Fix before ship:** WSJT-X field mapping (band and labels), dxFilters parity, the MODIS silent fallback with its dead GIBS slider, rotation-driven tile refetching (this one is also a tile-provider good-neighbor concern), the distance-blind click threshold, in-flight tile loads surviving abort, and some visible indication that the other plugin layers are 2D-only for now.

**Follow-up material** (fine to defer, and I am happy to file these as issues so they do not block you): Null Island DE guard, antipodal slerp degeneracy, projection token validation, {lang} passthrough, the satellite shader colorspace, ACTIVITY_COLORS drift, theme observer debounce, render-on-demand instead of a 60 fps loop, satellite geometry reuse, Leaflet churn while hidden, and deduping the band helpers.

None of this is architectural. The structure is right, it is the integration seams and the failure paths that need tightening. Ping me when you push and I will re-review quickly.

K0CJH

KoenVdH:
**Update since the PR was opened** — the branch has grown a bit beyond the original description, so a summary of what changed rather than making you diff the diff:

**The "Known limitation" section is now partly stale — satellites work in 3D.** It turned out the Leaflet plugin was only a renderer: the positions, ground tracks and footprint radii all come from the map-independent `useSatellites` hook, so the globe now draws that same data natively rather than needing the plugin. Dots sit at the satellite's true altitude above the sphere (with a nadir line down to the ground point), the orbit track rides at that altitude too, and the footprint ring stays on the surface where it belongs. Selection is shared with the flat map through the same `sessionStorage` key, so a satellite picked in Flat mode stays highlighted in 3D and back. Clicking a satellite toggles it; hovering shows name, altitude, az/el and mode. Aurora and lightning remain unported, and the satellite telemetry window / text labels are not reproduced — hover tooltips carry the identification.

**Styling: markers and arcs now hold a constant on-screen size** at any zoom level (the equivalent of `sizeAttenuation: false`, done in the shader since sizes vary per marker), with arcs additively blended on dark themes only — on Light/Retro backdrops additive saturates to invisible, so those fall back to normal blending.

**Auto-rotate became a screensaver**: enabled means the globe starts turning after 30 s without interaction, and any drag/zoom/hover stops it and restarts the clock. The preference persists.

**Fixes along the way**: the DE→DX arc and the DE/DX markers now both honour the *Show DE/DX markers* setting; arcs start exactly at the marker altitude instead of visibly below the dots; auto-rotate no longer gets silently dropped when the scene rebuilds after `lowMemoryMode` resolves.

All checks are green on the current head. Happy to split any of this out into separate PRs if you'd rather review the globe core and the satellite support independently.

KoenVdH:
Thanks for the depth here — the Pi-fleet framing was the useful part, because it turned what I'd have read as edge cases into the ones that actually matter. Pushed as `5ac5de8`. All four blockers and the seven fix-before-ship items are addressed; they landed in one commit because several share the same code paths.

**Blockers**

- **WebGL init failure (Globe3D.jsx:588)** — you were right that nothing ever threw. The catch now sets error state that rethrows during render, so the boundary genuinely fires, and the texture effect is gated on a live renderer so a dead scene no longer spends a full tile fetch before hanging on "Loading globe".
- **Cached lazy rejection + clobbered preference (WorldMap.jsx:2418)** — the lazy instance is discarded on failure so the next mount re-imports rather than replaying React's cached rejection. The fallback now switches the session without persisting; the saved projection survives, and persistence re-enables the next time the user picks a projection themselves. Your point about these two stacking was the sharpest part of the review — separately they're annoyances, together they're a device that never recovers.
- **lowMemoryMode rebuild (Globe3D.jsx:744)** — threaded `lowMem` into the marker/arc, satellite and pointer effect deps. I'd actually hit an earlier instance of this same hazard with auto-rotate being silently dropped on rebuild, and fixed it locally without generalising; your finding is what made me go back and treat the class rather than the symptom.
- **Far-side hits (Globe3D.jsx:1272)** — hover and click both intersect the sphere first and discard any point hit beyond that distance, for spots and satellites.

**Fix before ship**

- **WSJT-X (Globe3D.jsx:500)** — correct on both counts, and the band one was worse than cosmetic: an audio offset of ~1800 Hz classified as 160m, so decodes were being filtered and coloured as an entirely wrong band. Band now comes from `dialFrequency`, and the label mirrors the Leaflet `caller`/`deCall`/`dxCall` expression.
- **dxFilters parity (WorldMap.jsx:2449)** — the globe now receives `filterDXPaths(dxPaths, dxFilters)`.
- **MODIS (Globe3D.jsx:795)** — excluded from the picker in 3D; a persisted MODIS choice displays as the dark fallback the globe actually renders, without persisting the substitution, and the GIBS slider is hidden. I went with exclusion over passing the dynamic URL through, on the grounds that a half-supported style is worse than an absent one — happy to do it properly if you'd rather.
- **Style rotation (WorldMap.jsx:2466)** — paused while 3D is active. The good-neighbour point is well taken; I'd already been throttled into holey textures during development, which is exactly the failure you describe.
- **Pick radius (Globe3D.jsx:668)** — recomputed per event from world-units-per-pixel at the current camera distance, so it tracks the constant-pixel dot size.
- **Tile abort (globeTexture.js:47)** — abort listener clears `img.src` and rejects, so in-flight loads actually cancel.
- **2D-only layers (WorldMap.jsx:2492)** — a small note appears while 3D is active when enabled layers are being suppressed, with the affected layer names in its tooltip. Took your framing that visibility is enough for this PR and porting is follow-up work.

The branch has grown since you opened the review — satellite support is now rendered natively in 3D (positions, orbit tracks at altitude, footprint rings, selection shared with the Leaflet layer through the same sessionStorage key). That means the "Known limitation" paragraph in the PR description is stale. 

73, Koen ON9PK

accius (commented):
Verified 5ac5de8 claim by claim against the new head. 8 of the 11 land clean: WebGL failure now genuinely rethrows into the boundary, the lowMem dep sweep covers markers/satellites/pointer, the sphere-first occlusion cut is correct for both spots and satellites (and leaves limb satellites clickable via the horizon=Infinity path, nice touch), WSJT-X now matches the Leaflet expressions exactly, rotation pauses in 3D, the pick radius tracks screen pixels, and the tile abort actually cancels. The satellite native rendering came through the lowMem and occlusion changes with index alignment intact.

Three items need one more pass before I can approve, all small:

1. **Rotation tick bypasses the persistence block** (inline at WorldMap.jsx:652). This one matters because it defeats the exact guarantee the lazy-fallback fix just made.
2. **MODIS alias makes 'Dark' unselectable in 3D** (inline at WorldMap.jsx:2859). New regression from the aliasing approach.
3. **The layer-names tooltip is dead code under pointer-events: none** (inline at WorldMap.jsx:2555).

Ride-along nits, fine in this PR or as follow-ups, your call: terminator effect missed the lowMem dep sweep (Globe3D.jsx:859), aborted tiles still bump the next run's progress bar (globeTexture.js:117), re-selecting the current projection while persist-blocked does not trigger the save (WorldMap.jsx:2833), and the globe still draws empty-dxCall arcs the flat map drops (WorldMap.jsx:616). Your MODIS exclusion-over-plumbing call is fine by me, and the pick-radius center-distance approximation is acceptable as is.

On Michael's telemetry window question: agreed that hover identification is enough for this PR, the full telemetry dialog can ride the follow-up list with the other unported layers.

Fix the three above and this is approvable. Fast turnaround on the re-check, I promise.

K0CJH

src/components/WorldMap.jsx
```diff
@@ -612,6 +645,11 @@ export const WorldMap = ({
 
   useEffect(() => {
     if (!mapRotationConfig.enabled) return;
+    // Every style change makes the globe refetch its full tile set (up to 256
+    // tiles at retina zoom), and overlapping bursts get us throttled by the
+    // tile providers — throttled tiles are permanent holes in the texture.
+    // Rotation resumes when the user returns to a Leaflet projection.
+    if (isGlobe3D) return;
```
accius, line 658: The pause is right, but about 15 lines down the rotation tick still writes `mapProjection` straight into localStorage, bypassing projectionPersistBlockedRef. Scenario: stored projection is globe3d, the chunk fails after a redeploy, the session falls back to mercator with persistence correctly blocked, and then the first rotation tick (as soon as 5 s later, back on a Leaflet projection) writes mapProjection: 'mercator' anyway. That permanently discards the saved 3D preference, which is exactly what the block was built to prevent, under a common configuration. Cleanest fix: omit mapProjection from the rotation write entirely and let the save effect own that key.

src/components/WorldMap.jsx
```diff
@@ -2711,7 +2852,11 @@ export const WorldMap = ({
 
           {/* Style dropdown */}
           <select
-            value={mapStyle}
+            // The globe cannot build MODIS (its GIBS URL is generated
+            // dynamically by the 2D projections), so in 3D the option is
+            // hidden and a persisted MODIS choice displays as the dark
+            // fallback the globe actually renders — without persisting it.
+            value={isGlobe3D && mapStyle === 'MODIS' ? 'dark' : mapStyle}
```
accius: Aliasing the controlled select's display value to 'dark' while mapStyle state stays 'MODIS' means explicitly picking Dark fires no change event (the DOM value does not change), so the pick is silently ignored and the map snaps back to MODIS on returning to Flat. Either disable the MODIS option in 3D instead of aliasing, or track a separate display state so a real selection still lands.

src/components/WorldMap.jsx
```diff
@@ -2408,62 +2454,153 @@ export const WorldMap = ({
...
+            right: '10px',
+            zIndex: 1000,
+            background: 'var(--bg-panel)',
+            border: '1px solid var(--border-color)',
+            borderRadius: '4px',
+            padding: '3px 8px',
+            fontSize: '10px',
+            fontFamily: 'var(--font-mono)',
+            color: 'var(--text-muted)',
+            pointerEvents: 'none',
```
accius: pointer-events: none removes this div from hit testing, and browsers never show a title tooltip on an element the cursor cannot hit, so the layer names are unreachable and only the bare count is visible. Drop the pointer-events: none (the note is small enough not to block anything meaningful) or render the names inline.

src/components/Globe3D.jsx
```diff
@@ -0,0 +1,1580 @@
...
+          console.warn('[Globe3D] texture build failed:', e);
+          setTextureLoading(false);
+        }
+      });
+
+    return () => ac.abort();
+  }, [tileStyle, lowMem]);
+
+  // ── Terminator: track the subsolar point ─────────────────
+  useEffect(() => {
```
accius, line 870: Nit: this effect kept its empty dep array, so it missed the lowMem sweep. On a lowMem rebuild the new scene loses s.sunWorld and the terminator renders in the wrong place for up to 60 s until the interval fires. Add lowMem to the deps like the others.

src/utils/globeTexture.js
```diff
@@ -0,0 +1,188 @@
...
+  // the concurrency tileReproject.js settled on.
+  async function worker() {
+    while (next < queue.length) {
+      if (signal?.aborted) return;
+      const { tx, ty } = queue[next++];
+      const url = resolveTileUrl(tileUrlTemplate, tileZoom, tx, ty, lang);
+      try {
+        const img = await loadTile(url, signal);
+        if (!signal?.aborted) mctx.drawImage(img, tx * 256, ty * 256, 256, 256);
+      } catch {
```
accius, line 117: Nit: aborted workers still run the loaded++/onProgress continuation, so during a rapid style switch the old run's progress overwrites the new run's 0% and the bar visibly jumps. One-liner: return from the catch continuation when signal.aborted.

src/components/WorldMap.jsx
```diff
@@ -2689,10 +2826,14 @@ export const WorldMap = ({
             {[
               { key: 'mercator', label: 'Flat' },
               { key: 'azimuthal', label: 'Azimuthal' },
+              { key: 'globe3d', label: '3D' },
             ].map(({ key, label }) => (
               <button
                 key={key}
-                onClick={() => setMapProjection(key)}
+                onClick={() => {
```
accius, line 2843: Nit: after a fallback to mercator, clicking Flat calls setMapProjection('mercator'), React bails on the unchanged state, and the save effect never runs, so the cleared block does not persist anything until some unrelated map interaction saves. Self-heals in practice; worth persisting explicitly in this handler when clearing the block.

src/components/WorldMap.jsx
```diff
@@ -597,7 +610,27 @@ export const WorldMap = ({
   const [showMapRotationMenu, setShowMapRotationMenu] = useState(false);
   const [mapRotationMenuActivity, setMapRotationMenuActivity] = useState(0);
   const [mapProjection, setMapProjection] = useState(initialProjection);
+  // The Leaflet path applies mode/continent/watchlist filters at render time;
+  // the globe consumes paths as data, so hand it the already-filtered list or
+  // those filters silently stop working in 3D.
+  const globeDxPaths = useMemo(() => filterDXPaths(dxPaths, dxFilters), [dxPaths, dxFilters]);
```
accius: Nit: filterDXPaths covers mode/continent/watchlist, but the Leaflet path separately drops empty-dxCall entries at render time and that drop is not replicated here, so the globe draws a few arcs the flat map suppresses. Fold the empty-dxCall check into this memo for full parity.

KoenVdH:
Thanks for the quick turnaround.
All three must-fix items are in, plus all four ride-alongs. Head is now `6bcbccbc`.

**The three**

1. **Rotation tick bypassing the persistence block** (WorldMap.jsx:652) — good catch, and the scenario you spelled out is the damning part: the block survives the fallback and then a rotation tick five seconds later throws the preference away anyway. Took your suggested fix and dropped `mapProjection` from the rotation write entirely. There are now exactly two writers of that key, the save effect and the explicit toggle handler, both block-aware.

2. **MODIS alias making Dark unselectable** (WorldMap.jsx:2859) — a regression I introduced, and a nasty one because it looks like nothing happens. The option now stays present but `disabled`, labelled "(2D only)", and the controlled value is no longer aliased. Verified in the browser from the exact starting state (persisted MODIS + 3D): the select reports `MODIS`, picking Dark lands and persists as `dark`.

3. **Dead tooltip** (WorldMap.jsx:2555) — right, `pointer-events: none` meant the title could never fire. Dropped it and added `cursor: help`. The tooltip now resolves to the localised layer name.

**The four nits** — all folded in rather than deferred, since each was a line or two: the terminator effect joined the `lowMem` sweep, aborted tile workers no longer bump the next run's progress, clearing the persist block writes explicitly when the projection is unchanged, and `globeDxPaths` drops empty-`dxCall` entries for full parity with the Leaflet render path.

**Two additions since your review, and one of them touches your plugin — flagging that up front.**

You and Michael agreed the telemetry window could ride the follow-up list; it turned out to be tractable, so it is now in 3D as well. The Leaflet layer builds that window as an HTML string against a Leaflet container, which the globe has no counterpart for, so the presentation is a React component rather than a port. Rather than duplicate the logic, unit conversion, pass timing and visibility classification are extracted to `utils/satelliteTelemetry.js` and shared by both, and the i18n keys are unchanged — the same reasoning as your `ACTIVITY_COLORS` point, applied before it could bite.

That window then surfaced a real bug in my own work. Its Predict button did nothing, because it called `window.openSatellitePredict` — which `useSatelliteLayer` registers, and which never runs in 3D since WorldMap skips the Leaflet layers there. I had assumed a global was a global; it is only a global once that plugin has mounted. So `6bcbccbc` extracts the prediction modal to `utils/satellitePredict.js`, taking `satellites`, `config` and `t` as arguments instead of closing over plugin state. **`useSatelliteLayer.js` is modified as part of this** — the first time this branch touches your plugin. It is pure delegation: the plugin calls the shared function and still registers the global for existing callers, so the Flat path is byte-for-byte equivalent in behaviour. I verified both directions rather than assuming: in 3D the modal opens with real pass times while `window.openSatellitePredict` is still `undefined`, and in Flat the plugin window's Predict opens the same modal with the global registered as before. Beyond fixing the button, this removes a hidden coupling — a control in one component depending on the side effect of a plugin mounting somewhere else.

The telemetry window's presentation is still two implementations, which I would rather not leave indefinitely. Refactoring the Leaflet window onto the shared component is the obvious next step, but it means reworking drag and minimise handling in a 900-line plugin, and that felt like the wrong thing to bundle into a review round. 

**Follow-ups** — issues would still be welcome. Six are small and self-contained and I will take them promptly (colorspace include, importing the shared activity colours, `{lang}` passthrough, Null Island guard, antipodal slerp, projection token validation). The other five I would rather do properly on their own: render-on-demand instead of the unconditional 60 fps loop is the one that actually matters for the Pi fleet, and it deserves its own review rather than arriving at the end of a long branch. Same for satellite geometry reuse and short-circuiting the Leaflet layer effects while hidden.

**One thing you may not have noticed:** the workflow runs for my last few pushes are queued in `action_required`, so nothing has executed since `cc7e721a` — fork PRs need a maintainer to release each run. Locally the full set is clean on this head: `format:check`, `lang:check`, 548 tests, and `vite build`. Approving the queued runs should confirm it, and would spare you re-reviewing against an unverified head.

73, Koen ON9PK

KoenVdH:
> @KoenVdH satellite tracking you just added looking good! is it possible to include this dialog in the view, it disappears in 3D view, image

@MichaelWheeley  Should be ok now, I understand you work on the satellite aspects. Feel free to makes changes going forward as you see fit. I'm more into HF DX and POTA.
 
73, ON9PK

MichaelWheeley:
> > @KoenVdH satellite tracking you just added looking good! is it possible to include this dialog in the view, it disappears in 3D view, image
> 
> @MichaelWheeley Should be ok now, I understand you work on the satellite aspects. Feel free to makes changes going forward as you see fit. I'm more into HF DX and POTA.
> 
> 73, ON9PK

image
image
awesome - thank-you!

accius (approved):
Verified the second pass plus the new satellite work. All three must-fix items landed clean (rotation write no longer touches mapProjection, the MODIS option is disabled-with-suffix instead of aliased so real picks fire, tooltip reachable), and all four nits landed too, which I did not expect and appreciate.

The telemetry window and Predict extraction got a full review since they were new scope: the 234 lines moved out of useSatelliteLayer are behavior-identical to the old inline code (verified by diff, not by eyeball), the sessionStorage selection sync is byte-identical in both directions, duplicate panels are impossible, and the i18n keys and theme tokens all check out. Michael's screenshots confirm it works where it counts.

Follow-up list from the new code, none blocking: the Predict pass-walk runs twice per click (the first result is discarded, deleting it is free), re-opening Predict stacks a keydown listener per open, the info panel at top:50 right:10 covers the 2D-only layers note so one of them should offset, drag listeners attach even with nothing selected, satelliteTelemetry.js has no 2D callers yet so the two windows can still disagree, a hardcoded 'Ending' label, and Predict silently no-ops when omm is missing. File issues for these or fold them into your next push, either is fine.

One thing I am fixing directly on Staging right after this merge rather than holding you up: the satellite name is interpolated unescaped into the Predict modal innerHTML. It is pre-existing code you moved verbatim, but the name comes from third-party TLE data and the repo convention is esc() at every sink like this, so it gets the one-liner now that the sink is shared.

Merging. Genuinely good work across three rounds, and welcome to the contributor list.

K0CJH

KoenVdH:
Happy to be a contributor to this nice open-source project.

PR #1153, opened 2026-08-19
fix(globe): merge-review follow-ups from #1151

KoenVdH:
Six of the seven items K0CJH raised at merge. The seventh — unifying the two telemetry windows onto the shared derivation — is a refactor of the 900-line Leaflet plugin and gets its own PR, as discussed on the thread.

- Predict propagated the full 7-day pass list twice per open: the first result was rendered and then overwritten by an identical second walk before paint. One walk now, so half the SGP4 work per click.
- Reopening Predict called modal.remove() without the teardown, orphaning that invocation's document keydown handler — one listener leaked per open, each closing over a modal that no longer existed. The live modal's cleanup is now held module-level and run before a new one is built. Three opens leave one listener; closing leaves none.
- The telemetry panel (top 50, right 10) covered the 2D-only layers note (top 44, right 10). The note moves into the globe's own control column rather than the panel moving, since the panel's position deliberately mirrors the Leaflet window — and stacking in that column makes the collision structurally impossible, the same fix the DE/DX readout got.
- The panel bound pointermove/pointerup on mount, above its early return, so they stayed attached with nothing selected. They are now bound on pointerdown and released on pointerup, with an unmount guard for a drag in progress.
- 'Ending' was a hardcoded English literal; it now uses a new station.settings.satellites.passEnding key. Added to en.json only — i18n is configured fallbackLng: 'en' and lang:check verifies sorting rather than key parity, so the other 15 locales fall back cleanly until translated.
- Predict rendered as a live button when a satellite had no orbit elements and silently did nothing. It is disabled in that case, with a title saying why.

Not touched: escaping satName in the modal, which K0CJH fixed directly on Staging in 36e5c126.

Verified in-browser: three consecutive Predict opens leave a net of one keydown listener and one modal, Escape then leaves none, and the pass table still renders all 25 rows; the note and panel no longer intersect; no pointermove listener exists while the panel merely sits open, one appears during a drag, and it is released on pointerup with the panel having actually moved.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist


accius (approved):
All six verified against the diff: single pass-walk, the keydown cleanup singleton, the 2D-only note relocated into the globe's own control column where the panel cannot cover it, drag listeners bound only during an actual drag with mid-drag unmount covered, the passEnding key, and the disabled Predict button with tooltip. Also confirmed this branch carries the esc() fix from Staging. Merging.

K0CJH

PR #1154, opened 2026-08-19
perf(globe): render on demand instead of an unconditional 60 fps loop

KoenVdH:
A motionless globe redrew sixty times a second to produce identical pixels. On the Pi-class hardware this project targets that is continuous CPU and heat for no visible change, which is why K0CJH singled it out.

The loop now parks itself when nothing is happening and restarts on invalidation. controls.update() reports whether it moved the camera, which covers the damping tail after a drag without special-casing it; everything else that mutates the scene calls requestRender() — texture swaps, overlay and satellite rebuilds, the night-darkness uniform, backdrop visibility on theme change, the terminator's 60 s sun update, programmatic camera moves, and resize. OrbitControls' own 'change' event covers drag and wheel zoom.

Auto-rotate is explicitly exempt from parking. update() reports movement below its epsilon as "no change", and parking on such a frame would have been terminal: the event that would wake the loop comes from the very update that no longer runs, so the globe would have stopped mid-spin and stayed stopped. Caught by measuring 6 rendered frames out of 10 driven before the guard, 20 of 20 after.

Measured by stepping the loop directly, since requestAnimationFrame is throttled in a background tab: idle draws 1 frame per 50 steps (the last damping tick) and then nothing, invalidation draws exactly one, a theme change repaints, and auto-rotate never parks.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist

### DIFF
--- a/src/components/Globe3D.jsx
+++ b/src/components/Globe3D.jsx
@@ -352,7 +352,10 @@ export default function Globe3D({
     if (!autoRotateEnabledRef.current) return;
     idleTimerRef.current = setTimeout(() => {
       const s2 = gl.current;
-      if (s2.controls && autoRotateEnabledRef.current) s2.controls.autoRotate = true;
+      if (s2.controls && autoRotateEnabledRef.current) {
+        s2.controls.autoRotate = true;
+        s2.requestRender?.(); // loop is parked by now; nudge it back to life
+      }
     }, AUTOROTATE_IDLE_MS);
   }, []);
   const [textureLoading, setTextureLoading] = useState(true);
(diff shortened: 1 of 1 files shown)


accius (approved):
Ran a full invalidation-path audit against the head file: first paint, drag and damping tail via the controls change listener, auto-rotate never parks and both enable paths wake a parked loop, texture swap, night darkness, backdrop and theme, marker and satellite rebuilds, the terminator sun tick, centering, and resize all invalidate correctly, and hover is DOM-only so it correctly does not. No missed-wakeup or deadlock path found. Best-executed of the three follow-ups. Two cosmetic nits not worth a round trip: a duplicated comment line at the markers-effect tail, and the custom change listener is not removed before controls.dispose(), which is harmless. Merging.

K0CJH

PR #1155, opened 2026-08-19
refactor(satellites): render both telemetry windows from one derivation

KoenVdH:
## What does this PR do?

The last of the seven follow-ups @accius raised when merging #1151: *"satelliteTelemetry.js has no 2D callers yet so the two windows can still disagree."*

`satelliteTelemetry.js` was extracted during #1151 but only the new 3D panel called it — the Leaflet window still computed its own figures. The two agreed only because I had mirrored the maths by hand, and nothing enforced that. Change the imperial conversion or the next-pass rule in one place and the same satellite would report a different altitude in Flat than in 3D. Exactly the `ACTIVITY_COLORS` hazard you called out, which is presumably why you spotted it.

**The Leaflet window now calls `deriveSatelliteTelemetry` for its numbers.** Its HTML generation is untouched. The defect was divergent *figures*, not divergent markup, so this closes it without going near the window's drag and minimise handling. Net 48 lines lighter: `safeNum`, `safeArr` and the local `formatSecsFromNow` are dead and gone. `safeStr` stays — it still escapes interpolated values.

**Also adds unit tests for the shared module, which had none.** That is the part that actually stops the drift recurring; the refactor alone only makes drift preventable. 13 tests covering the unit conversions, the pass-selection rule (first pass not yet finished, not merely first listed), the three visibility states, and missing-data behaviour.

### On step two

The obvious next move is rendering the shared React component inside the Leaflet container so the presentation is single-sourced too. **I have deliberately not done that, and my recommendation is not to.** It means standing up a React root inside a 900-line plugin while preserving its drag and minimise behaviour, and the payoff is deduplicating markup rather than fixing a defect. Two windows drawing the same verified numbers is a normal state for a codebase. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. `npm ci`, then `node server.js` and `npm run dev`.
2. Select two satellites in **Flat** and open the telemetry window. Every row — latitude, longitude, altitude, speed, Az/El, status, next pass, mode, downlink, uplink, tone — should read exactly as it did before this change.
3. Switch to **3D** with the same satellites selected. Both windows should agree on labels, units and formatting; only the genuinely time-varying values (position, countdowns) differ, and only by the seconds elapsed between looking at them.
4. Change **Settings → Units** between metric and imperial and confirm both windows convert together.
5. `npx vitest run src/utils/satelliteTelemetry.test.js`.

## Checklist

## Notes on verification

Equivalence was established **before** the swap rather than assumed afterwards. I ran the plugin's original formulas against the shared module over the live satellite feed plus synthetic edge cases the feed will not produce (no fields at all, visible, above-horizon-but-below-minimum, below-horizon):

**860 field comparisons across 39 live satellites and 4 synthetic ones, in both unit systems — identical throughout.**

Then confirmed in the browser, reading both windows for the same satellites: same labels, same units, same formatting, differing only where the satellite had actually moved between the two readings.

One note on the tests. `altitude` and `speed` go through `toLocaleString`, so my first draft hardcoded US thousands separators and would have failed for any contributor in a European locale — my own UI renders `15.932` where CI renders `15,933`. The expectations now build through `toLocaleString` as well, so they pin the conversion and the unit without pinning the separator. Verified passing under both `en_US` and `nl_NL`.

### DIFF
--- a/src/plugins/layers/useSatelliteLayer.js
+++ b/src/plugins/layers/useSatelliteLayer.js
@@ -2,6 +2,7 @@ import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { addMinimizeToggle } from './addMinimizeToggle.js';
 import { replicatePoint, replicatePath } from '../../utils/geo.js';
+import { deriveSatelliteTelemetry } from '../../utils/satelliteTelemetry.js';
 import { openSatellitePredict as openSatellitePredictShared } from '../../utils/satellitePredict.js';
 
 export const metadata = {
@@ -52,14 +53,6 @@ export const useLayer = ({ map, enabled, satellites, setSatellites, opacity, con
   };
 
   // Helper to format seconds from now into a string representation e.g. "00:12:34"
-  const formatSecsFromNow = (secsFromNow) => {
-    return secsFromNow > 3600
-      ? `${String(Math.floor(secsFromNow / 3600)).padStart(2, '0')}:${String(Math.floor((secsFromNow % 3600) / 60)).padStart(2, '0')}:${String(secsFromNow % 60).padStart(2, '0')}`
-      : secsFromNow > 60
-        ? `00:${String(Math.floor(secsFromNow / 60)).padStart(2, '0')}:${String(secsFromNow % 60).padStart(2, '0')}`
-        : `00:00:${String(secsFromNow).padStart(2, '0')}`;
-  };
-
   const fetchSatellites = async () => {
     try {
       const response = await fetch('/api/satellites/data');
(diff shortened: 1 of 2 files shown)


MichaelWheeley (changes requested):
@KoenVdH 
PR needs to target accius:Staging and not accius:main

https://github.com/community/community/assets/1710840/21bb0aec-26bc-4aed-af81-1ad7de652753

MichaelWheeley:
- updates `useSatelliteLayer.js` so that data formatting source is common to 3D, works as described for metric/USCustomary/Flat/3D
- safeties removed where no longer needed
- adds unit tests, nice
- good catch on toLocalString

PR #1156, opened 2026-08-19
fix(maps): share the activity palette, fix satellite colour space and…

KoenVdH:
… 3D labels

Three of the deferred review items, all user-visible.

**Activity colours.** The globe declared its own POTA/WWFF/SOTA/WWBOTA values with a comment claiming they matched the other projections. Only POTA did. WWFF, SOTA and WWBOTA were shades I had invented, so the same spot appeared in one colour on the globe and another in the panels, the layer picker and the azimuthal map. The four programme colours now come from utils/activityColors.js, taken from the values POTAPanel, WWFFPanel, SOTAPanel, WWBOTAPanel, PotaSotaPanel and the layer picker already agree on. The globe keeps its own constants for PSK, WSJT-X and the band fallback, which no other view draws.

Adopting the shared source also fixes an older inconsistency in the azimuthal map, which drew WWBOTA in #4488ff — the same blue as the DE station marker, so a WWBOTA spot could be mistaken for your own station.

**Satellite dot colour space.** That shader was copied from the spot-dot shader without its `#include <colorspace_fragment>`, so satellite dots were written as linear values into an sRGB target and drifted from the palette. Same fix as the other three shaders already carry.

**Basemap language in 3D.** The globe never passed `{lang}` to the texture builder, so it fell through to the builder's 'en' default and non-English operators got English labels in 3D while Flat and Azimuthal honoured their language. It now resolves the language the same way AzimuthalMap does, and the texture rebuilds when the language changes.

Verified the three invented colours appear nowhere in src or in the built bundle, and that both maps now import the shared palette.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1.
2.
3.

## Checklist

### DIFF
--- a/src/components/AzimuthalMap.jsx
+++ b/src/components/AzimuthalMap.jsx
@@ -15,6 +15,7 @@ import { MAP_STYLES } from '../utils/config.js';
 import { createTileReprojector } from '../utils/tileReproject.js';
 import { createAzimuthalCRS } from '../utils/azimuthalCRS.js';
 import { matchesDXSpotPath } from '../utils/dxClusterSpotMatcher';
+import { ACTIVITY_COLORS } from '../utils/activityColors.js';
 
 // ── Projection Math ────────────────────────────────────────
 const DEG = Math.PI / 180;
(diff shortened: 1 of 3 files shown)


accius (approved):
Palette values verified against all six existing consumers by grep, the colorspace include now matches the spot shader, and {lang} rides the texture deps so a language switch rebuilds the globe. The WWBOTA colour change on the azimuthal map is user-visible and correct (it was colliding with the DE-marker blue), noted for the changelog. One follow-up worth an issue rather than a blocker: the four activity panels and PotaSotaPanel still hardcode their matching copies, so the new util only enforces the two map projections. Merging.

K0CJH

PR #1160, opened 2026-08-21
feat(rig-bridge): Yaesu band-select on band change, band-aware SSB

KoenVdH:
## What does this PR do?

Yaesu CAT only ever received `FA;` when tuning, which moves the VFO without telling the radio it has changed band. Band-specific settings — gain, ATU memory, antenna selection — stay on whichever band was previously selected, so clicking a spot on a new band leaves the radio set up for the old one.

Rig Bridge now sends the `BS` band-select command first, waits 100 ms for the radio to settle, then sends `FA;`. Frequencies above 450 MHz have no `BS` code and are sent exactly as before.

Originally reported and prototyped in discussion #1032; this goes further than that draft in three ways.

### `BS;` only on an actual band change

`BS;` makes the radio recall the band's stored VFO, so issuing it for an in-band retune would bounce the frequency to the stored value and back, taking the ATU and band relays with it. It is therefore sent only when the target is on a different band than the radio is on now.

The comparison uses the radio's **live frequency from the `IF;` poll**, not a remembered "band we last selected". A remembered value goes stale the moment the operator turns the band knob — the rig would be on 40m, the memo would still say 20m, and the next click on a 40m spot would skip the `BS;` it actually needed. An unknown current frequency counts as a band change, so the first tune after connect always sets the band.

### Mode is applied after the band change, not before

A band change also restores the band's stored **mode**, which silently overwrites a `MD;` issued before the switch completed.

A mode requested alongside the tune is therefore sent after the band change. It is resolved when the deferred write fires rather than when `setFreq` is called: the caller sends frequency and mode as two commands milliseconds apart, so at fire time the mode just requested wins instead of the one that was in effect when the tune started. Capturing it up front would have re-asserted the old mode 100 ms later and undone the mode the caller had just set — breaking the case that already worked.

When no mode was requested, nothing is sent and the radio keeps the mode it recalled for the new band.

`MD;` precedes `FA;` inside that deferred write because switching to CW shifts the displayed frequency by the pitch offset — observed live as a 700 Hz error — so the frequency has to be set last.

### Generic `SSB` now resolves to the correct sideband

**This part is an existing bug, not a consequence of the band-select work.** `MODE_ALIASES` mapped `SSB` unconditionally to `'2'` (USB), so `POST /mode {"mode":"SSB"}` gave USB on 40m and 80m. OpenHamClock itself is unaffected — `mapModeToRig()` resolves the sideband before sending — but any other client using the generic alias got the wrong sideband below 10 MHz.

`SSB` now resolves from the frequency: LSB below 10 MHz, USB at and above, with 60m as the USB exception. The bounds (5.300–5.405 MHz) are the ones `src/utils/bandPlan.js` `getSideband()` already uses, so the bridge and the app cannot disagree about what `SSB` means; that span covers both the IARU R1 band and the US channels. An unknown frequency still falls back to USB, as before. The `SSB` entry was removed from `MODE_ALIASES` rather than left beside the new branch, so there is one source of truth for the sideband.

`protocol-kenwood.js` and `protocol-icom.js` carry the same band-agnostic `SSB: '2'`. Left alone deliberately — happy to fix them in a follow-up, but I can only test against a Yaesu.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

Needs a Yaesu on USB CAT. With Rig Bridge running and connected:

1. **Band change sets the band.** From 20m, click a spot on 40m. The radio should switch band — check that a band-specific setting you have configured differently per band (RF gain, ATU memory, antenna) follows, rather than staying on the 20m setting.
2. **In-band retune stays quiet.** Click another spot on the same band. The frequency should move directly, with no jump to the band's stored frequency and back, and no relay/ATU activity.
3. **Mode survives a band change.** Click a CW spot on a different band. The radio must end in CW, not in whatever mode that band had stored.
4. **Sideband.** `curl -X POST -H 'Content-Type: application/json' -H 'X-RigBridge-Token: <token>' -d '{"mode":"SSB"}' http://localhost:5555/mode` on 40m gives LSB, on 20m gives USB, and on 60m (5.3515) gives USB.
5. **Above 450 MHz.** Tune 1296 MHz if your radio covers it: no `BS;`, plain `FA;` as before.

Run with `--debug` to watch the CAT frames directly.

## Checklist

- [ ] Tested in **Dark**, **Light**, and **Retro** themes — *n/a, no UI change*
- [ ] Responsive at different screen sizes — *n/a, no UI change*

## Verification

CAT output, verified against the module directly:

```
band change + SSB to 40m         BS03; MD01; FA007150000;
band change + SSB to 60m         BS02; MD02; FA005351500;
band change + CW                 BS03; MD03; FA007020000;
band change, no mode requested   BS03; FA007150000;
in-band retune                   FA014246000;
above 450 MHz                    FA1296100000;
```

Sideband boundaries, exact rather than rounded: 5.299999 → LSB, 5.300000 → USB, 5.405000 → USB, 5.405001 → LSB, 9.999999 → LSB, 10.000000 → USB. Explicit `USB`/`LSB`/`CW`/`FT8`/`DATA-USB` are unchanged, and lowercase `ssb` resolves too.

Confirmed on a live radio (FT-series, 38400 baud):

- 40m → 20m with no mode command ended in **CW**, where the band's stored mode was USB — proof the `MD;` landed after the band recall.
- 20m → 40m with a mode command right behind it ended in the **requested** mode, not the pre-change CW — the case a naive implementation breaks.
- Generic `SSB` on 40m produced **LSB** on the radio, where it previously produced USB.

The deferred write is safe across a disconnect: `write()` in `plugins/usb/index.js` returns false when the port is closed, so a late `FA;` is a no-op rather than a throw.

None — no UI change.

### DIFF
--- a/rig-bridge/plugins/usb/index.js
+++ b/rig-bridge/plugins/usb/index.js
@@ -327,42 +327,64 @@ function createUsbPlugin(radioType) {
         console.log(`[USB/${radioType}] Disconnected`);
       }
 
-      // A band change makes the radio recall that band's stored mode, which would
-      // overwrite a mode command issued before the band finished switching. The
-      // mode is therefore resolved when the deferred write fires: a mode
-      // requested alongside the tune is applied after the band change, and when
-      // none was requested the radio keeps the mode it recalled for that band.
+      // A band change makes the radio recall that band's stored mode and VFO, so
+      // both the mode to apply afterwards and the band to compare against have
+      // to survive the gap between a tune being requested and the radio
+      // reporting it back on the next IF; poll.
       let lastRequestedMode = null;
-      let modeRequestSeq = 0;
+      let lastRequestedModeAt = 0;
       let lastRequestedFreq = 0;
       let lastRequestedFreqAt = 0;
 
-      // Frequency a band-dependent mode (generic SSB) resolves against. A tune
-      // still in flight has not been echoed by the radio yet, so the frequency
-      // just requested wins; after that window state.freq is authoritative and
-      // also reflects the operator turning the VFO by hand.
-      const FREQ_IN_FLIGHT_MS = 1000;
+      // A tune and its mode arrive as two separate commands milliseconds apart,
+      // and nothing guarantees which lands first. Anything inside this window
+      // belongs to the same tune, whichever order it arrived in.
+      const TUNE_WINDOW_MS = 1000;
+
+      /**
+       * Frequency to reason about the radio's current band and sideband.
+       *
+       * state.freq only refreshes on the IF; poll (500 ms by default), so a tune
+       * still in flight has not been echoed back yet. Within the window the
+       * frequency just requested wins; after it, state.freq is authoritative and
+       * also reflects the operator turning the VFO by hand.
+       */
       const modeReferenceFreq = () =>
-        Date.now() - lastRequestedFreqAt < FREQ_IN_FLIGHT_MS ? lastRequestedFreq : state.freq;
+        Date.now() - lastRequestedFreqAt < TUNE_WINDOW_MS ? lastRequestedFreq : state.freq;
 
       function setFreq(hz) {
         console.log(`[USB/${radioType}] SET FREQ: ${(hz / 1e6).toFixed(6)} MHz`);
+        // Both read before this tune is recorded: comparing the target against
+        // itself would make every band change look like an in-band retune, and
+        // the mode has to be the one in effect before BS; recalls another.
+        const referenceFreq = modeReferenceFreq();
+        const modeBefore = state.mode;
         lastRequestedFreq = hz;
         lastRequestedFreqAt = Date.now();
+
         if (radioType === 'icom') {
           proto.setFreq(hz, write, getIcomAddress());
         } else {
-          // Third arg is the radio's live frequency: Yaesu uses it to decide
-          // whether this is a band change. Kenwood ignores both extra args.
-          const seqAtCall = modeRequestSeq;
-          proto.setFreq(hz, write, state.freq, () => (modeRequestSeq !== seqAtCall ? lastRequestedMode : null));
+          // Third arg is the frequency Yaesu compares to decide whether this is a
+          // band change. Kenwood ignores both extra args.
+          proto.setFreq(hz, write, referenceFreq, () => {
+            // A mode requested around this tune wins. Otherwise re-assert the
+            // mode from before the band change: the caller skips the mode command
+            // when the mode is already correct, so letting the band's stored mode
+            // win would drop the operator out of it — clicking a 20m CW spot from
+            // 40m CW would land in USB.
+            if (lastRequestedMode && Date.now() - lastRequestedModeAt < TUNE_WINDOW_MS) {
+              return lastRequestedMode;
+            }
+            return modeBefore;
+          });
         }
       }
 
       function setMode(mode) {
         console.log(`[USB/${radioType}] SET MODE: ${mode}`);
         lastRequestedMode = mode;
-        modeRequestSeq++;
+        lastRequestedModeAt = Date.now();
         if (radioType === 'icom') {
           proto.setMode(mode, write, getIcomAddress());
         } else {
--- a/rig-bridge/plugins/usb/index.js
+++ b/rig-bridge/plugins/usb/index.js
@@ -327,21 +327,46 @@ function createUsbPlugin(radioType) {
         console.log(`[USB/${radioType}] Disconnected`);
       }
 
+      // A band change makes the radio recall that band's stored mode, which would
+      // overwrite a mode command issued before the band finished switching. The
+      // mode is therefore resolved when the deferred write fires: a mode
+      // requested alongside the tune is applied after the band change, and when
+      // none was requested the radio keeps the mode it recalled for that band.
+      let lastRequestedMode = null;
+      let modeRequestSeq = 0;
+      let lastRequestedFreq = 0;
+      let lastRequestedFreqAt = 0;
+
+      // Frequency a band-dependent mode (generic SSB) resolves against. A tune
+      // still in flight has not been echoed by the radio yet, so the frequency
+      // just requested wins; after that window state.freq is authoritative and
+      // also reflects the operator turning the VFO by hand.
+      const FREQ_IN_FLIGHT_MS = 1000;
+      const modeReferenceFreq = () =>
+        Date.now() - lastRequestedFreqAt < FREQ_IN_FLIGHT_MS ? lastRequestedFreq : state.freq;
+
       function setFreq(hz) {
         console.log(`[USB/${radioType}] SET FREQ: ${(hz / 1e6).toFixed(6)} MHz`);
+        lastRequestedFreq = hz;
+        lastRequestedFreqAt = Date.now();
         if (radioType === 'icom') {
           proto.setFreq(hz, write, getIcomAddress());
         } else {
-          proto.setFreq(hz, write);
+          // Third arg is the radio's live frequency: Yaesu uses it to decide
+          // whether this is a band change. Kenwood ignores both extra args.
+          const seqAtCall = modeRequestSeq;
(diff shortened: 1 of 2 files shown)


accius (commented):
Really strong PR — thank you for the write-up, it made this fast to review. The reasoning all holds up against the code: why `BS;` has to be conditional, why the mode is resolved when the deferred write fires rather than at `setFreq` time, and why `MD;` has to precede `FA;` because of the CW pitch offset. I also confirmed the two claims that would have worried me otherwise: `write()` in `plugins/usb/index.js` does return `false` on a closed port, so a late `FA;` is a genuine no-op, and the `BS` code table matches the documented Yaesu mapping. The `SSB` sideband fix is a real standalone bug and I'm glad you pulled it out.

One thing needs fixing before this can go in, plus a few smaller notes.

---

## Blocker — mode is lost on a band change when the target mode equals the current mode

`src/contexts/RigContext.jsx:536` only calls `setMode` when the mode actually changes:

```js
if (targetMode && targetMode !== rigState.mode) {
  setMode(targetMode);
}
```

`mapModeToRig()` returns exactly the strings the `IF;` parser writes into `rigState.mode` (`'CW'`, `'DATA-USB'`, `'USB'`, …), so that comparison really does match and really does skip the call. When it skips, `modeRequestSeq` never increments, the deferred `getMode()` returns `null`, and the radio keeps whatever mode `BS;` recalled for the new band.

Concrete failure: radio on 40m CW, operator clicks a 20m CW spot.

- `targetMode === 'CW'`, `rigState.mode === 'CW'` → `setMode` is not called
- bridge sends `BS05;` then `FA014025000;`
- 20m recalls its stored mode — USB for most operators — and the radio ends up in **USB on a CW spot**

Before this PR no `BS;` was sent at all, so the radio stayed in CW. The same shape hits 20m → 15m FT8: both resolve to `DATA-USB`, so the guard skips, the band recall wins, and the operator drops out of data mode. That's a common click pattern, which is what makes this a blocker rather than an edge case.

Two ways out, either works:

1. **Bridge-local.** When no mode was requested, re-assert `state.mode` — the mode the radio was in *before* the band change — instead of letting the band recall win. This contradicts the design note in your description ("the radio keeps the mode it recalled for that band"), so it's your call, but it restores the pre-PR behaviour and it also covers third-party clients that only `POST /freq`.
2. **App-side.** Drop the `targetMode !== rigState.mode` guard in `tuneTo` when `autoMode` is on. Cheaper, but leaves `/freq`-only clients exposed.

I'd lean toward (1) since it keeps the fix inside the bridge where the band change actually happens, but I don't have a radio to test against — your judgement.

---

## Notes — not blocking

**1. The band comparison reads a stale `state.freq`.**

`setFreq` passes `state.freq`, which only refreshes on the `IF;` poll (default 500 ms, `plugins/usb/index.js:110`). Two clicks on the same new band inside that window will re-issue `BS;`, bouncing the VFO and taking the ATU and band relays with it — precisely what the conditional exists to prevent.

You already track `lastRequestedFreq` / `lastRequestedFreqAt` for mode resolution. Feeding `modeReferenceFreq()` into the band comparison too would close this and make the two paths consistent, rather than having in-flight tracking that only one of them benefits from.

**2. `setMode`-before-`setFreq` ordering.**

The sequence-counter trick assumes the caller sends frequency first, then mode. OpenHamClock does exactly that, so this is fine today. But a client that posts `/mode` then `/freq` captures an already-incremented `modeRequestSeq` in `seqAtCall`, so `getMode()` returns `null` and its mode is silently dropped on a band change. Worth at least a comment noting the assumption.

**3. Out-of-band frequencies snap to the nearest ham band code.**

`freqToBandCode()` has no gaps, so 15 MHz WWV resolves to `BS06` (17m) and 12 MHz to `BS05` (20m). `FA;` still lands afterwards, so general-coverage RX probably behaves — but there's no `11` (GEN) escape hatch for frequencies outside any ham band. Low confidence this matters in practice; you have the radio, so I'd rather ask than guess.

**4. Tests.**

None added, but `rig-bridge/` has no test suite at all, so that's consistent with the repo rather than a gap in this PR. Not asking you to build one here.

---

Leaving `protocol-kenwood.js` and `protocol-icom.js` alone was the right call — happy to take the `SSB` sideband fix for those as a follow-up whenever someone with the hardware can confirm it.

Sort the blocker and this is good to merge.

KoenVdH:
Thanks — the blocker is real and you're right about the mechanism. Fixed, along with all three notes.

## Blocker — mode lost on a same-mode band change

Confirmed before changing anything: `mapModeToRig()` returns exactly the strings the `IF;` parser writes (`'CW'`, `'DATA-USB'`, `'USB'`, …), so the guard genuinely matches and genuinely skips, and the deferred write then sent no `MD;` at all.

Took your option 1. When no mode is requested, the mode from **before** the band change is re-asserted.

Worth saying plainly: my PR description argued the opposite ("the radio keeps the mode it recalled for that band"), and that was wrong. Re-asserting isn't overriding the caller — `tuneTo()` only stays silent when the mode is *already correct*, so re-asserting keeps the mode the app has already decided on. The case I'd worried about (carrying USB from 20m to 40m) can't arise through OpenHamClock: on a 40m SSB spot `targetMode` is `LSB` and the current mode is `USB`, they differ, so a mode **is** sent and the requested mode wins. And as you note, it covers `/freq`-only clients too.

Verified on the radio, which is the case you described:

```
40m CW → 20m CW spot, no mode command
→ BS05; MD03; FA014030000;      radio ends in CW
```

## Note 1 — stale `state.freq` in the band comparison

Fixed; the band comparison now uses the same in-flight frequency the sideband resolution already used. One wrinkle worth flagging for review: it has to be read *before* the tune is recorded, or the target compares against itself and `BS;` never fires at all.

```
two clicks on a newly selected band, 300 ms apart
→ BS07; MD03; FA021074000; FA021200000;      one BS;
```

## Note 2 — `/mode` before `/freq`

Fixed rather than commented. The sequence counter is replaced by a recency window on the requested mode, which is order-independent and drops a variable:

```
/mode CW then /freq (band change)
→ MD03; BS05; MD03; FA014030000;      radio ends in CW
```

The duplicate `MD;` there is deliberate: the first may be swallowed while the band is still switching, the second lands after `BS;` has settled.

## Note 3 — out-of-band frequencies

Implemented; you were right to be suspicious. `BAND_MAP` now carries lower edges as well, and a frequency outside every ham band selects no band and is sent as a plain `FA;` — the path frequencies above 70cm already took. Edges span the widest regional allocation per band (7.0–7.3 covers both R1 and the US), so no legitimate in-band frequency is refused a band change; a frequency legal in one region but not another still tunes, it just keeps the current band's settings.

```
15 MHz WWV   was BS06 (17m)   → FA only
12 MHz       was BS05 (20m)   → FA only
 5 MHz       was BS02 (60m)   → FA only
0.9 MHz MW   was BS00 (160m)  → FA only
70.2 MHz 4m  no code          → FA only
```

All thirteen in-band cases still select the same code as before.

## Note 4 — tests

Left as is, per your note.

accius (approved):
Verified 26f8757e against the review point by point. The blocker fix is right at every edge I pressed: modeBefore is captured synchronously before the tune is recorded, the sub-second window prefers a just-requested mode over a possibly stale echo, an unknown mode at startup skips the MD; write instead of sending garbage, and no third write appears when a mode was requested. The band comparison reading modeReferenceFreq() before recording the new tune avoids the self-comparison trap, the recency window handles mode-before-freq ordering both ways, and the BAND_MAP min/max ranges route every gap frequency I threw at them (WWV at 10.0 MHz included) to plain FA;. Your on-radio traces match the code paths.

One code-level follow-up worth filing, not blocking: the 100 ms deferred write still cannot be cancelled, so two rapid cross-band clicks can interleave as BS_A BS_B FA_A FA_B with a ~50 ms bounce to A's frequency before B's corrects. A clearTimeout of the previous deferred write at the top of setFreq closes it. The band-edge table living separately from lib/wsjtx-enrich's getBandFromHz is a second small follow-up.

Merging. The writeup quality on this PR and the follow-through made it easy.

K0CJH

PR #1161, opened 2026-08-25
feat(globe): real polar imagery for the 3D globe's satellite basemap

KoenVdH:
## What does this PR do?

The 3D globe's polar caps had no imagery. Web Mercator stops at ±85.0511° — latitude runs to infinity at the poles, so the tile grid cannot contain them — and the equirectangular remap filled the gap by repeating the last real row. `SphereGeometry` collapses every column of a row onto the pole, so those 56 identical rows rendered as thousands of wedges radiating from the point.

For the **Satellite** basemap the data does exist, just in a different projection. Esri publishes `Polar/Arctic_Imagery` and `Polar/Antarctic_Imagery` alongside `World_Imagery`, and the ArcGIS `export` endpoint will reproject on request — asking for `imageSR=4326` returns the cap already in the projection this texture uses. No reprojection maths, no second tile grid to walk: one request per pole, drawn straight into the cap rows.

**Only Satellite is affected.** Carto, Google and the other Esri basemaps publish Mercator only, so they are left exactly as they were — the polar code is behind a check for a configured polar source, and a basemap without one takes the same path it always has.

### Details worth knowing

**Coverage.** Real imagery reaches 89.47°N; the services cut a clean circular hole above that — 100% covered below, 0% above, no ragged edge. That half-degree hole is blurred and faded to the average of the last real row, which is also the fallback if a fetch fails.

**Colour.** The two services render the same ocean differently: Arctic imagery came in at mean luminance 30.8 where World Imagery is 67.9 at the same latitude, and dropping it in raw left a bright ring at 85°. Each cap is scaled about its own mean to match the basemap's level *and* spread — matching the mean alone still left the cap interior visibly wrong — measured over a band of rows so one stray row of cloud cannot skew it. The strip's alpha then ramps in over the first 15% of the cap, because on a sphere any residual difference lands on a single circle of latitude and reads as a ring.

**Edges.** The exported strips are transparent for ~11px either side of ±180. That gap is closed by interpolating between the last valid column on each side — they are neighbours once the texture wraps.

**Failure is silent.** Any fetch error, CORS problem or missing service returns null and the cosmetic fallback covers the whole cap. Real polar imagery is an improvement on the fallback, never a requirement for building a texture.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] Translation
- [ ] Map layer plugin

## How to test

1. `npm ci`, then `node server.js` and `npm run dev`. Switch the map to **3D** and the style to **Satellite**.
2. Drag down over the North Pole. The wedge pinwheel should be gone, replaced by Arctic sea ice and bathymetry that continues across 85° rather than starting there.
3. Same at the South Pole.
4. Switch through **Dark**, **Terrain**, **Countries**, **Hybrid**. Each should build and render exactly as before — none of them touch the polar code.
5. Block `services.arcgisonline.com/arcgis/rest/services/Polar/*` in devtools and reload Satellite: the globe should still build, with the caps falling back to the faded treatment.

## Checklist

## Verification

Measured on the built texture (4096×2048, 56 cap rows) rather than judged by eye. Per-row luminance across the north cap; the basemap below the boundary sits at 68:

```
           mean before   mean after
85.21°         68            69.7
86.79°         68            69.7
88.11°         68            66.9
89.43°         68            64.5
```

Before, every row from the boundary to the pole was byte-identical — one repeated row of coastline. After, each row carries its own imagery, and the tone tracks the basemap the whole way in.

South cap, where the basemap below reads 240.6:

```
-85.21°  240.0     -86.79°  239.8     -88.55°  241.2     -89.78°  242.2
```

Cap edge columns read ordinary ocean and ice — `(23,74,90)` north, `(238,240,252)` south — with no seam at the antimeridian.

## Notes

MODIS could get the same treatment from NASA GIBS, which publishes both EPSG:4326 and polar stereographic. Left out deliberately: its true-colour product has no data at whichever pole is in polar night — the August Antarctic tile is a 5KB black image — so it needs a seasonal story of its own.

Before:
image
After:
image

### DIFF
--- a/src/components/Globe3D.jsx
+++ b/src/components/Globe3D.jsx
@@ -879,6 +879,8 @@ export default function Globe3D({
       // Countries ships transparent overlay tiles; flat mode paints this same
       // blue behind them via the map div's background.
       baseColor: MAP_STYLES[style].countriesOverlay ? '#4a90d9' : undefined,
+      // Real imagery for the polar caps, where this basemap has a polar source.
+      polar: MAP_STYLES[style].polar,
       onProgress: (p) => setTextureProgress(p),
       signal: ac.signal,
     })
--- a/src/utils/config.js
+++ b/src/utils/config.js
@@ -358,6 +358,14 @@ export const MAP_STYLES = {
     name: 'Satellite',
     url: 'https://services.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
     attribution: '&copy; Esri',
+    // Web Mercator stops at ±85.05°, so the globe's polar caps have no imagery
+    // from the tiles above. These polar-projected services cover the rest, and
+    // ArcGIS reprojects them to plate carrée server-side on request — see
+    // fetchPolarStrip() in utils/globeTexture.js.
+    polar: {
+      north: 'https://services.arcgisonline.com/arcgis/rest/services/Polar/Arctic_Imagery/MapServer',
+      south: 'https://services.arcgisonline.com/arcgis/rest/services/Polar/Antarctic_Imagery/MapServer',
+    },
   },
   MODIS: {
     name: 'Modis Truecolor', // NASA GIBS MODIS Truecolor Imagery
--- a/src/utils/globeTexture.js
+++ b/src/utils/globeTexture.js
@@ -138,13 +138,74 @@ export async function buildGlobeTexture({ tileUrlTemplate, tileZoom = 3, lang, b
 
   for (let y = 0; y < outH; y++) {
     const lat = 90 - ((y + 0.5) / outH) * 180;
-    // Beyond ±85.05° Mercator has no data; clamping stretches the last real
-    // row over the caps, which is the usual compromise for globe textures.
+    // Beyond ±85.05° Mercator has no data, so clamping repeats the last real
+    // row across the cap. resolvePolarCaps() below then fades those repeats out
+    // — on a sphere every column of a row converges to the pole, so a repeated
+    // row of coastline becomes a pinwheel of wedges.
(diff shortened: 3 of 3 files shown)


accius (approved):
Reviewed in depth against the head. The parts that had to be right, are: I verified the settlePolarCap index math for both poles including the failed-fetch case, the strip orientation and blend-ramp directions, the antimeridian interpolation, and the abort-guarded application in Globe3D — and the non-polar styles are provably untouched (the polar work is fully behind the if (polar) guard). Canvas taint is unreachable (CORS-mode fetch + createImageBitmap cannot taint, tiles load with crossOrigin anonymous). The exclusion-over-generality call for non-Satellite basemaps is the right one. Merging.

Follow-ups worth a pass when you're next in this file, none blocking:

- fetchPolarStrip's bare catch swallows AbortError, so an abort mid-polar-fetch still runs both settlePolarCap fallbacks and resolves with a doomed texture instead of rejecting 'aborted' like the tile stage does. Today only Globe3D's own aborted-check prevents a stale texture being applied; a signal re-check after the await restores the contract.
- onProgress hits 1.0 when tiles land, then the build awaits two dynamic ArcGIS exports with no further reporting — on a slow export the overlay reads as hung at 100%. Either report a polar phase or apply the caps after first paint.
- The equirect context should be created with willReadFrequently: true now that the polar path does ~6 getImageData readbacks on it (the scratch contexts already do this).
- Smaller: the polar config is ArcGIS-export-shaped behind a generic-looking key (worth a type discriminator or URL template before a second provider shows up), buildGlobeTexture's JSDoc @returns predates the {canvas, meanLuma} shape, and the smoothstep/loadTile duplications could fold into the existing helpers.

Good work getting the poles to stop being wedge soup.

K0CJH

PR #1172, opened 2026-09-02
Hotfix: stop paging watchtower on single fletcher relay blips

accius:
Post-26.7.0 ops fix. Watchtower has been paging ok→degraded→ok pairs on prod: fletcher is a passive relay with no background refresh, so a single transient upstream timeout left lastUpstreamErrorAt newer than lastUpstreamOkAt with nothing to clear it — /api/health sat degraded until the 10-minute window expired, producing two Discord pings per blip.

- fletcher /stats now reports `consecutiveUpstreamFails` (reset on any success)
- health verdict extracted to `server/utils/fletcherRelayHealth.js` (8 tests) — degraded only at ≥2 consecutive failures; a lone blip reports ok with the blip noted in the detail
- the real failure mode (serial 403s, v26.6.0 release day) still trips immediately
- older fletcher builds without the counter are treated as a single blip so a mixed-version deploy window under-pages rather than re-noising

Verified live on staging before this PR (health shows the new "relays ok" detail).

### DIFF
--- a/fletcher/server.js
+++ b/fletcher/server.js
@@ -56,9 +56,13 @@ const stats = {
   // reads these so "fletcher alive but every relay failing" no longer
   // reports green. 403/429/5xx and network errors count as errors; 404 is a
   // routine missing-satellite answer and counts as ok.
+  // consecutiveUpstreamFails lets the probe tell one transient blip (which
+  // used to page watchtower for a guaranteed 10 minutes) from an upstream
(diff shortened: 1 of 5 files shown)


accius:
Superseding with #—: Staging picked up the footprint-cone feature after this PR was opened, and since this PR's head is the Staging branch itself, merging now would carry that feature into prod. Re-cut as a surgical cherry-pick from main.

